PR #73, opened 2022-11-23
ci(root): fixed conflicting peer dependancies

GCHQ-Developer-741:
Downgraded lerna back to 4.0.0 to fix conflicting peer dependencies

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,7 @@
         "husky": "^7.0.4",
         "jest": "^27.5.1",
         "jest-cli": "^27.5.1",
-        "lerna": "^6.0.3",
+        "lerna": "^4.0.0",
         "lerna-audit": "^1.3.3",
         "prettier": "2.5.0",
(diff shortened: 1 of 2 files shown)


GCHQ-Developer-112 (approved):

ad9242:
I still get errors. To recreate:
- clone repo
- delete all package-lock.json files (at the root & from under each of the packages)
- run npm install (succeeds, but some vulnerabilities)
- run npm run bootstrap - fails

ad9242:
could we try leaving the version of lerna as is at 6 & instead move to a newer version of commitlint? - we are using 13 but looks like it's up to 17 now

I've gone for these versions on my current branch, to fix vulnerabilities:
"@commitlint/cli": "^17.3.0",
    "@commitlint/config-conventional": "^17.3.0",
    "@commitlint/config-lerna-scopes": "^17.2.1",

ad9242 (approved):

Issue #92, opened 2022-11-28
Overflowing component previews


GCHQ-Developer-530:
## Summary of the bug
After the work has been merged that makes component preview responsive and not require three dimensional scrolling, some of the components overflow their component preview. 

## 🪜 How to reproduce
Tell us the steps to reproduce the problem:
1. Go to page: components
2. Click on: page header
3. Click on: code
4. Resize to iPhone
5. See that page header overflows out of the component preview box

## 🧐 Expected behaviour
Every component should sit within the component preview and not overflow

## 📸 Screenshots or code
<img width="314" alt="Screenshot 2022-11-28 at 13 46 30" src="https://user-images.githubusercontent.com/111882034/204293155-839c25bb-24ca-41f1-b75d-370557c24188.png">

<img width="438" alt="Screenshot 2022-11-28 at 13 50 51" src="https://user-images.githubusercontent.com/111882034/204294071-057fd154-3eab-43bf-b73a-874096c7ffc5.png">

## Additional info
Component demos needing work: 
- [x] Checkbox conditional and dynamic (merged)
- [x] Data entity with status tags
- [x] Page header (every demo overflows here)
- [x] Radio conditional and dynamic  (merged)
- [x] Side nav react router (can be fixed in design system)
- [x] Stepper (marked as done as the new compact stepper component will fix this. The minimum value for the step connect should stay at 100px)
- [x] Tooltip (left and right overflow out of the box and aren't readable) (Can be fixed in design system)

GCHQ-Developer-112:
Page header should be resolved. I recently worked on this in the ui-kit repo but we haven't done a release for our packages so you won't see a change until there's a new release.

Stepper can't be resolved until the compact stepper has been built. In the cases where the stepper is overflowing, the expected behaviour is that it would switch to the compact stepper.

In terms of the checkbox & radio's conditional and conditional dynamic variants, there's a CSS fallback on the input width. If you set the variable --input-width on the text field e.g. <style> --input-width: 250px </style>, this will override the 320px width. Alternatively you may want to add a media query in the checkbox & radio CSS files for the XS breakpoint that sets --input-width at 100%. This is what I did for the page header and it can be overwritten if the end user sets a different value for --input-width

PR #113, opened 2022-11-29
fix(web-components): remove margin left and add max width on xs breakpoint

GCHQ-Developer-530:
## Summary of the changes
Remove margin left on end component and add max width to value on xs breakpoint

## Related issue
#92

## Checklist


GCHQ-Developer-741 (dismissed):

packages/web-components/src/components/ic-data-row/ic-data-row.css
```diff
@@ -55,3 +55,11 @@
   width: 128px;
   margin-bottom: var(--ic-space-xs);
 }
+
+:host(.breakpoint-xs) .value {
+  max-width: 75vw;
```
jd239, line 60: I think this should be `75%` of the parent container not 75% off the viewport.

jd239 (changes requested):
A single comment :)

packages/web-components/src/components/ic-data-row/ic-data-row.css
GCHQ-Developer-530: I've updated it to 95% as 75 seemed too squished when converting from vw

jd239 (approved):

GCHQ-Developer-847 (approved):

Issue #80, opened 2022-11-24
Input components should respond to form reset


ad9242:
## Summary
When included on a form, the input components should be reset to their original values/state if the form is reset, eg by having a button of type "reset" on the form

## 💬 Description
The components will need to store their initial state so that this can be reset when the form reset occurs. it should be possible to listen for the reset event on the parent form

Affects the following components:
textfield
radio
checkbox
switch 
select

maybe others?

## 💰 Use value
This would make the components behave like their native input counterparts

PR #116, opened 2022-11-29
feat(web-components): adds support for form reset event

ad9242:
## Summary of the changes
Input components now reset to initial state when used on a form & a form reset is triggered.
Adds test coverage for components too.
Fixes issue in switch component where switch did not update correctly on toggle when showState & checked props were set

## Related issue
#80 

## Checklist


packages/web-components/src/components/ic-radio-option/ic-radio-option.tsx
```diff
@@ -76,10 +81,14 @@ export class RadioOption {
   }
 
   private defaultRadioValue: string = "";
+  private skipFocus = false;
```
jd239, line 84: Just so I understand better, why do we have a skipFocus flag? Was there initial a focus state on the radio button when reset was pressed?

packages/web-components/src/components/ic-radio-option/ic-radio-option.tsx
```diff
@@ -37,7 +40,7 @@ export class RadioOption {
   /**
    * Provide a value for the input.
    */
-  @Prop() value!: string;
+  @Prop({ mutable: true }) value!: string;
```
jd239, line 43: Does this need to be `mutable` as I cannot see where we are setting the value internally again?

jd239 (commented):
Only two questions and not really code changes. Looks good! :)

packages/web-components/src/components/ic-radio-option/ic-radio-option.tsx
ad9242: We are changing the state of selected which invokes the handleClick by way of the watch on the selected prop. When this happens the radio receives focus, but we do not want it to in this case

packages/web-components/src/components/ic-radio-option/ic-radio-option.tsx
ad9242: Was giving a warning in the unit tests, as the value is updated in handleClick method

jd239 (approved):

GCHQ-Developer-741 (approved):

Issue #111, opened 2022-11-29
Add a "no results" state to the search bar component when the disableFilter prop is applied


GCHQ-Developer-847:
## Summary
Add a "no results" state to the search bar component when the disableFilter prop is applied.

## 💬 Description
Add a "no results" state to the search bar component when the disableFilter prop is applied i.e. when the options passed to the search bar are already filtered from an external source.

## 💰 Use value
When the filtering is not handled by the search bar component itself, there is no feedback to inform the user that no results have been found if the search term matches none of the options.

This issue is evident with the search bar on the guidance site.

PR #124, opened 2022-12-01
feat(web-components): add "no results" state to search bar variant with disabled filter

GCHQ-Developer-847:
## Summary of the changes
Added "no results" state to search bar variant with disabled filter, fixed z-index of tooltips on clear and search buttons, stopped icSubmitSearch being emitted when enter pressed while search button showing as disabled, set text of ARIA live region to emptyOptionListText when no results are found (was previously being set to "1 result available"), stopped live region updating when menu not displayed due to charactersUntilSuggestions prop.

I have checked this by importing it into the guidance site locally. Another way this can be tested is to not pass any options to it and it should display "No results found"

## Related issue
#111

## Checklist

### DIFF
--- a/packages/web-components/src/components/ic-search-bar/ic-search-bar.tsx
+++ b/packages/web-components/src/components/ic-search-bar/ic-search-bar.tsx
@@ -226,7 +226,7 @@ export class SearchBar {
   @Listen("keyup", {})
   handleKeyUp(ev: KeyboardEvent): void {
     if (ev.key === "Enter") {
-      if (this.preventSubmit) {
+      if (this.preventSubmit || this.isSubmitDisabled()) {
         return;
       }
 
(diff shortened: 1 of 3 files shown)


packages/web-components/src/components/ic-search-bar/ic-search-bar.tsx
```diff
@@ -509,23 +512,41 @@ export class SearchBar {
...
+        } result${this.filteredOptions.length > 1 ? "s" : ""} available`;
+      }
     }
   };
 
   private hasOptionsOrFilterDisabled = (): boolean =>
     this.options.length > 0 || this.disableFilter;
 
+  private hadNoOptions = (): boolean =>
+    this.filteredOptions.length > 0 &&
```
jd239, line 541: Should we make this stricter and see if it's equal to 1 as there should only be one item when 'no results' is displayed

jd239 (approved):

GCHQ-Developer-530 (approved):

PR #136, opened 2022-12-05
feat(web-components): fixes radio and checkbox in WHCM

ad9242:
## Summary of the changes
fixes radio & checkbox component styling in WHCM
also fixes radio spacing as did not match figma specs

## Checklist


GCHQ-Developer-741:
So a few WHCM issues I spotted:

- The conditional dynamic form appears to be disabled based on text colour (for both checkbox and radio).
Screenshot from 2022-12-05 16-58-25
- The indeterminate checkboxes appear ok at first, but then once unchecked and checked look like normal checkboxes.
Screenshot from 2022-12-05 16-48-03 Screenshot from 2022-12-05 16-48-10

ad9242:
The placeholder is behaving correctly I think - this example looks the same in WHCM: https://www.w3schools.com/tags/tryit.asp?filename=tryhtml5_input_placeholder

Re the indeterminate, again, this is behaving as per a standard check box: Once a check box is clicked, it can only be toggled between checked & unchecked eg https://www.w3schools.com/jsref/tryit.asp?filename=tryjsref_checkbox_indeterminate

I have raised #137 to investigate the indeterminate behaviour, but we decided to leave as is for the WHCM as otherwise we have to add lots of css to override it which we would then have to remove when that issue is resolved

GCHQ-Developer-530 (approved):

PR #163, opened 2022-12-13
fix(web-components): updated foreground colour calculations, added colour contrast check

GCHQ-Developer-847:
## Summary of the changes
Updated foreground colour calculations - updated dark mode threshold and changed -1000 to -10000 as foreground colour was sometimes showing as grey because this number wasn't large enough to bring it over 255, ran prettier:fix, added colour contrast check to ic-theme component which creates a console warning if the chosen colour has a colour brightness which is too close to being halfway between that of the black and white foregound colours - and therefore not a high enough contrast.

## Related issue
#3

### DIFF
--- a/packages/web-components/src/components/ic-theme/ic-theme.tsx
+++ b/packages/web-components/src/components/ic-theme/ic-theme.tsx
@@ -14,6 +14,11 @@ import {
   hexToRgba,
   rgbaStrToObj,
 } from "../../utils/helpers";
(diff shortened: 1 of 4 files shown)


GCHQ-Developer-847:
rgba(133, 133, 133, 1) is one colour that doesn't provide enough contrast with the black or white text - change the color prop of the ic-theme component to that to see the console warning

GCHQ-Developer-530 (approved):

Issue #141, opened 2022-12-06
Static and dismissible chips


GCHQ-Developer-530:
## Summary
A Chip displays a selected input item and can be interactive to allow the removal of the selection. Often used in filters or to trigger an action similar to a button. 

## 💰 User value
Chips add to the different selection methods that can be used. 

## 📚 User Stories
As a ICDS user
I need the following chip variants:

basic
outline
So that I have a choice of appearance and can differentiate hierarchy. 

 

As an ICDS user

I need to be able to set if a chip is clickable 

So that it can be used to trigger an action. 

 

As an ICDS user

I need a clickable chip to have the following interaction states:

Default
Hover
Pressed
Focussed
Disabled
So that a user can interact with the link effectively.

 

As an ICDS user

I need to enable a chip to be dismissible

So that I can remove/de-select it if needed.

## 📝 Acceptance Criteria 
Given that a dismissible chip component exists on a page,

When I Hover over the chip,

Then I should see the Hover state.

 

Given that a dismissible chip component exists on a page,

When I click down/tap down on the chip,

Then I should see the pressed state.

 

Given that a dismissible chip component exists on a page,

When I tab to the chip,

Then I should see the focused state.

 

Given that a dismissible chip component exists on a page, 

When it has the disabled attribute,

Then I should see the disabled state,

and the chip should not be clickable.  

Given that a dismissible chip component exists on a page,

When I click on the chip,

Then the chip should be removed.

 

Given that a dismissible chip exists on a page, 

When it is clicked,

And it does not dismiss,

Then it should throw an error.  

Given that a static chip exists on a page, 

When I interact with the chip,

Then its state should not be changed. 

## ✏ Designs
If there's a Figma design file (or other mock-up), include it here.

## 🧾 Guidance
Variants:

Dismissible
---Filled
---Outlined
Static
Sizes:

Small
Default
Large

GCHQ-Developer-847:
Clickable chips had initially been developed due to an old AC that was left in this ticket, but, after a discussion with the designers, these will be implemented in the future because they need more investigation e.g. about focus states. 

I have 'saved' the clickable chips work by making a copy of all the files that were added / changed to make the chip component (before clickable chips were removed): [GCHQ-Developer-847/141-chips-with-clickable](url).

PR #174, opened 2022-12-15
141 static and dismissible chips

GCHQ-Developer-530:
## Summary of the changes
Implement chip compononent. Includes static, clickable and dimissible chips. The chip can also be
filled or just have an outline, and it can be disabled. If clickable, either an <a> or <button> is
rendered depending on whether an href is provided. Includes small, default and large size variants.
Ran prettier:fix which fixed some prettier issues in other files.

## Related issue
#141 

## Checklist


packages/web-components/src/components/ic-chip/ic-chip.tsx
```diff
@@ -0,0 +1,224 @@
...
+                class={{
+                  ["svg-container"]: true,
+                  ["dismiss-icon"]: true,
+                }}
+                innerHTML={dismissIcon}
+                onClick={this.dismissAction}
+                variant="icon"
+                size={size === "small" ? "small" : "default"}
+                appearance={IcThemeForegroundEnum.Dark}
+                title="dismiss"
```
GCHQ-Developer-847, line 215: I tried using a screen reader on the clickable dismissible chip and when I focus on the whole chip, it reads out "Label dismiss". It might be reading out the tooltip text because the tooltip is a child of the chip `<button>` so wondering if a solution could be to move the close icon button outside and have it next to the `<Component>` rather than inside it (and then add styling so it appears inside it)? Not completely sure if that would work though but just an idea :)

Also just noticed that there are a couple of accessibility issues appearing in the Accessibility tab in storybook - one of which is about having nested focusable elements, so that solution might fix that issue

packages/web-components/src/components/ic-chip/ic-chip.css
```diff
@@ -0,0 +1,142 @@
...
+.filled.clickable:hover {
+  background-color: var(--ic-architectural-300);
+  cursor: pointer;
+}
+
+.filled.clickable:active {
+  background-color: var(--ic-architectural-400);
+}
+
+.outline {
```
GCHQ-Developer-847, line 85: Looks like there's a background colour on the outline variant:
image
but it's just transparent in the designs (and this is making the change to the active / pressed colour when it's clickable less noticeable)

Also the hover state on the clickable variant doesn't seem to work for me? Maybe it's just not showing up because of the background

GCHQ-Developer-847 (dismissed):
Looking good! Just added a few comments

GCHQ-Developer-847:
Not quite ready for review again yet

GCHQ-Developer-847:
Ready for review again now

GCHQ-Developer-847:
I have discussed this PR with the designers and there are some changes that need to be made - so have turned it into a draft PR for now

GCHQ-Developer-847:
Ready for review again now - clickable chip functionality has been removed as requested by the designers

ad9242:
There doesn't seem to be a pressed state color change for the dismissible variant. The Figma specs show darker colour when pressed, but it seems to just revert to the unhovered background colour

GCHQ-Developer-530:
> There doesn't seem to be a pressed state color change for the dismissible variant. The Figma specs show darker colour when pressed, but it seems to just revert to the unhovered background colour

I think the pressed state in the figma file was referring to when clickable chip functionality was there, i.e when the whole chip is pressed then the pressed state would change colour, as opposed to when the dismissible button was pressed. I might be wrong though, so it might be worth @DBD324 taking a look?

ad9242:
> > There doesn't seem to be a pressed state color change for the dismissible variant. The Figma specs show darker colour when pressed, but it seems to just revert to the unhovered background colour
> 
> I think the pressed state in the figma file was referring to when clickable chip functionality was there, i.e when the whole chip is pressed then the pressed state would change colour, as opposed to when the dismissible button was pressed. I might be wrong though, so it might be worth @DBD324 taking a look?

ah ok. tbh, now I've looked at other components, it does make it consistent with them

ad9242:
should the WHCM focus border color be on the whole chip, rather than just the button? - not sure!
image

ad9242:
could we also apply the disabled text color for disabled variants?
image

GCHQ-Developer-530:
> should the WHCM focus border color be on the whole chip, rather than just the button? - not sure! image

Hmm not sure, I think because it can't be clicked, maybe it shouldn't be focusable, like the whole alert doesn't have a focus border

GCHQ-Developer-530:
> should the WHCM focus border color be on the whole chip, rather than just the button? - not sure! image

Hmm not sure, I think because it can't be clicked, maybe it shouldn't be focusable, like the whole alert doesn't have a focus border

GCHQ-Developer-530:
> could we also apply the disabled text color for disabled variants? image

Done :)

GCHQ-Developer-847:
> > > There doesn't seem to be a pressed state color change for the dismissible variant. The Figma specs show darker colour when pressed, but it seems to just revert to the unhovered background colour
> > 
> > 
> > I think the pressed state in the figma file was referring to when clickable chip functionality was there, i.e when the whole chip is pressed then the pressed state would change colour, as opposed to when the dismissible button was pressed. I might be wrong though, so it might be worth @DBD324 taking a look?
> 
> ah ok. tbh, now I've looked at other components, it does make it consistent that there is no pressed darker state

Yep we originally thought the hovered and pressed states in Figma were for clickable chips, but after speaking to @fbd399 it turns out that they actually refer to when the dismiss button is hovered on / pressed.

When I added the hover state for when the dismiss button is pressed, I didn't add the pressed state because when you click on the chip (mouse down), the focus indicator appears and the focus state should always override other states.

Because of this, I didn't think it would be necessary to set the background to being darker when pressed as it would just be overridden by the focus state so wouldn't be seen at all. If we decide in the future that the focus indicator should only appear on mouse up then we would need to revisit the pressed state on chips, but felt it wouldn't be worth implementing it just yet? Since it isn't going to be used (especially as it would be a bit more complicated than adding CSS - it will probably be the same as how the hover state is done, using a variable in state)

packages/web-components/src/components/ic-chip/ic-chip.tsx
```diff
@@ -0,0 +1,142 @@
...
+            <div class="icon">
+              <slot name="icon" />
+            </div>
+          )}
+          <ic-typography
+            variant="label"
+            apply-vertical-margins={false}
+            class="label"
+          >
+            <span>{label}</span>
```
jd239, line 115: I'm thinking if is it worth passing through the label as a slot instead of a prop so it appears in the Light DOM? Just been testing the React variant of chips using React Testing library. I appreciate the dev can use `data-testid` but putting the label as a slot allows the dev to also use `getByText` as well as `getByTestId` so there is added flexibility.

GCHQ-Developer-530: Yeah I'm not sure it'd be possible. @GCHQ-Developer-847 I'm not sure if you have ideas?
One of the other issues with having it as a slot would be that it'd be harder to restrict devs from not putting anything other than text in the chip, i.e we wouldn't want them to add a button etc.

jd239: Yeah, I get what you mean. I think the onus should be on the developer to get that right (re: adding a button). We can have something in our guidance for it as well just to cover our backs. 

However, after further investigation, I have been able to get the following to work, so we should be okay.

```jsx
const { getByTestId } = render(<IcChip data-testid='chip' label='User' />)

const chip = getByTestId('chip');

await waitFor(() => {
  expect(chip).toHaveClass('hydrated');
})

expect(chip.label).toBe('User')
```

We can force dev in this direction if they have any issues.

jd239:
The size of outline and static are slightly different in WHC mode. The static is slightly bigger by 2px.

ad9242:
in WHCM, as you tab between the dismissible variants, they jump in size as they have a focus border applied\removed

ad9242 (approved):

Issue #198, opened 2022-12-22
Select component has empty space for input validation when none is provided


GCHQ-Developer-741:
## Summary of the bug
<ic-input-validation> is still rendered even when no values are passed in, leaving an empty space below the component

## 🪜 How to reproduce
1. Add a select with no validation status or text
2. Highlight the component in devtools
3. There is empty space below

## 🧐 Expected behaviour
The space should be removed if no validation is provided, to give a cleaner look in real apps.

## 📸 Screenshots or code
![Screenshot from 2022-12-22 11-29-16](https://user-images.githubusercontent.com/117368893/209124997-d7e9bc05-27f2-438a-8840-2883e82a2f1a.png)

PR #202, opened 2022-12-22
fix(web-components): removed empty validation space when no status is…

GCHQ-Developer-741:
… provided

When no status or validation text is passed in, select will no longer render the empty space

## Related issue
#198 

## Checklist


MI6-255:
This was based off a request from an internal customer. However, wouldn't this cause forms to jump around the place as validation appears?

jd239:
It would yes. Just had a look at example forms and based on the position of the validation text, all the examples pushed down the remaining inputs as the validation text appeared.

jd239 (approved):

GCHQ-Developer-741:
> This was based off a request from an internal customer. However, wouldn't this cause forms to jump around the place as validation appears?

Have spoken to designers about this, has been decided that this is acceptable. Only comments were around screen reader testing, which I have undertaken to ensure that it still reads out the status correctly.

MI6-255 (changes requested):
I want us to discuss the jumping around a little further, I'll reach out

MI6-255 (approved):

Issue #197, opened 2022-12-22
Back to top component overlaps footer content


ad9242:
## Summary of the bug
The back to top component overlaps the footer component on a page

## 🧐 Expected behaviour
Back to top should display above footer when it is scrolled into view (as per guidance)

## 📸 Screenshots or code
![image](https://user-images.githubusercontent.com/116102637/209100265-0a1b9961-303d-4d9a-a522-3da456e45cb8.png)

## Additional info
Tell us anything else useful to help us fix or understand the problem.

PR #207, opened 2023-01-03
fix(web-components): fixes back to top positioning

ad9242:
## Summary of the changes
Fixes positioning so that it no longer overlaps ic-footer

## Related issue
#197 

## Checklist

### DIFF
--- a/packages/web-components/src/components/ic-back-to-top/ic-back-to-top.tsx
+++ b/packages/web-components/src/components/ic-back-to-top/ic-back-to-top.tsx
@@ -42,7 +42,13 @@ export class BackToTop {
   };
 
   private setFooterVisible = (visible: boolean) => {
-    this.footerVisible = visible;
+    this.checkForClassificationBanner();
+    if (typeof window !== "undefined" && window.scrollY === 0) {
+      this.footerVisible = false;
+    } 
+    else {
+      this.footerVisible = visible;
+    } 
   };
 
   private targetElObserverCallback = (entries: IntersectionObserverEntry[]) => {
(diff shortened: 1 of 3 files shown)


GCHQ-Developer-530:
Now that it doesn't overlap the footer, is there an issue with overlapping the rest of the page content? e.g here I made the text longer and back to top overlaps the text
Screenshot 2023-01-03 at 14 13 28

packages/web-components/src/components/ic-back-to-top/ic-back-to-top.tsx
```diff
@@ -42,7 +42,13 @@ export class BackToTop {
   };
 
   private setFooterVisible = (visible: boolean) => {
-    this.footerVisible = visible;
+    this.checkForClassificationBanner();
+    if (typeof window !== "undefined" && window.scrollY === 0) {
```
jd239, line 46: Even though it's against our guidance, what happens if the footer is not below the fold but window.scrollY is 0?

jd239 (changes requested):
Just a single comment! 😃

packages/web-components/src/components/ic-back-to-top/ic-back-to-top.tsx
ad9242: The back to top would not be displayed at all, as the target el would not have scrolled off screen yet

jd239 (approved):

ad9242:
> Now that it doesn't overlap the footer, is there an issue with overlapping the rest of the page content? e.g here I made the text longer and back to top overlaps the text. Screenshot 2023-01-03 at 14 13 28 Not sure if there's a way for it to cause the text to wrap?

I think this is ok as the guidance shows the back to top overlaying content

GCHQ-Developer-530 (approved):

Issue #228, opened 2023-01-06
Duplicate events and add deprecation warnings


ad9242:
As identified in #227, there are inconsistencies in the events emitted by components. This issue covers making changes which would be classed as breaking by putting in place deprecation warnings ahead of the eventual breaking change

For each of the components below, we should duplicate the events and emit both for the time being. The docs for the old event should also be modified to include a deprecation warning.

Note: some of the new names proposed are open for discussion

**ic-alert:**
"dismiss" event should become "icDismiss"

**ic-checkbox:** 
"checkboxChecked" should become "icCheck"? (Q: should this match name we use for radio option?)

**ic-radio-option:**
"radioOptionSelected" should become "icCheck"? 

**ic-search-bar**
icInputBlur, icInputFocus - can be removed as not required.  These were initially used internally but no longer are. There is a risk that someone may be using these, but they should really be using "icSearchBarBlur" and "icSearchBarFocus". We should add a note to that effect & explain that "icInputBlur" and "icInputFocus" are deprecated 

**Note:** after some discussion we have opted to keep icSearchBarFocus instead of making this "icFocus", as the ic-top-navigation component also listens on this. Without a way to target event listeners within Stencil, we cannot change this to be "icFocus" as other components used within top navigation may emit an event with the same name.
Possibly we would need to pass additional info in the event detail to describe which component emitted it. If so we would probably need a separate issue to explore this

**ic-tab-context:** 
"tabSelect" should become "icTabSelect"?

Issue #230, opened 2023-01-06
Remove deprecated events and props


ad9242:
A number of props\events have been renamed and marked as deprecated (including Issue #228 which implemented a deprecation warning for some events)

The following changes need to be made:

- remove any references from the component docs of the old names (can search for `@deprecated` in the code)
- remove any code which is emitting the old events
- ensure new props\events are covered in testing
- remove testing of old props\event names (if required)
- ensure any examples do not reference the old names (if they do on the site, a separate ticket may need to be raised on the ic-design-system repo  to remove them)

Also look into removing any props that seem redundant but haven't been deprecated yet - an example of this is "current" prop for ic-step, which could just be replaces by type="current" which we already have.

GCHQ-Developer-530:
Remove any props/events/methods on the dialog related to buttons. Buttons should only be added to dialog using dialog-controls slot

MI6-255:
Bare in mind there may be more tests for code coverage

PR #284, opened 2023-01-11
refactor(web-components): update events to more consistent names

GCHQ-Developer-530:
## Summary of the changes
Update events to more consistent names, use deprecated flag on previously used names, and add internal to inline prop in side navigation

## Related issue
#228

## Checklist


packages/web-components/src/components/ic-alert/ic-alert.tsx
```diff
@@ -68,17 +68,22 @@ export class Alert {
...
   @Event() dismiss: EventEmitter<void>;
 
+  /**
+   * Is emitted when the user dismisses the alert.
+   */
+  @Event() icDismiss: EventEmitter<void>;
+
   private dismissAction = (): void => {
-    this.dismiss.emit();
+    this.icDismiss.emit();
```
ad9242, line 86: We should still emit both events for the time being.

ad9242 (changes requested):
We should still emit the old events for the time being for any events that were previously available\documented. We have a separate issue (#230 ) for when we come to make the breaking change in a future major release

packages/web-components/src/components/ic-tab-context/ic-tab-context.tsx
```diff
@@ -54,10 +54,15 @@ export class TabContext {
...
   /**
-   * Emitted when a user selects a tab.
+   * @deprecated This event should not be used anymore. Use icTabSelect instead.
    */
   @Event() tabSelect: EventEmitter<IcTabSelectEventDetail>;
 
+  /**
+ * Emitted when a user selects a tab.
+ */
+  @Event() icTabSelect: EventEmitter<IcTabSelectEventDetail>;
```
GCHQ-Developer-847, line 64: Similar to my other comment - 'tabSelect' / 'onTabSelect' is still being used in the React tabs stories, I think we should change it to 'onIcTabSelect' there now

packages/web-components/src/components/ic-tab-context/ic-tab-context.tsx
```diff
@@ -66,7 +71,7 @@ export class TabContext {
     ) {
       this.selectedTab = event.detail.position;
     }
-    this.tabSelect.emit({
+    this.icTabSelect.emit({
       tabIndex: event.detail.position,
     });
   }
```
GCHQ-Developer-847, line 77: Line 139 in this file still has a reference to the old event name

packages/web-components/src/components/ic-checkbox/ic-checkbox.tsx
```diff
@@ -67,15 +67,20 @@ export class Checkbox {
...
   /**
-   * Emitted when a checkbox has been checked
+   * @deprecated This event should not be used anymore. Use icCheck instead.
    */
   @Event() checkboxChecked: EventEmitter<void>;
 
+  /**
+ * Emitted when a checkbox has been checked.
+ */
+  @Event() icCheck: EventEmitter<void>;
```
GCHQ-Developer-847, line 77: Also, I'm wondering if we should call this 'icCheckboxCheck' if we're calling the event on the radio buttons 'icRadioCheck'? I guess ideally we would call them both 'icCheck' but I guess it's possible that may cause issues with event listeners if you have both radio buttons and checkboxes on a page and they're using the same event name

GCHQ-Developer-847 (changes requested):

packages/web-components/src/components/ic-checkbox/ic-checkbox.tsx
ad9242: I think maybe use icCheck for both. We already use icFocus, icBlur on multiple components. If you were creating an event listener, you would set it on the component, I would have thought?

GCHQ-Developer-530: I tried using icCheck for both and it did seem to cause issues, however that might have been if I hadn't updated other values of it. I'll have another look :)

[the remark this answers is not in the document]
GCHQ-Developer-530: Due to @ad9242 's comment about emitting both events and then fixing the breaking changes later, I think this work would go in the other ticket, once I've added back the work to emit both events. Let me know if you think it needs to be done in this PR though.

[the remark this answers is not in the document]
GCHQ-Developer-847: Ok great, just wanted to make sure it's done at some point so that sounds good :)

GCHQ-Developer-847 (approved):

ad9242 (approved):

Issue #243, opened 2023-01-09
[ic-tooltip] Add disableHoverListener


GCHQ-Developer-530:
## Summary
Enable a tooltip to be launched by a trigger other than hover. 

This would enable click to display functionality. 

## 💰 Use value
Would allow the implementer to control the open close state of the tooltip manually and remove the on hover listener if required.

## Additional info
**Given** that a component exists with a click event that displays a tooltip,
**and** the tooltip is set to have a disabled hover listener, 
**When** the click is registered, 
**Then** the tooltip should display. 

**Given** that a component exists with a click event that displays a tooltip,
**and** the tooltip is set to have a disabled hover listener, 
**When** the component is hovered, 
**Then** the tooltip should not display.

PR #288, opened 2023-01-13
ic-tooltip add disablehoverlistener

GCHQ-Developer-530:
## Summary of the changes
Added disableHover prop that will stop the tooltip appearing on hover, and will require a click to appear. The tooltip will still disappear on mouse leave though.

## Related issue
#243 

## Checklist


ad9242 (commented):
Just one comment really: Does it need to emit an event? Given that it's an @internal event and the listener is the same component instance?
Not against it, but maybe see what others think?

GCHQ-Developer-530:
> Just one comment really: Does it need to emit an event? Given that it's an @internal event and the listener is the same component instance? Not against it, but maybe see what others think?

I'm not sure, I'd based it off other @internal events which still seemed to emit an event, but that might just be my lack of knowledge on this stuff, rather than being best practice.

ad9242:
> > Just one comment really: Does it need to emit an event? Given that it's an @internal event and the listener is the same component instance? Not against it, but maybe see what others think?
> 
> I'm not sure, I'd based it off other @internal events which still seemed to emit an event, but that might just be my lack of knowledge on this stuff, rather than being best practice.

So I think where we emit other @internal events is where they are child components that need to communicate back with the parent. In which case the parent will have a @Listener

GCHQ-Developer-530:
> > > Just one comment really: Does it need to emit an event? Given that it's an @internal event and the listener is the same component instance? Not against it, but maybe see what others think?
> > 
> > 
> > I'm not sure, I'd based it off other @internal events which still seemed to emit an event, but that might just be my lack of knowledge on this stuff, rather than being best practice.
> 
> So I think where we emit other @internal events is where they are child components that need to communicate back with the parent. In which case the parent will have a @Listener

Ah okay, so instead of the listener and emitted event, it could just have this.disableHover && this.showEvents = ["click"];?

packages/web-components/src/components/ic-tooltip/ic-tooltip.stories.mdx
```diff
@@ -31,6 +31,27 @@ import readme from "./readme.md";
...
+      html`<ic-tooltip
+        label="This shows 'Code' has been copied"
+        target="test-button-disable-hover"
+        disable-hover="true"
+        ><button
+          id="test-button-disable-hover"
+          aria-describedby="ic-tooltip-test-button-disable-hover"
+          onclick="navigator.clipboard.writeText('Code')"
+        >
+          Disable Hover
```
ad9242, line 48: Really minor one, but this implies to me that clicking the button will disable the hover. Maybe "Show tooltip" instead?

ad9242 (approved):

Issue #179, opened 2022-12-15
[ic-top-navigation] Allow use of gatsby (or any slotted) links for app title link


GCHQ-Developer-530:
## Summary
Allow for the app title to be a gatsby link.

## 💬 Description
We currently render an `<a>` tag in top nav around the app-icon & title, but this means you cannot take advantage of gatsby routing for the link (which will cause flashing of the site)

We probably need to add a slot to the top nav, but need to look into it as we already have a slot for the app-icon

We would also potentially need to remove the need for appTitle to be a required prop

## 💰 Use value
This will help with flashing of the design system.

PR #364, opened 2023-02-13
Allow use of gatsby or any slotted links for app title link in top nav

GCHQ-Developer-530:
## Summary of the changes
Add a slot for use of router links for top navigation app title, and update app title prop to not be required, but still show the console error if a top nav has no app title or slotted app title link.
Updated the react router story to use a routed app title.

## Related issue
#179 

## Checklist

### DIFF
--- a/packages/react/src/stories/ic-top-navigation.stories.mdx
+++ b/packages/react/src/stories/ic-top-navigation.stories.mdx
@@ -476,7 +476,10 @@ export const DailyTippers = () => (
       ),
     ]}
   >
-    <IcTopNavigation appTitle="ApplicationName" status="alpha" version="v0.0.7">
+    <IcTopNavigation status="alpha" version="v0.0.7">
+    <NavLink to="/" slot="router-app-title">
+            ApplicationName
+          </NavLink>
       <svg
         slot="app-icon"
         xmlns="http://www.w3.org/2000/svg"
--- a/packages/react/src/stories/ic-top-navigation.stories.mdx
+++ b/packages/react/src/stories/ic-top-navigation.stories.mdx
@@ -477,9 +477,9 @@ export const DailyTippers = () => (
     ]}
   >
     <IcTopNavigation status="alpha" version="v0.0.7">
-    <NavLink to="/" slot="router-app-title">
-            ApplicationName
-          </NavLink>
+      <NavLink to="/" slot="app-title">
+        ApplicationName
+      </NavLink>
       <svg
         slot="app-icon"
         xmlns="http://www.w3.org/2000/svg"
--- a/packages/web-components/src/components/ic-top-navigation/ic-top-navigation.tsx
+++ b/packages/web-components/src/components/ic-top-navigation/ic-top-navigation.tsx
@@ -31,6 +31,7 @@ import {
 
 /**
  * @slot app-icon - Content will be rendered to left of app title.
+ * @slot app-title - Handle routing by nesting a route in the app title.
  * @slot search - Content will be rendered in search area to left of buttons.
  * @slot toggle-icon - Icon to be displayed on the button to toggle search slot content on smaller devices
  * @slot navigation - Content will be rendered in navigation panel.
@@ -45,9 +46,9 @@ export class TopNavigation {
   @Element() el: HTMLIcTopNavigationElement;
 
   /**
-   * The app title to be displayed.
+   * The app title to be displayed. This is required, unless a slotted app title link is used.
    */
-  @Prop() appTitle!: string;
+  @Prop() appTitle: string;
 
   /**
    *  The URL to navigate to when the app title is clicked.
@@ -261,10 +262,11 @@ export class TopNavigation {
   componentDidLoad(): void {
     checkResizeObserver(this.runResizeObserver);
 
-    onComponentRequiredPropUndefined(
-      [{ prop: this.appTitle, propName: "app-title" }],
-      "Top Navigation"
-    );
+    !isSlotUsed(this.el, "app-title") &&
+      onComponentRequiredPropUndefined(
+        [{ prop: this.appTitle, propName: "app-title" }],
+        "Top Navigation"
+      );
   }
 
   disconnectedCallback(): void {
@@ -299,6 +301,12 @@ export class TopNavigation {
       : "Show search";
     const menuSize = this.deviceSize <= DEVICE_SIZES.S ? "small" : "default";
 
+    const Component = isSlotUsed(this.el, "app-title") ? "div" : "a";
+
+    const attrs = Component == "a" && {
+      href: this.href,
+    };
+
     return (
       <Host
         class={{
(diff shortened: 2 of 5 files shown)


packages/web-components/src/components/ic-top-navigation/ic-top-navigation.tsx
```diff
@@ -312,17 +323,30 @@ export class TopNavigation {
             <header role="banner">
               <div class="top-panel-container">
                 <div class="app-details-container">
-                  {hasTitle && (
-                    <a class="title-link" href={this.href}>
+                  {this.hasRouterAppTitle() ? (
+                    <div class="title-link" tabindex="0" role="link">
```
GCHQ-Developer-847, line 327: To make things simpler, could we just allow the user to focus on the actual slotted link itself? I.e. remove the tabindex attributes, and then set the focus indicator styling by doing :focus-within on the title-link class. And then I don't think we would need the aria-label or role to be set as the screen reader will read out the content of the focussed link and the fact that it's a link since it's focussing on the `<a>`

Not 100% sure if it would work but maybe worth a try? Sorry if you've already given it a go :)

packages/web-components/src/components/ic-top-navigation/ic-top-navigation.tsx
```diff
@@ -312,17 +320,30 @@ export class TopNavigation {
             <header role="banner">
               <div class="top-panel-container">
                 <div class="app-details-container">
-                  {hasTitle && (
-                    <a class="title-link" href={this.href}>
+                  {this.hasRouterAppTitle() ? (
```
GCHQ-Developer-847, line 323: Just wondering if the isSlotUsed helper function could be used here instead

GCHQ-Developer-847 (changes requested):
Noticed a couple more things - sorry!

GCHQ-Developer-847 (approved):

packages/web-components/src/components/ic-top-navigation/ic-top-navigation.tsx
```diff
@@ -312,17 +314,30 @@ export class TopNavigation {
...
+                  {isSlotUsed(this.el, "router-app-title") ? (
+                    <div class="title-link">
                       {this.hasAppIcon && (
                         <div class="app-icon-container" aria-hidden="true">
                           <slot name="app-icon" />
                         </div>
                       )}
                       <ic-typography variant={appTitleVariant}>
-                        <h1 class="title-wrap">{this.appTitle}</h1>
+                        <slot name="router-app-title"></slot>
```
MI6-255, line 325: Should we consider just naming the slot something like "app-title". I mention so we can be consistent with our naming and dropping the word "router". For example, Page header uses "navigation-item" slot, same as side navigation for the navigation items. Footer also uses just "link" for the same use case.

packages/web-components/src/components/ic-top-navigation/ic-top-navigation.tsx
```diff
@@ -312,17 +314,30 @@ export class TopNavigation {
             <header role="banner">
               <div class="top-panel-container">
                 <div class="app-details-container">
-                  {hasTitle && (
-                    <a class="title-link" href={this.href}>
+                  {isSlotUsed(this.el, "router-app-title") ? (
```
MI6-255, line 317: Could we remove the repeated code from the ternary if possible? Where we repeat the code for the entire icon/links.

packages/web-components/src/components/ic-top-navigation/ic-top-navigation.tsx
MI6-255, line 325: Is the expectation here then that when they use a router, they need to add the relevant heading tag like h1 themselves? May have read this incorrectly so apologies if not. I'm curious if so though, do we have a ticket where we'll be documenting on the website how to use a router with the app title, where we can make sure this specific point is mentioned?

packages/react/src/stories/ic-top-navigation.stories.mdx
```diff
@@ -476,7 +476,10 @@ export const DailyTippers = () => (
       ),
     ]}
   >
```
MI6-255, line 478: Not directly relevant to this ticket. But I presume we need a ticket to implement this same behaviour to top navigation?

MI6-255 (changes requested):

packages/web-components/src/components/ic-top-navigation/ic-top-navigation.tsx
GCHQ-Developer-530: The appTitleVariant will control the relevant heading tag, so the user won't need to change anything :)

packages/web-components/src/components/ic-top-navigation/ic-top-navigation.tsx
MI6-255: Yeah I saw that. I wondered if we could extract the code that's inside either the <a> or the <div> and pull it in? As everything inside the <a> is the same right? Unless I've read it wrong

packages/react/src/stories/ic-top-navigation.stories.mdx
MI6-255: Just to check on this, are you opening this ticket or was you wanting me too?

packages/web-components/src/components/ic-top-navigation/ic-top-navigation.tsx
MI6-255: We'll still need a ticket to show how to use app title with React router in the website though right. Let me know if this exists already or if you were creating it

GCHQ-Developer-530: I'll add a ticket :)

packages/react/src/stories/ic-top-navigation.stories.mdx
GCHQ-Developer-530: Happy to create the ticket, do you mean that we need to implement it in the guidance site top navigation?

packages/web-components/src/components/ic-top-navigation/ic-top-navigation.tsx
```diff
@@ -312,17 +320,21 @@ export class TopNavigation {
...
                           <slot name="app-icon" />
                         </div>
                       )}
                       <ic-typography variant={appTitleVariant}>
-                        <h1 class="title-wrap">{this.appTitle}</h1>
+                        {isSlotUsed(this.el, "app-title") ? (
+                          <slot name="app-title"></slot>
+                        ) : (
+                          <h1 class="title-wrap">{this.appTitle}</h1>
+                        )}
```
GCHQ-Developer-847, line 335: I think it would be worth moving the `<h1>` out of this block so it's always rendered - at the moment, if a slot is used, there would be no `<h1>` on the page so it might be confusing for screen reader users understanding the structure. This would also ensure the 'title-wrap' class is applied to the slot as well so it has the same wrapping / hyphenation as when using the app-title prop

```suggestion
                      <h1 class="title-wrap">  
                        {isSlotUsed(this.el, "app-title") ? (
                          <slot name="app-title"></slot>
                        ) : (
                          {this.appTitle}
                        )}
                      </h1>
```

GCHQ-Developer-847 (approved):

packages/react/src/stories/ic-top-navigation.stories.mdx
MI6-255: Yes please!

MI6-255 (approved):

Issue #263, opened 2023-01-10
[ic-button] Button icon slot throws off spacing if used with MUI icon


GCHQ-Developer-530:
When using the icon slot on buttons, the padding for the icon is thrown off if used with a MUI icon.

![Image](https://user-images.githubusercontent.com/111882034/219006419-57f6d8ca-936f-454e-924e-84dccfa1a087.png)

PR #379, opened 2023-02-15
fix(web-components): fix icon slots spacing for button and side nav

GCHQ-Developer-530:
## Summary of the changes
Fix icon slots for button and side nav where svg is wrapped in an element to fix issues with using mui icons.
Updated icon button story to add an example.
Updated side nav story to add an example with one of the navigation items.

## Related issue
#263 

## Checklist


packages/web-components/src/components/ic-button/ic-button.css
```diff
@@ -326,7 +326,7 @@ div.icon-container {
   align-items: center;
 }
 
-::slotted(svg) {
+::slotted(*) {
```
jd239, line 329: Isn't the MUI icon wrapped in an SVG element? Might be worth displaying the MUI and button variant in different ways (in the React storybook - then we can copy this across into the guidance site as an example):
- Parent div with the slot and nested MUI icon
- slot on the MUI icon (if possible)
- using innerHTML on a div (I think this will output the same way as the first bullet point)

..and then check if we an use a more specific selector for `slotted`. Can this also be done for side navigation please.

jd239 (commented):
Just a single comment. You may have thought is already. If so, comment on the ticket and I will approve. Cheers

packages/web-components/src/components/ic-button/ic-button.css
GCHQ-Developer-112: I found a way to import the Material Icons yesterday (see image below). I've tried importing MUI icons in our `SlottedSVG` component in past and I've struggled. I believe this is because they are [Material Design Icons (MDI)](https://pictogrammers.com/docs/library/mdi/getting-started/react/) exported as React components.

MDI, however, seem to be SVGs. Material Design has a similar component to our `SlottedSVG` component called `Icon` component and so MDI can be passed through our `SlottedSVG` component in the `path` prop in the same way as you would with their `Icon` component.

Screenshot 2023-02-16 at 14 09 40

GCHQ-Developer-530: Thank you! This is helpful for testing on a test app! I'm not sure if there's a way to do this in our story examples though? Other than if we installed mdi/js or mui icons

GCHQ-Developer-112 (approved):

GCHQ-Developer-465 (approved):

packages/web-components/src/components/ic-button/ic-button.css
jd239: > I found a way to import the Material Icons yesterday (see image below). I've tried importing MUI icons in our `SlottedSVG` component in past and I've struggled. I believe this is because they are [Material Design Icons (MDI)](https://pictogrammers.com/docs/library/mdi/getting-started/react/) exported as React components.
> 
> MDI, however, seem to be SVGs. Material Design has a similar component to our `SlottedSVG` component called `Icon` component and so MDI can be passed through our `SlottedSVG` component in the `path` prop in the same way as you would with their `Icon` component.
> 
> Screenshot 2023-02-16 at 14 09 40

So using the MUI Icon does return the SVG element hence why it doesn't work with our `slottedSVG`. `SlottedSVG` works by using the path or whatever elements are nested within the svg such as `<circle>`. Probably got an error because it was rendering 
```
<svg slot='foo'>
   <svg></svg> // MUI Icon
</svg>
```
I've created a ticket to add more guidance on this. https://github.com/mi6/ic-design-system/issues/301

Issue #322, opened 2023-01-27
icSideNavigation zIndex


MI6-613:
## Summary of the bug
Tell us, in a few words, what the problem is.

## 🪜 How to reproduce
Tell us the steps to reproduce the problem:
1. Have a page with IcSideNavigation and some MUI components (chips or data tables work) in the main content
2. Zoom in so that the icSideNavigation drawer overlays over the content
3. You will now see the MUI items appear above the drawer

## 🧐 Expected behaviour
All content is behind the expanded drawer

## 📸 Screenshots or code
If you can, add screenshots to show the problem. If you need to show a code example, you can include a snippet or link to a repository with an example.
<img width="496" alt="Screenshot 2023-01-27 at 13 25 48" src="https://user-images.githubusercontent.com/122873275/215099652-51b0a113-55e2-4135-a404-1fad8ef54fa1.png">
<img width="558" alt="Screenshot 2023-01-27 at 13 25 42" src="https://user-images.githubusercontent.com/122873275/215099763-14ec8e95-05ba-458f-a705-ec875660066b.png">

## 🖥 Desktop
 - OS: Windows & Mac
 - Browser: Chrome
 - Version: 2.1.0-beta.0

MI6-682:
Having looked at the code, I think the solution may be potentially giving the side-nav the same css variable as the ic-classification-banner (--ic-z-index-overlay)?. 

It would also be useful if in the documentation somewhere the z-index of different components was defined and explained. An example of what this could look like and explain from MUI: https://mui.com/material-ui/customization/z-index/

jd239:
> Having looked at the code, I think the solution may be potentially giving the side-nav the same css variable as the ic-classification-banner (--ic-z-index-overlay)?.
> 
> It would also be useful if in the documentation somewhere the z-index of different components were defined and explained. An example of what this could look like and explain from MUI: https://mui.com/material-ui/customization/z-index/

Ticket created: https://github.com/mi6/ic-design-system/issues/313

PR #380, opened 2023-02-15
322 icsidenavigation zindex

gd2910:
## Summary of the changes
Items and components from other DS were appearing above the expanded drawer of ic-side-nav, a new Z-index has been applied which stops this behaviour.

## Related issue
#322 

## Checklist


GCHQ-Developer-530:
I think prettier:fix needs running, but other than that looks great 👍

GCHQ-Developer-530 (dismissed):

packages/web-components/src/components/ic-side-navigation/ic-side-navigation.css
```diff
@@ -11,6 +10,7 @@
     --ic-transition-duration-slow
   );
   --side-navigation-width: 320px;
+  z-index: var(--ic-z-index-overlay);
```
jd239, line 13: Could we create a new z-index variable for side nav? This will help us get started with https://github.com/mi6/ic-ui-kit/issues/391

jd239 (changes requested):
NOICE! One comment.

packages/web-components/src/global/variables.css
```diff
@@ -279,6 +279,7 @@
   --ic-space-xxl: 3rem;
 
   /* z-index */
+  --ic-z-index-sidebar: 1200;
```
GCHQ-Developer-847, line 282: Might we want to call this '--ic-z-index-side-nav' to match the component name? Although if the purpose of creating this variable was for #391 then I wonder if something like '--ic-z-index-nav-menu' might be better as it's more generic, and we could use that for the top navigation menu as well for example when we come to that ticket

gd2910: Agreed, will rename to nav-menu

MI6-255 (approved):
Happy to approve. Would you be able to make a ticket as part of this to document z-index for consumers, similar to how it is done here: https://mui.com/material-ui/customization/z-index/

packages/web-components/src/components/ic-side-navigation/ic-side-navigation.css
```diff
@@ -11,6 +10,7 @@
     --ic-transition-duration-slow
   );
   --side-navigation-width: 20rem;
+  z-index: var(--ic-z-index-sidebar);
```
GCHQ-Developer-847, line 13: Think we just need these updating to the new name of the variable :)

gd2910: D'oh!

jd239:
Could we remove the unaffected Loki screenshots please? Since we are just changing the z-index, I don't think we need any visual regression screenshots in the PR unless I'm missing something?

jd239 (dismissed):

GCHQ-Developer-847:
Looks like the visual tests are failing. Also seems as though an old commit message is being used for the commit

GCHQ-Developer-465 (approved):

MI6-255 (changes requested):
Awaiting commit change and failing checks

Issue #177, opened 2022-12-15
[ic-radio] Horizontal option


GCHQ-Developer-530:
## Summary
There should be an option for users to show their radio buttons in a row.

## 💬 Description
For radio groups that contain only 2 options (such as 'yes' and 'no') we should provide the ability to lay out the radio buttons horizontally instead of stacked. 

But, if more than 2 radios are used in a group, the group should automatically use the stacked layout option, and not the horizontal option. 

## 💰 Use value
It gives developers more freedom and would take up less space on a page.

## Additional info
**Given** that a radio group exists with only 2 options, 
**When** setting the group's properties, 
**Then** the radio group 'orientation' should be selectable; either 'Horizontal' or 'Vertical', 
**And** if set to horizontal then the radio buttons should sit next to each other, and their width should hug the radio button's label. 

**Given** that a radio group exists with more than 2 options,
**When** the orientation prop is set to horizontal,
**Then** the radio group should still appear vertically stacked.

PR #393, opened 2023-02-20
Add horizontal option to radio buttons

GCHQ-Developer-112:
## Summary of the changes

- Audit fix the dependencies
- Add orientation prop so users have the option to have radio options laid out horizontally, given there are two or less radio options in a radio group
- Refactor CSS files so the padding is correct in the radio group & radio-option components
- Update visual tests affected by padding changes
- Refactor padding & margins in the checkbox group and checkbox components

## Related issue
#177 

## Checklist


packages/web-components/src/components/ic-radio-group/ic-radio-group.tsx
```diff
@@ -42,6 +42,11 @@ export class RadioGroup {
    */
   @Prop() required: boolean = false;
 
+  /**
+   * The orientation of the radio buttons in the radio group. If there are more than two radio buttons in a radio group, then the orientation of the radio will always be vertical.
+   */
+  @Prop() orientation: IcOrientation = "vertical";
```
ad9242, line 48: Does this need to be mutable to prevent warnings as it's potentially gets modified inside the component code?

packages/web-components/src/components/ic-radio-group/ic-radio-group.tsx
```diff
@@ -42,6 +42,11 @@ export class RadioGroup {
    */
   @Prop() required: boolean = false;
 
+  /**
+   * The orientation of the radio buttons in the radio group. If there are more than two radio buttons in a radio group, then the orientation of the radio will always be vertical.
```
ad9242, line 46:
```suggestion
   * The orientation of the radio buttons in the radio group. If there are more than two radio buttons in a radio group, then the orientation will always be vertical.
```

ad9242 (changes requested):

ad9242:
it doesn't seem to switch orientation when you add more options:

image

packages/web-components/src/components/ic-radio-group/ic-radio-group.tsx
```diff
@@ -42,6 +42,11 @@ export class RadioGroup {
    */
   @Prop() required: boolean = false;
 
+  /**
+   * The orientation of the radio buttons in the radio group. If there are more than two radio buttons in a radio group, then the orientation will always be vertical.
+   */
+  @Prop({ mutable: true }) orientation: IcOrientation = "vertical";
```
ad9242, line 48:
```suggestion
  @Prop({ mutable: true, reflect: true }) orientation: IcOrientation = "vertical";
```
think this will fix orientation switching issue

GCHQ-Developer-112: Ah thank you! I was wondering why the attribute wasn't changing

DBD324:
To answer comment above, I think that we should try to restrict the conditional field on radios that are part of a two-option horizontal group. We would run into layout difficulties here. So if a conditional field is specfified, then it should automatically revert to the vertical layout.

DBD324 (changes requested):
Small point, but in general it would be bad practice to put one option in a two option rafio group to disabled. Can we update the example in storybook to show two enabled options please? Whilst it would technically be possible to do this, this is bad practice as it forces a user into one option. If only one option is available, then the developer should consider using a different UI element to display the information.

DBD324:
> it doesn't seem to switch orientation when you add more options:
> 
> image

With regards to this comment, if more than 2 options are provided, then it should automatically revert to vertical layout.

GCHQ-Developer-112:
> > it doesn't seem to switch orientation when you add more options:
> > image
> 
> With regards to this comment, if more than 2 options are provided, then it should automatically revert to vertical layout.

This issue has been resolved now :)

packages/react/src/stories/ic-radio-group.stories.mdx
```diff
@@ -26,6 +26,29 @@ import radioGroupReadme from '../../../web-components/src/components/ic-radio-gr
...
+        >
+          <IcRadioOption
+            value="valueName2"
+            label="Selected / Default"
+            selected
+          ></IcRadioOption>
+          <IcRadioOption
+            value="valueName3"
+            label="Unselected / Disabled"
+            disabled
```
GCHQ-Developer-530, line 46: The react story also needs updating to not have a disabled option

GCHQ-Developer-530 (approved):

GCHQ-Developer-530 (dismissed):

packages/web-components/src/components/ic-radio-group/ic-radio-group.tsx
```diff
@@ -161,6 +168,29 @@ export class RadioGroup {
...
 
+    let radioHasAdditionalField = false;
+    let i = 0;
+
+    while (
+      this.radioOptions &&
+      i < this.radioOptions.length &&
+      !radioHasAdditionalField
+    ) {
+      radioHasAdditionalField = isSlotUsed(
```
ad9242, line 179: As we already know inside each radio option (via private hasAdditionalField flag), would it be cleaner to pass that information back to the radio group somehow (eg emitting an internal event) rather than duplicating the logic to do the detection?

GCHQ-Developer-112: Discussed this further with @ad9242 and the agreed approach was to combine the logic in the while loop into the existing if statement

GCHQ-Developer-530 (approved):

ad9242 (dismissed):

GCHQ-Developer-741 (dismissed):

DBD324:
Hi, when using dev tools to change the orientation from vertical to horiztonal, I still see the issue around the conditional fields not being full width. 
image
Is this just because of the way I am doing it in dev tools?

GCHQ-Developer-530 (approved):
Looks good to me 👍

DBD324 (approved):

Issue #30, opened 2022-07-26
Toasts


ASM995:
**Component Description** : 
Toasts give brief updates about events that happen in an application. Sometimes called 'snackbars' these messaging components  provide non-interruptive feedback to user after completing an action. They can clear themselves automatically, or persist until dismissed. 

💰Business Value💰
Foundational component required to provide ICDS users the tools to build rich applications. Toasts provide important method of display temporary messaging to users based on a user action. 

📚User Story 📚 
**As an ICDS user** 
**I need to** be able to define a toast as temporary or permanent
**So that** the toast can either automatically dismiss itself, or remain on screen requiring user attention. 

**As an ICDS user** 
**I need** a temporary toast to dismisses itself after specified timeout period
**So that** its visible duration can be set dependent on its context. 

**As an ICDS user**
**I need** a permanent toast to remain on screen until dismissed and it must have a method of being dismissed
**So that** it can be cleared from the screen manually. 

**As an ICDS user**
**I need** to be able to define a toast as dismissible
**So that** a user can manually clear it from their screen.

**As an ICDS user**
**I need** these interaction states for the dismiss action of a toast:
Default
Hover
Pressed
Focussed
**So that** visual feedback is provided to user interacting with the control. 

**As an ICDS user**
**I need** to include an optional single action (tertiary button) on a toast
**So that** a user is provided with a contextual action.

**As an ICDS user**
**I need** to have the following toast variants:
Default (grey)
Info (blue)
Error (red)
Warning (orange)
Success (green)
**So that** there are multiple options to chose from for creating a UI. 

**As an ICDS user**
**I need** to have the option for a single line or multi line toast variant
**So that** I can fit longer messages in if required. 

📝Acceptance Criteria 📝

**Interaction behaviour**
**Given** that a dismissible toast is displayed on screen, 

**When** interacting with the dismiss icon button, 

**Then** the icon button interaction states should display.     

**Given** that a dismissible toast is displayed on screen, 

**When** using keyboard to interact with it, and focus is within the toast

**Then** the toast can be dismissed using the escape key.   

**Given** that a dismissible toast is displayed on screen, 

**When** using keyboard to interact with it, and focus is outsde of the toast

**Then** the toast can only be dismissed using the escape key if no other elements need dismissing first. 

(tooltips, dropdowns etc should dismiss with escape key before toast.   

**Content behaviour**
**Given** that a toast is displayed,

**When** a second (or additional) toast needs to displayed., 

[...]

PR #422, opened 2023-02-22
Toast component

GCHQ-Developer-741:
## Summary of the changes
Cloned PR for toast due to rebasing issue on previous PR.

## Related issue
#30 

## Checklist


GCHQ-Developer-741:
On the whole, it looks great. I had a couple of questions however:

With regards to focus, when a toast is displayed we trap focus on the toast element, but you can still use pointer to move focus to something on the base layer. I'm not sure that trapping focus is correct. As the toast doesn't behave modally (you can still interact with the background layer) I would imagine focus would continue to move to other elements on the page. The toast would be injected into the focus order immediately after the triggering element.

The aria dialog spec says this:

"Dialogs can be modal or non-modal. When a modal dialog appears on the screen, it's not possible to interact with any page content outside the dialog. In other words, the main application UI or page content is considered to be temporarily disabled as long as the modal dialog is showing. For non-modal dialogs it is still possible to interact with content outside of the dialog while the dialog is showing. Note that for non-modal dialogs there will have to be a global keyboard shortcut that allows focus to be moved between opened dialogs and the main page."
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role

When a non-modal dialog is displayed, we should move focus to the first focusable element within, but should we also allow focus to move out of the non-modal dialog? Could we provide a mechanism (landmark?) for moving between base layer content and dialog content?

Another option would be to automatically dismiss the non-modal dialog (toast) when a user interacts with an element outside of the toast. See this example from deque univesity. 
https://dequeuniversity.com/library/aria/tooltip-dialog

Inconveniently, although the WAI-ARIA authoring practices repeatedly reference non-modal dialogs being an option, there doesn't seem to be specific practices around them.
https://w3c.github.io/aria-practices/#dialog_modal

I think we should double check with the a11y team on what they suggest the best approach here is.

GCHQ-Developer-741:
Screenshot from 2023-02-22 16-35-58
Screenshot from 2023-02-22 16-35-51

GCHQ-Developer-530 (commented):
Looks like the commit messages might need changing as currently the changelog won't describe that there's a new toast component, it'll look more like different issues have been fixed.

GCHQ-Developer-741:
> Similar to @ad9242's [comment for the dialogs PR](https://github.com/mi6/ic-ui-kit/pull/438#issuecomment-1443964805), I'm not sure the visual tests are adding much value. Might be worth investigating if there's a way to show the toast in the visual tests or remove them?

Have removed them from the loki tests

GCHQ-Developer-741:
> Looks like the commit messages might need changing as currently the changelog won't describe that there's a new toast component, it'll look more like different issues have been fixed.

Sorted 😌

MI6-451 (approved):

DBD324 (commented):
For the neutral status version, it would be good to make it so that if a user forgets to specify an icon, no icon is displayed rather than the default. Worried the default ? icon being used as a placeholder could be confusing at it symbolizes a question. If a user doesn't override the icon, can we hide it and have the text move over to fill the width of the toast?

DBD324:
On the back of the a11y testing session conducted last week, did we manage to get the auto-dismiss toast contents to be announced as a live region? It wasn't doing this when tested, but not sure if this was a browser issue.

GCHQ-Developer-741:
> For the neutral status version, it would be good to make it so that if a user forgets to specify an icon, no icon is displayed rather than the default. Worried the default ? icon being used as a placeholder could be confusing at it symbolizes a question. If a user doesn't override the icon, can we hide it and have the text move over to fill the width of the toast?

Addressed

DBD324 (dismissed):

MI6-451 (approved):

DBD324 (approved):

MI6-451 (approved):

Issue #437, opened 2023-02-24
[all] Investigate use of delegatesFocus for interactive components


GCHQ-Developer-530:
## Summary
Investigate adding `shadow: {
delegatesFocus: true
}` for components to remove our custom setFocus methods and just call .focus() on the components instead.

## 💬 Description
We can add delegatesFocus to the `@Component()` decorator at the top of each component like so:
```
@Component({
  tag: "ic-component",
  styleUrl: "ic-component.css",
  shadow: {
    delegatesFocus: true
  },
})
```
This seems to work for less involved components, such as radio option, where calling `this.host.querySelectorAll("ic-radio-option")[0].focus()` focuses the radio option without need for the setFocus method. But it seems to cause issues for more complicated setFocus usage, such as the button. An example of this is that the close menu button in top nav focuses on menu open for the current implementation:

<img width="318" alt="Screenshot 2023-02-24 at 12 56 00" src="https://user-images.githubusercontent.com/111882034/221184866-96ebdb5b-d57b-40bc-aa4c-27eeb1252f84.png">

But after adding delegatesFocus to the relevant components, and removing setFocus methods in favour of calling the standard javascript .focus() method on the component, the close menu button doesn't gain focus on menu open, and can't be tabbed to to force focus:

<img width="324" alt="Screenshot 2023-02-24 at 12 55 37" src="https://user-images.githubusercontent.com/111882034/221184903-c06a8733-011f-4572-af15-c03a90c9f851.png">

The work for this ticket is therefore to first investigate if this is possible. Then (if it is) update interactive components to use delegateFocus: true, and remove setFocus methods. 

## 💰 Use value
It will make it easier to call focus on our own components, and also reduce code by removing our custom focus methods.

Issue #238, opened 2023-01-09
Dialogs


GCHQ-Developer-530:
## Summary
Dialogs are modal windows that provide important messaging to users. As they are presented modally, they interrupt the user's flow to display vital information. The user must interact with the dialog in order to continue. This is usually by performing the primary action provided in a modal, or dismissing the modal. 
_Please ensure modal shadow is applied_ 

https://web.dev/building-a-dialog-component/

The modal container provides the background and behaviour for content that needs to be displayed modally. Any other content can be added to the container and the mechanism of the modal will be consistent with all others. 

## 💰 User value
Foundational component required to provide ICDS users the tools to build rich applications. Modal containers provide a method to house content in a modal layer above the context of the base layer. 

## 📚 User Stories
If relevant, describe the high-level functionality as user stories.

**As an** ICDS user
**I need**  a default dialog that contains:
Title
Description
Dismiss button (X button)
Primary action button
Dismiss action button
So that I can display important messages to users. 

**As an** ICDS user
**I need** an information dialog
**So that** I can simply provide information that can then be closed.  

**As an** ICDS user
**I need** to be able to select a status (success, warning, error, information) to display on the dialog
**So that** i can indicate the severity of the message. 

**As an** ICDS user
**I need** to be able to choose the dialog size (small, medium, large) 
**So that** I can add longer messages. 

**As an** ICDS user
**I need** to be able to enter markdown into the dialog body
**So that** I can add different types of content. 

**As an** ICDS user
**I need** to be able to add a child element to the dialog body
**So that** I can add interactive content. 

**As an** ICDS user
**I need** to have the content area scroll when the maximum vertical modal size is reached
**So that** longer content can be viewed in the dialog. 

**As an** ICDS user
**I need** a modal container to appear above the background layer
**So that** I can house other content.

**As an** ICDS user
**I need** a dark overlay to appear over the background content, but behind the modal container
**So that** the modal is differentiated from the background layer. 

**As an** ICDS user
**I need** the following different sizes of modal container:
Small (4 columns?)
Medium (6 columns?)
Large (8 columns?)
**So that** i can add different amounts of content. 

**As an** ICDS user
**I need** to be able to dismiss the modal using a control in the container content, as well as by pressing the modal background overlay
**So that** I can quickly close the modal content and return to the base layer. 

**As an** ICDS user
**I need** to ensure that only one modal is displayed at a time
**So that** modal layers do not stack.  

**As an** ICDS user
[...]

Issue #173, opened 2022-12-14
[ic-alert] default - allow icon to be changed


GCHQ-Developer-465:
## Summary of the bug
A prop needs to be passed in for the icon. An example of a use case is attached below.

## 🧐 Expected behaviour
The Icon in default Alert should be changeable, so that users can amend it to whatever they like. The other alert types (info, success, warning, and error) all use specific icons to relate to the status of the alert, but the default one should be customisable by the user so that it is flexible in many different scenarios. 

## 📸 Screenshots or code
![Screenshot 2022-12-14 at 16 01 26](https://user-images.githubusercontent.com/111882114/207645768-b0443fc4-19f1-42fc-b754-732ddadadf62.png)

## Additional info
**AC** 
Given that a user is implementing a default type alert, 

When settings its props, 

Then they should be able to specify a different icon,

and they should be able to specify no icon.

MI6-255:
Add a slot to replicate design behaviour, to only be allowed to be done on "default", not other states. Include guidance to only include on default and add a code example on the site

PR #438, opened 2023-02-24
Feature/238 dialog

MI6-451:
## Summary of the changes
Adds the dialog component and associated tests and stories. AC requires the embedded alert component to allow for a different icon to be slotted in the neutral variant however this not currently implemented on the main alert component. Once work is completed on this the dialog will be updated to allow for the same functionality. Due to current issues setting focus on ic components only the 'default' variants of dialog implement focus traps and initial focus setting. When 'interactive' content is slotted the 'default' focus behaviour (focus trap, initial focus) must be disabled or it will not work correctly. Therefore this work is blocked while investigation into handling focus on ic components is ongoing. 

## Related issue(s)
Main ticket: #238 
AC blocked by: #173 
Blocked by: #437, #372 

## Checklist


ad9242:
re the visual tests: There doesn't seem much point adding these as they just capture what the buttons look like. Think it should either:
have stories with the dialogs already loaded
or
turn off the visual tests for these

MI6-451:
> re the visual tests: There doesn't seem much point adding these as they just capture what the buttons look like. Think it should either: have stories with the dialogs already loaded or turn off the visual tests for these

Agreed, can probably do the same for the tooltip stories as well as they are also just buttons in the captured images.

packages/web-components/src/components/ic-dialog/ic-dialog.tsx
```diff
@@ -0,0 +1,371 @@
...
+})
+export class Dialog {
+  /**
+   * @slot - Content will be placed in the middle of the dialog, between the heading and the controls.
+   * @slot dialog-controls - Content will be place at the bottom of the dialog.
+   */
+  @Element() el: HTMLIcDialogElement;
+
+  /**
+   * Sets the maximum and minimum height and width for the dialog.
```
GCHQ-Developer-530, line 29: Do we need to add something to explain that for xs breakpoint, these have no effect and the dialog is always full screen? Or is that something that should be mentioned in the guidance?

MI6-451: I agree its definitely worth mentioning, not sure whether it would be better as guidance or mentioned somewhere within the component.

ad9242:
found an issue when having other content on the page: if you have a side nav, it overlays the dialog & you can still interact with it:

image

ad9242:
also an issue with slotting a ic-select on there: 

image

GCHQ-Developer-465 (commented):
I think the px values need to be changed to rem in this file: 'packages/web-components/src/components/ic-dialog/ic-dialog.css'. This is to keep up with consistency with the rest of our components now using rem.

GCHQ-Developer-465:
On slotted content the checkbox is automatically focused upon. This should not be the case? 
Screenshot 2023-03-09 at 15 57 28

GCHQ-Developer-465:
Perhaps having a scrollable dialog would be good to have too as a story? (From looking at designs)

MI6-451:
> On slotted content the checkbox is automatically focused upon. This should not be the case should it? Screenshot 2023-03-09 at 15 57 28

As in the AC, if users slot interactive elements in the content area then the first of these elements should initially receive focus. This is controlled by adding a 'focus' id to the relevant element so users could bypass this if they wanted and allow the focus to go to the default buttons or, if they are using slotted buttons, set focus on their own buttons.

GCHQ-Developer-465:
Overall, this is a great piece of work for the dialog component! Awesome work 👍

packages/web-components/src/components/ic-dialog/ic-dialog.css
```diff
@@ -0,0 +1,131 @@
...
+
+.close-icon-svg {
+  display: flex;
+  align-items: center;
+}
+
+.content-area {
+  overflow-y: scroll;
+  -ms-overflow-style: none;
+  scrollbar-width: none;
```
ad9242, line 86:
```suggestion
  scrollbar-width: none;
  padding: var(--ic-space-xs);
  margin: calc(-1 * var(--ic-space-xs));
```
I think if you add the above padding & margins, it should allow for focus indicator to appear for slotted elements (eg textfield, checkbox, radio etc) without being cropped. You would then not need to set margins on the checkbox group in the slotted story

packages/web-components/src/components/ic-dialog/ic-dialog.tsx
```diff
@@ -0,0 +1,385 @@
...
+      this.el.shadowRoot.querySelectorAll("ic-button")
+    );
+    const slottedInteractiveElements = Array.prototype.slice.call(
+      this.el.querySelectorAll(
+        `a[href], button, input, textarea, select, details, [tabindex]:not([tabindex="-1"]), 
+        ic-button, ic-checkbox, ic-select, ic-search-bar, ic-tab-group, ic-radio-group, 
+        ic-back-to-top, ic-breadcrumb, ic-chip[dismissible="true"], ic-footer-links, ic-link, ic-navigation-button, 
+        ic-navigation-item, ic-switch`
+      )
+    );
```
jd239, line 209: Is the `slice` array method required here? It looks like we just want to put a list of items into an array?

packages/web-components/src/components/ic-dialog/ic-dialog.tsx
```diff
@@ -0,0 +1,385 @@
...
+
+    let focusedElement;
+
+    if (this.el.querySelector("#focus") !== null) {
+      focusedElement = this.el.querySelector("#focus") as HTMLElement;
+    } else {
+      focusedElement = this.el.shadowRoot.querySelector(
+        "#focus"
+      ) as HTMLElement;
+    }
```
jd239, line 173: For this to work with slotted content, doesn't there need to be a focus ID on the slotted interactive element (which I imagine would be unlikely)?

packages/web-components/src/components/ic-dialog/ic-dialog.tsx
```diff
@@ -0,0 +1,385 @@
...
+        (this.interactiveElementList[i] as HTMLElement) ===
+        (this.el.shadowRoot.activeElement || document.activeElement)
+      ) {
+        this.focusedElementIndex = i;
+      }
+    }
+  };
+
+  private setAlertVariant = () => {
+    if (this.status !== undefined) {
```
jd239, line 192: Do we need to do a check for null as well?

packages/web-components/src/components/ic-dialog/ic-dialog.tsx
```diff
@@ -0,0 +1,385 @@
...
+   */
+  @Prop() alertHeading?: string;
+  /**
+   * If a status is set, sets the message for the displayed alert.
+   */
+  @Prop() alertMessage?: string;
+  /**
+   * Sets the number of default buttons to be displayed at the bottom of the dialog.
+   */
+  @Prop() buttons?: "none" | "one" | "two" | "three" = "two";
```
jd239, line 59: Should the number of buttons be based on the number of buttonProps (with a max of 3)? That way, every button prop is account for. There will never be an instance where the buttons prop above is 1 but there are 3 buttonProps.

A .`map` on line https://github.com/mi6/ic-ui-kit/pull/438/files#diff-262ef4802b44b3bf5d24bf0350b75541588fc233edfa4ff24bb846bc3330670fR328 would do the work for you..

packages/web-components/src/global/variables.css
```diff
@@ -279,7 +279,9 @@
   --ic-space-xxl: 3rem;
 
   /* z-index */
+  --ic-z-index-base-value: 1000;
```
jd239, line 282: Is this for testing or meant to overwrite the previous z-index ticket? https://github.com/mi6/ic-ui-kit/pull/471

jd239 (changes requested):
NEW COMPONENT HYPE! 🔥 
Added some comments.

packages/web-components/src/global/variables.css
ad9242: now #471 is merged, can rebase & use the new properties/values

packages/web-components/src/components/ic-dialog/ic-dialog.tsx
```diff
@@ -0,0 +1,385 @@
...
+   * If set to 'true' the dialog will not close when the scrim is clicked.
+   */
+  @Prop() closeOnClick?: boolean = true;
+  /**
+   * If default buttons are displayed, sets the label and onclick function passed as [{label: string, onclick: string},{},{}] with empty objects for unused buttons.
+   */
+  @Prop() buttonProps?: { label: string; onclick: string }[] = [
+    { label: "Confirm", onclick: "this.icDialogConfirm()" },
+    { label: "Cancel", onclick: "this.hideDialog()" },
+    { label: "", onclick: "" },
```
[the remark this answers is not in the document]
MI6-451: Previously it was required to pass three objects even if fewer than three default buttons were required and any unwanted buttons were passed as empty objects. The default button behaviour, including the method of receiving and using props, has been reworked so this is no longer the case and the redundant item has been removed from the default value.

packages/web-components/src/components/ic-dialog/ic-dialog.tsx
```diff
@@ -0,0 +1,385 @@
...
+
+  @State() fadeIn: boolean = false;
+
+  /**
+   * Confirmation event for default primary button behaviour.
+   */
+  @Event() icDialogConfirmed: EventEmitter<void>;
+  icDialogConfirm = () => {
+    this.icDialogConfirmed.emit();
+  };
```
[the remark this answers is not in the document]
MI6-451: Added a cancel event and set the default 'Cancel' button behaviour to emit the event in addition to closing the dialog.

packages/web-components/src/components/ic-dialog/ic-dialog.tsx
MI6-451: The intention here was that users would set an id of 'focus' on whatever element they wanted to receive initial focus when using slotted content and that this would override/come before the default buttons which had 'focus' set by the component. 

Reworked this so the component now automatically adds a 'data-gets-focus' ID on whatever element should receive focus initially, including slotted content.

packages/web-components/src/components/ic-dialog/ic-dialog.tsx
```diff
@@ -0,0 +1,385 @@
...
+    this.dialogRendered = true;
+    setTimeout(() => {
+      this.fadeIn = true;
+    }, 10);
+    setTimeout(() => {
+      this.setInitialFocus();
+    }, 75);
+    setTimeout(() => {
+      this.getFocusedElementIndex();
+    }, 80);
```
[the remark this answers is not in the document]
MI6-451: I think it's worth investigating as its also used on Toasts and similar functionality may be needed later for new components or to add animation to older components. I have opened #476 to look into the issue.

packages/web-components/src/components/ic-dialog/ic-dialog.css
```diff
@@ -0,0 +1,141 @@
...
+  margin-top: var(--ic-space-xs);
+}
+
+.message {
+  margin-top: var(--ic-space-xs);
+}
+
+.dialog-controls {
+  margin-top: auto;
+  margin-bottom: var(--ic-space-xxs);
```
GCHQ-Developer-847, line 104: Looks like the gap underneath the buttons is slightly larger than the 8px in the designs because of this margin:

packages/web-components/src/components/ic-dialog/ic-dialog.tsx
```diff
@@ -0,0 +1,403 @@
...
+        return this.destructive ? "destructive" : "primary";
+      }
+    } else {
+      if (index === 2) {
+        return this.destructive ? "destructive" : "primary";
+      } else {
+        return "secondary";
+      }
+    }
+  };
```
GCHQ-Developer-847, line 289: Wonder if you could create a variable to avoid repeating the `this.destructive` check, e.g. something like this (probably not completely necessary but up to you):
```suggestion
  private getButtonVariant = (index: number) => {
    const mainVariant = this.destructive ? "destructive" : "primary";
    if (this.buttonProps.length === 1) {
      return mainVariant;
    } else if (this.buttonProps.length === 2) {
      if (index === 0) {
        return "tertiary";
      } else {
        return mainVariant;
      }
    } else {
      if (index === 2) {
        return mainVariant;
      } else {
        return "secondary";
      }
    }
  };
```

GCHQ-Developer-847 (changes requested):
Looks really good :) just a few small things from me

jd239 (dismissed):
Once @GCHQ-Developer-847's comments are addresses, we have to remember to change the branch target to `a11y/dialog`

packages/web-components/src/components/ic-dialog/ic-dialog.tsx
```diff
@@ -0,0 +1,403 @@
...
+
+  private getInteractiveElements = () => {
+    this.interactiveElementList = Array.from(
+      this.el.shadowRoot.querySelectorAll("ic-button")
+    );
+    const slottedInteractiveElements = Array.from(
+      this.el.querySelectorAll(
+        `a[href], button, input, textarea, select, details, [tabindex]:not([tabindex="-1"]), 
+        ic-button, ic-checkbox, ic-select, ic-search-bar, ic-tab-group, ic-radio-group, 
+        ic-back-to-top, ic-breadcrumb, ic-chip[dismissible="true"], ic-footer-links, ic-link, ic-navigation-button,
```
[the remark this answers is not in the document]
MI6-451: Because while the child components, tab and radio option, are technically the focusable elements they cannot be navigated to with tab. Instead we tab to the tab group or radio group and then use keyboard controls to navigate within those, this replicates that behaviour rather than allowing/forcing tabbing between radio options or tabs. The checkbox element itself can be tabbed to within a checkbox group, however, so that should be included in the interactive elements as opposed to the parent.

packages/web-components/src/components/ic-dialog/ic-dialog.tsx
MI6-451: Good idea, I have added this

packages/web-components/src/components/ic-dialog/ic-dialog.tsx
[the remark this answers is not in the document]
GCHQ-Developer-847: Ah makes sense - thanks

jd239 (approved):

Issue #326, opened 2023-01-31
[ic-select]: add icInput event, disableFilter and debounce onto ic-select when searchable prop is set


jd239:
## Summary
At the moment, Select with the `searchable` prop turns the `ic-select` into an autocomplete component which should behave similarly to SearchBar. This is not the case.

## 💬 Description
ic-select with the `searchable` prop should have similar functionality/events as SearchBar. The missing events/props are:
- icOptionSelect (event)
- disableFilter (prop)
- debounce (prop)
- icInput (prop)

## 💰 Use value
Will enrich the searchable select functionality and cover more use cases for behaviour expected from an autocomplete component. 

## Additional info
It may be possible to lift and shift what is currently in SearchBar. One thing to note is that SearchBar uses IcTextField whereas Select does not so props/events like debounce will need to be implemented differently.

PR #439, opened 2023-02-24
Add icInput event, disableFilter and debounce onto ic-select when searchable prop is set

ad9242:
## Summary of the changes
ic-select:
Adds support for filtering options externally in order to match functionality provided by ic-search-bar component
Adds debounce prop which delays firing of icChange event
Adds unit test coverage for changed code
Adds storybook examples to show external filtering

ic-menu:
Renames some internal events in ic-menu to remove "ic" prefix
Replaces Listen decorator with target of document with events bound to parent element, to avoid too many event handlers running

ic-search-bar
Removes Listen decorators from ic-search-bar as already assigning callback functions. Moves previous Listen code into those functions. Also updates unit tests to reflect changes

## Related issue
#326  

## Checklist


packages/web-components/src/components/ic-select/ic-select.tsx
```diff
@@ -299,6 +367,11 @@ export class Select {
     this.searchable && this.handleFocusIndicatorDisplay();
   };
 
+  private handleMenuKeyPress = (ev: CustomEvent): void => {
+    const debounce = ev.detail.isNavKey === true ? 0 : this.debounce;
```
jd239, line 371:
```suggestion
    const debounce = ev.detail.isNavKey ? 0 : this.debounce;
```

Can we also add a comment above this so developers know it's to prevent lagging due to the debounce..

packages/web-components/src/components/ic-select/ic-select.tsx
```diff
@@ -316,9 +389,16 @@ export class Select {
     }
   };
 
+  private isExternalFiltering = (): boolean =>
+    this.searchable && this.disableFilter;
+
   private handleClick = (event: MouseEvent): void => {
-    this.noOptions = null;
-    this.menu.options = this.options;
+    if (this.isExternalFiltering()) {
```
jd239, line 396: Should filteredOptions be added to options if `searchable` is true as well?

packages/web-components/src/components/ic-select/ic-select.tsx
```diff
@@ -354,22 +435,26 @@ export class Select {
...
-      if (
-        !(
-          (event.key === "ArrowDown" || event.key === "ArrowUp") &&
-          this.noOptions !== null
-        ) &&
-        this.isMenuEnabled()
-      ) {
+      if (!(isArrowKey && this.noOptions !== null) && this.isMenuEnabled()) {
+        if (this.isExternalFiltering() && isArrowKey) {
+          if (this.debounce > 0) {
```
jd239, line 454: This check can probably be added to the line above

packages/web-components/src/components/ic-select/ic-select.tsx
```diff
@@ -385,9 +470,14 @@ export class Select {
 
   private handleSearchableSelectInput = (event: Event): void => {
     this.searchableSelectInputValue = (event.target as HTMLInputElement).value;
+    this.icInput.emit({ value: this.searchableSelectInputValue });
 
-    if (this.getValueFromLabel(this.searchableSelectInputValue) === undefined) {
-      this.icChange.emit({ value: null });
+    if (this.disableFilter === true) {
```
jd239, line 475:
```suggestion
    if (this.disableFilter) {
```

packages/web-components/src/components/ic-select/ic-select.tsx
```diff
@@ -385,9 +470,14 @@ export class Select {
...
+      this.emitIcChange(this.searchableSelectInputValue);
+    } else if (
+      this.getValueFromLabel(this.searchableSelectInputValue) === undefined
+    ) {
+      this.emitIcChange(null);
```
jd239, line 480: Don't we still want to emit the value even if there is no matching value form the list? The value has still changed.

packages/web-components/src/components/ic-select/ic-select.tsx
```diff
@@ -396,60 +486,62 @@ export class Select {
       this.setMenuChange(false);
     }
 
-    const options = [...this.options];
+    if (this.disableFilter === false) {
```
jd239, line 489:
```suggestion
    if (!this.disableFilter) {
```

jd239 (changes requested):
🔥 !!! 
A few comments. Let me know if you want to go through them together.

packages/web-components/src/components/ic-select/ic-select.spec.ts
```diff
@@ -1088,6 +1091,110 @@ describe("ic-select", () => {
...
+    expect(page.rootInstance.filteredOptions).toHaveLength(1);
+    expect(page.rootInstance.filteredOptions[0].label).toEqual(
+      "No results found"
+    );
+
+    expect(page.rootInstance.open).toBe(true);
+
+    input.click();
+    await page.waitForChanges();
+    expect(page.rootInstance.open).toBe(false);
```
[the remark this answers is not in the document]
ad9242: The behaviour is unchanged: i.e. clicking in the input should toggle the open\close state of the menu. This was test snippet added to increase code coverage

packages/web-components/src/components/ic-menu/ic-menu.tsx
```diff
@@ -184,12 +186,12 @@ export class Menu {
   };
 
   /**
-   * If menu is opened with the mouse, emit icMenuStateChange custom event.
+   * If menu is opened with the mouse, emit menuStateChange custom event.
```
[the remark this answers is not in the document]
ad9242: opted to make the handleClickOpen method an internal method

packages/web-components/src/components/ic-select/ic-select.tsx
ad9242: Leaving the behaviour as is, as this would be a breaking change. If it's something requested in future we can add under a new prop

jd239 (approved):

Issue #275, opened 2023-01-10
Check and review disabled states for different backgrounds


GCHQ-Developer-530:
## Summary

Check the disabled states for the below buttons, ensuring they match the ACs. If not, make the necessary updates.

### Default or dark buttons
**Given** that a default or dark coloured, disabled, primary button exists on a page, 
**When** viewed, 
**Then** its background colour should use 20% darken (rgba(0,0,0,0.2)), 
**And** its foreground colour should use 20% darken (rgba(0,0,0,0.2).

**Given** that a default or dark coloured, disabled, secondary button exists on a page, 
**When** viewed, 
**Then** its border colour should use 20% darken (rgba(0,0,0,0.2)), 
**And** its foreground colour should use 20% darken (rgba(0,0,0,0.2).

**Given** that a default or dark coloured, disabled, tertiary button exists on a page, 
**When** viewed, 
**Then** its foreground colour should use 20% darken (rgba(0,0,0,0.2).

**Given** that a default or dark coloured, disabled, icon button exists on a page, 
**When** viewed, 
**Then** its foreground colour should use 20% darken (rgba(0,0,0,0.2).

### Light buttons
**Given** that a light coloured, disabled, primary button exists on a page, 
**When** viewed, 
**Then** its background colour should use 20% lighten (rgba(255,255,255,0.2)), 
**And** its foreground colour should use 20% lighten (rgba(255,255,255,0.2).

**Given** that a light coloured, disabled, secondary button exists on a page, 
**When** viewed, 
**Then** its border colour should use 20% lighten (rgba(255,255,255,0.2)), 
**And** its foreground colour should use 20% lighten (rgba(255,255,255,0.2).

**Given** that a light coloured, disabled, tertiary button exists on a page, 
**When** viewed, 
**Then** its foreground colour should use 20% lighten (rgba(255,255,255,0.2).

**Given** that a light coloured, disabled, icon button exists on a page, 
**When** viewed, 
**Then** its foreground colour should use 20% lighten (rgba(255,255,255,0.2).

### Destructive buttons
**Given** that a disabled, destructive button exists on a page, 
**When** viewed, 
**Then** its background colour should use 20% darken (rgba(0,0,0,0.2)), 
**And** its foreground colour should use 20% darken (rgba(0,0,0,0.2).

MI6-94:
Additional info:
- Changing colours to up contrast without making button seem 'enabled'.
[...]

Issue #242, opened 2023-01-09
[ic-navigation-group] - Horizontal scroll behaviour


GCHQ-Developer-465:
## Summary
On smaller screens, it may be possible that not all nav items can be displayed on the screen. In this case we need a way to view and access all nav items so that the content can still be navigated. 

To maintain consistency in operation regardless of interaction mode, a side scrolling behaviour including additional buttons is used. 
### **User story**
**As a** user

**I need** a way to display large numbers of nav items in small containers

**So that** I can view and interact with every nav item in a nav bar. 

## 💰 Use value

**### Acceptance criteria**

- [ ] Given that a nav bar exists on any breakpoint, when the available width is not great enough to display all nav items, then the nav bar responsive behaviour should be displayed allowing all nav items to be viewed. 

 

- [ ] Given that a responsive nav bar is being displayed, when the nav bar is scrolled fully to the left (showing the first nav item), then the scroll left button should be disabled. And the scroll right button should be enabled.  

 

- [ ] Given that a responsive nav bar is being displayed, when the nav bar is scrolled fully to the right (showing the last nav item), then the scroll right button should be disabled, and the scroll left button should be enabled.  

 

- [ ] Given that an enabled scroll button is displayed, when clicked once, then it should scroll the nav bar so that one additional nav item comes into view. And it should animate the position change with a duration of 100ms and a ease-in-out curve. 

 

- [ ] Given that an enabled scroll button is displayed, when clicked and held, then it should scroll the nav bar to the farthest position. And it should animate the position change with a duration of 300ms and a ease-in-out curve. 

 

- [ ] Given that a responsive nav bar exists, when the nav bar is focussed and focus on individual nav items is changed using the tab key, then the nav bar should automatically scroll to allow the currently focussed nav item to be visible.  

 

- [ ] Given that a responsive nav bar exists, when clicking and selecting a nav item that is not fully in view, then once the click action is registered, then that nav item should be automatically scrolled into view. And the scroll buttons should become enabled/disabled if the nav bar is scrolled fully one way or another. 

 

- [ ] Given that a responsive nav bar exists, when tabbing through page content, then the order should be:

- Tab to nav bar first item 
- Tab to nav bar second item
- ...
- Tab to nav bar last item
- Tab out of tab bar. 

 

[...]

PR #440, opened 2023-02-26
Feature/242 ic navigation group horizontal scroll

GCHQ-Developer-847:
## Summary of the changes
- Moved scrolling functionality out of tabs and into horizontal scroll component - which has then been added to the tabs and nav items (in top nav and page header)
- Created 'appearance' prop to allow for dark-, light- or default-themed horizontal scroll, and converted scroll buttons to ic-buttons to make use of their styling
- Added 'mouseleave' event listener to prevent long scroll continuing if user moves cursor off button
- Adjusted sizing of nav items to match designs, and also so scroll buttons are aligned with nav item text 
- Added more tabs and nav items to stories for demonstrating scroll behaviour
- Removed tab group tests and added horizontal scroll tests

The icons on the scroll buttons when disabled currently use the keyline-lighten and keyline-darken colour tokens to match the colours used in the Figma designs. This styling should be removed as part of #275 so they will inherit the updated button icon colours.

## Related issue
#242 

## Checklist


ad9242:
Should "Home" and "End" keys work for going to first & last items? They do for tabs (it's handled in the tab-context), but not currently for navigation items

ad9242:
Seem to be some errors re accessibility in Storybook. On Tabs, Page Header with tabs & Top nav with the horizontal scroll

packages/web-components/src/components/ic-horizontal-scroll/ic-horizontal-scroll.css
```diff
@@ -0,0 +1,98 @@
...
+
+.disabled .ic-tooltip-container {
+  display: none !important;
+}
+
+.scroll-splitter-left,
+.scroll-splitter-right {
+  display: flex;
+  height: var(--ic-space-lg);
+  width: 0.063rem;
```
ad9242, line 72: Seem to be a lot of places that 0.0.63rem is used (not just in this PR). Should we consider a separate ticket to make this a css variable?

GCHQ-Developer-847: Yep I think that makes sense - I will make it :)

GCHQ-Developer-847:
> Should "Home" and "End" keys work for scrolling to beginning & end? For tabs they focus the items too, but that's not required for nav items from what I can see from the Acceptance criteria

@ad9242 I think because tabs use a different method of navigation (using the left / right arrows as opposed to tabbing), it would be unlikely for the user expect the Home and End keys to work the same way as the tabs.

I couldn’t find any top navigation on other sites with the same scrolling behaviour as ours but I would assume if we did want the Home and End keys to move focus when the scroll buttons are visible, we would also want them to work when the scroll buttons aren’t shown - and this doesn’t happen on other sites I’ve looked at like on [GOV.UK](https://design-system.service.gov.uk/) and [W3C](https://www.w3.org/WAI/).

So I think it’s acceptable not to have this behaviour?

GCHQ-Developer-847:
> Seem to be some errors re accessibility in Storybook.
> 
> Tabs: image
> 
> Page Header with tabs: image
> 
> Top nav with the horizontal scroll: image

@ad9242 the accessibility checks in Storybook seem to be a bit temperamental - if I refresh the page when viewing these stories, these errors just disappear? So I'm not sure they're completely reliable.

I have added an a11y test for the ic-horizontal-scroll which should capture any issues, but I'm not getting any accessibility failures from it - or from anything else (neither the top navigation, page header or tabs). 

I have also checked with a screen reader and when I focus on the tabs / nav items, the correct information is still being read out (same as on the develop branch - so a screen reader user can navigate through the items without knowing that the scroll buttons exist which I believe is what we want).

I have checked the accessibility trees in the DevTools and these are all as expected.

packages/web-components/src/components/ic-page-header/ic-page-header.tsx
```diff
@@ -227,12 +227,14 @@ export class PageHeader {
...
-                    aria-label="navigation-landmark-page-header"
-                  >
-                    <slot name="tabs" />
-                  </nav>
+                  <ic-horizontal-scroll>
+                    <nav
+                      class="tabs-slot"
+                      aria-label="navigation-landmark-page-header"
+                    >
+                      <slot name="tabs" />
```
ad9242, line 235: Do we need to add <ul> here to fix a11y errors with list-item role on child items? We do for top navigation etc

ad9242:
> > Seem to be some errors re accessibility in Storybook.
> > Tabs: image
> > Page Header with tabs: image
> > Top nav with the horizontal scroll: image
> 
> @ad9242 the accessibility checks in Storybook seem to be a bit temperamental - if I refresh the page when viewing these stories, these errors just disappear? So I'm not sure they're completely reliable.
> 
> I have added an a11y test for the ic-horizontal-scroll which should capture any issues, but I'm not getting any accessibility failures from it - or from anything else (neither the top navigation, page header or tabs).
> 
> I have also checked with a screen reader and when I focus on the tabs / nav items, the correct information is still being read out (same as on the develop branch - so a screen reader user can navigate through the items without knowing that the scroll buttons exist or having to deal with them which I believe is what we want).
> 
> I have checked the accessibility trees in the DevTools and these are all as expected.

I think the only thing i would query is if we need to add <ul> around the slot in page header to match other comps (eg top navigation). i.e. to fix errors about the child items having the role "list-item"

ad9242:
found a little bit of a strange issue with the tabs on page header: if you start full screen & then make smaller, the horizontal scroll appears ok:
image

but if I then make the window full size again, it doesn't hide the scroll buttons:
image

packages/web-components/src/components/ic-navigation-item/ic-navigation-item.css
```diff
@@ -25,16 +25,17 @@ ic-typography {
 
 :host(.navigation-item) .link,
 :host(.navigation-item) ::slotted(a) {
-  height: 100%;
+  height: calc(100% - 4px);
```
ad9242, line 28: is this (and padding below) ok being px or should it be rem?

ad9242 (approved):

GCHQ-Developer-741 (approved):

DBD324 (approved):
Looks great. As far as I can tell it operates as per the intended design.

Issue #456, opened 2023-03-06
[ic-select (searchable)]: Unable to set default value


jd239:
## Summary of the bug
When setting a default value with searchable select, the input does not display the value. The consumer is currently passing through the value once the data has been loaded which triggers the `icChange` event. This ticket to allow a value to be set by default and for it to appear as the label (if value is matched within the options list) or display the custom value. 

## 🪜 How to reproduce
Tell us the steps to reproduce the problem:
1. Add 'Cap' to searchable select value

## 🧐 Expected behaviour
Expect to see the label of the value or value itself if does not match any items in options list on first load.

## 📸 Screenshots or code
N/A

## 🖥 Desktop
 - OS: Windows
 - Browser: Chrome
 - Version: 110

## 📱 Mobile
 - Device: [e.g. iPhone13]
 - Browser: [e.g. chrome, safari]
 - Version: [e.g. 22]

## Additional info
A method is required to pass the value on load to `this.searchableSelectInputValue`.

PR #469, opened 2023-03-09
Bugfix/456 ic select searchable default value

jd239:
## Summary of the changes
- Default value appears for IcSelect searchable
- When initial value is set and there are no results, the menu does not display
- Check for searchResultsEl before applying innerText
- Set value and option value to lowercase so they match when typing value. This was not working when matching all uppercase value with lowercase.

## Related issue
#456 

## Checklist


GCHQ-Developer-465 (approved):

packages/web-components/src/components/ic-select/ic-select.spec.tsx
```diff
@@ -1199,4 +1200,38 @@ describe("ic-select", () => {
...
+    page.root.options = [];
+    await page.waitForChanges();
+
+    expect(page.rootInstance.searchableSelectInputValue).toBe("Test value 01");
+
+    const input = page.root.shadowRoot.querySelector("input");
+    expect(input.value).toBe("Test value 01");
+  });
+
+  it("should set the default value of searcahable as option label if matching label/value exists", async () => {
```
ad9242, line 1219:
```suggestion
  it("should set the default value of searchable as option label if matching label/value exists", async () => {
```

packages/web-components/src/components/ic-select/ic-select.e2e.ts
```diff
@@ -1471,6 +1481,15 @@ describe("ic-select", () => {
     expect(value).toBe(undefined);
   });
 
+  it("shouldnt display no results found on initial load and setting default value on disable filter", async () => {
```
ad9242, line 1484:
```suggestion
  it("shouldn't display no results found on initial load if setting default value and disable filter", async () => {
```

ad9242 (commented):
a couple of little suggestions

packages/web-components/src/components/ic-select/ic-select.e2e.ts
GCHQ-Developer-465: Perhaps it should be 'should not' for consistency as most e2e tests are using 'should not'.

GCHQ-Developer-465 (changes requested):

packages/react/src/stories/ic-select.stories.mdx
```diff
@@ -489,6 +507,32 @@ export const searchableOptionsWithRecommended = [
   </Story>
 </Canvas>
 
+export const SearchableDefaultValue = () => {
```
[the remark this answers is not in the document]
jd239: This is capitalized because it returns a React component.

ad9242:
if using on a form with a reset button, then the searchable variant doesn't reset to the original value, only to null. Should it reset to the original value? Although I guess you would not necessarily have the original options passed in in order to set the selected item

ad9242 (dismissed):

ad9242 (approved):

GCHQ-Developer-465 (approved):

Issue #391, opened 2023-02-17
[all]: Review `z-index` on all components. 


jd239:
## What is the purpose of the work?
The purpose of this ticket is to review all components `z-index` property and make sure the layering of components is correct. So, for example, a button does not appear above a side navigation once expanded.

This ticket should fix the issues in the following tickets:
- https://github.com/mi6/ic-ui-kit/issues/375
- https://github.com/mi6/ic-design-system/issues/168 (Contains screenshot)

## Why do we need it?
There are instances where text within the content area bleed through page header. In general, the components should be styled with a hierarchy in mind so they do not bleed through each other.

Inspiration from MUI can be used in this instance: https://mui.com/material-ui/customization/z-index/. However, a page within the guidance site is not required.

jd239:
The guidance update for this ticket: https://github.com/mi6/ic-design-system/issues/313

ad9242:
proposing to put this on hold until Dialogs and Popover components are implemented, so we can test when other comps are placed on dialogs for example. Eg would a popup menu on a dialog be visible

ad9242:
potentially tooltip should have highest z-index of all to resolve issues where they disappear behind menus

PR #471, opened 2023-03-13
fix(web-components): updates z-index values for components

ad9242:
## Summary of the changes
Adds a base z-index value that is used to calculate z-index of all elements. Values can also be overridden at the component level

## Related issue
#391 

## Checklist


jd239 (approved):
All good. Can we add an example of how we would update the the base and a component specific in the [guidance ticket](https://github.com/mi6/ic-design-system/issues/313)? This is so whoever picks it up doesn't need to test an the implementation beforehand

MI6-451 (approved):

Issue #268, opened 2023-01-10
[ic-select/ic-search-bar] Add loading state to dropdown components: Select, multi-select, search, searchable select.


GCHQ-Developer-530:
## Summary
Consider default state are.
- If using a prop, consider what happens if the developer forgets to use it. Probs don't use prop.  

Consider how animations may mask the delay times
if using animations, consider what happens when user sets prefs to turn off animations. (already have animation dropdown)
Ensure that loading state is displayed for long enough if displayed at all. 

### Automatic loading state (may decide not to include this functionality)
**Given** that a select component exists, 
**When** the select component is clicked,
**Then** the empty droplist starts to animate into view over a duration of 300ms,
**And** the droplist is initially sized to house only the loading state. 

**Given** that a select component exists and has been clicked, 
**When** the droplist animation starts, 
**And** If the list options are loaded before the animation finishes, 
**Then** the droplist grows to display its full height with the options loaded. 

**Given** that a select component exists and has been clicked, 
**When** the droplist animation starts, 
**And** If the list options are not loaded before the animation finishes, 
**Then** the droplist displays the loading state,
**And** the loading state is displayed for a minimum duration of 1000ms. 

**Given** that a select component exists and has been clicked, 
**And** the loading state is displayed, 
**When** the list options are loaded,
**Then** the droplist animates to display its full height with the options loaded. 

**Given** that a select component exists and has been clicked, 
**And** the loading state is displayed, 
**When** the 'timeout' prop is set to true, 
**Then** the loading error state should be displayed. 

**Given** that a select component drop list is showing the loading error state
**When** the 'retry' button is pressed
**Then** the droplist should return to loading state whilst the next load takes place. 

### Search input loading
**Given** that a search component exists, 
**When** a user types a search term into the field, 
**Then** the droplist animates into view over a 300ms duration.

**Given** that a search component exists and a term has been entered,
**When** the droplist animation starts, 
**If** the search results are loaded before the animation finishes, 
**Then** the droplist grows to display its full height with the search results loaded. 

**Given** that a search component exists and a term has been entered,
**When** the droplist animation starts, 
**If** the search results are not loaded before the animation finishes, 
**Then** the droplist displays the loading state,
**And** the loading state is displayed for a minimum duration of 1000ms. 

**Given** that a search component exists and a term has been entered,
**And** the loading state is displayed, 
**When** the search results are loaded,
**Then** the droplist animates to display its full height with the results displayed. 

[...]

PR #517, opened 2023-03-27
268 search select loading states

GCHQ-Developer-741:
## Summary of the changes
Implemented loading states for select and search bar when using external filtering.

## Related issue
#268 

## Checklist


GCHQ-Developer-741:
Since the deployment links didn't get added...

Web Components: https://mi6.github.io/ic-ui-kit/branches/268-search-select-loading-states/web-components/
React: https://mi6.github.io/ic-ui-kit/branches/268-search-select-loading-states/react/

audit-ci.json
```diff
@@ -4,6 +4,7 @@
     "GHSA-w5p7-h5w8-2hfq",
     "GHSA-7p7h-4mm5-852v",
     "GHSA-pfrx-2q88-qq97",
-    "GHSA-9c47-m6qq-7p4h"
+    "GHSA-9c47-m6qq-7p4h",
+    "GHSA-hc6q-2mpp-qw7j"
```
ad9242, line 8: think a rebase will remove need for this

ad9242 (changes requested):

ad9242:
Found a bit of a strange one:

- type a value in so that loading displays (and then loading error)

image

- tab focus to retry button (a bit strange that focus border is chopped off):
image

- now use ArrowUp or ArrowDown: the loading error gets the focussed option styling:
image

jd239:
Can the commits be fixed please. There are updates in different packages but only a single web-component commit.

packages/web-components/src/components/ic-search-bar/ic-search-bar.tsx
```diff
@@ -159,9 +166,18 @@ export class SearchBar {
    */
   @Prop() options?: IcMenuOption[] = [];
 
+  /**
+   * If using external filtering, set a timeout for when loading takes too long.
+   */
+  @Prop() timeout?: number = null;
```
jd239, line 172: Think we need to add `null` as a data type here as it's not just number.

packages/web-components/src/components/ic-search-bar/ic-search-bar.tsx
```diff
@@ -38,13 +38,17 @@ import searchIcon from "../../assets/search-icon.svg";
 
 let inputIds = 0;
 
+const LOADING_LABEL = "Loading...";
+const LOADING_ERROR_LABEL = "Loading Error";
```
jd239, line 42: Should these two be a prop so they can be changed but have the default as they are above.

jd239:
2 issues when using the loading stats

1.
- Type in two letters
- Trigger the loading error
- Tab to retry button and press enter
- Menu disappears
Should that keep the menu open and show loading again?

2.
- Type in two letters
- Trigger the loading error
- Tab to retry button and press space
- Loading error has focus with the action blue so initially couldn't tell that the retry button was still there.

jd239 (changes requested):
Added a few comments. More to do with coding standards rather than implementation. Good work. Looks slick!

packages/web-components/src/components/ic-menu/ic-menu.tsx
```diff
@@ -26,6 +32,7 @@ import {
   styleUrl: "ic-menu.css",
   shadow: true,
 })
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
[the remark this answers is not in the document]
GCHQ-Developer-741: The actual Menu class. It never gets actually referenced so all the components have a warning which isn't really relevant. Thought to add this to get rid of the warning

jd239:
Just had a thought. Is it possible to use the loading state with internal filtering?
- RE: Search bar. When typing the initial search phrase, the request is made to the back end to retrieve all the options (loading state appears) and then the internal filtering is used to filter the options
- RE: Select/Searchable. User clicks on the input/select to open the menu and then makes a request (loading state appears). Searchable will then use the internal filter.

MI6-451 (approved):

jd239:
> Just had a thought. Is it possible to use the loading state with internal filtering?
> 
> * RE: Search bar. When typing the initial search phrase, the request is made to the back end to retrieve all the options (loading state appears) and then the internal filtering is used to filter the options
> * RE: Select/Searchable. User clicks on the input/select to open the menu and then makes a request (loading state appears). Searchable will then use the internal filter.

Has this been implemented as well? If so, is there a story to test this?

ad9242:
Space is working now - thanks! Just a couple of things spotted:

1) when using searchable select eg searchable default story (https://mi6.github.io/ic-ui-kit/branches/268-search-select-loading-states/web-components/?path=/story/web-components-select--searchable-default), the select loses the focus border as you type eg
- click in field: focus is correct with drop down displayed:
image


- now type "ca" - it loses focus border (but displays correct filtered options)
image

this is the behaviour on develop branch (different options though)
image

2) the "x" clear button disappears when in the loading error state:
image

ad9242:
sorry, a couple more issues:

1) type "es" & get the loading error\retry:
image
- now clear the clear selection "x"
- type in the field & you no longer get any suggestions

2) type into the field so that loading error\retry appears
- use keyboard to select "retry"
- options are updated, but the input has lost focus so you cannot select anything with arrow keys

ad9242 (approved):

MI6-451 (approved):

ad9242 (approved):

MI6-451 (approved):

jd239 (approved):

Issue #522, opened 2023-03-28
Searchable select 'min char' prop is unwanted functionality


fbd399:
## Summary of the bug
'Min char' prop allows users to set the filtering to happen after a minimum amount of characters are inputted. Unlike search, select should always show the options available in the dropdown list, as the searchable prop is simply added functionality to filter the dropdown list results.

## 🪜 How to reproduce
1. Set 'min char' prop to 2
2. Focus on searchable select
3. Notice how list remains closed
4. Type in 1 character
5. Notice how no results are still not shown

## 🧐 Expected behaviour
As soon as searchable select is in focus, the dropdown list should be displayed.

fbd399:
Since the results exist already, the ‘min char prop’ is blocking access to the list rather than filter the list. Ideally, we shouldn't allow min char on searchable select, so even though it's a breaking change it essentially is fixing an important bug. 

For V3 we’ll want to remove the ‘min char’ prop to remove this functionality entirely.
However, because it’s a breaking change, between V2 and V3 we may want to set the dropdown list to open immediately, with the filtering being the only prop affected by the ‘min char’ prop.

In practice, the user would focus on the select, and the dropdown list would appear. If min char is set to 2, if they type in ‘f’ nothing happens, but they still have access to the dropdown. As soon as they type in a second character it would filter the results.

PR #536, opened 2023-03-31
fix(web-components): deprecate charactersUntilSuggestions prop for select and remove stories/tests

GCHQ-Developer-530:
## Summary of the changes
Deprecate charactersUntilSuggestions prop for select and remove stories/tests with references to charactersUntilSuggestions

## Related issue
#522

## Checklist


ad9242:
Looks good! Just one question: the searchable external filtering stories (web comps & react) have a check that only fetches the list after the second character is typed. Should we leave this in there, or remove it now?

GCHQ-Developer-530:
> Looks good! Just one question: the searchable external filtering stories (web comps & react) have a check that only fetches the list after the second character is typed. Should we leave this in there, or remove it now?

I'll remove it as otherwise it looks like there are no results and that could be misleading

GCHQ-Developer-847 (approved):

Issue #484, opened 2023-03-16
[ic-select]: a11y violation regarding aria-owns and aria-controls


jd239:
## Summary of the bug
While using the IBM Accessibility tool on ic-select, accessibility rule 4.1.2 (Name, Role, Value) has failed.

## 🪜 How to reproduce
Tell us the steps to reproduce the problem:
1. Open up a page with ic-select
2. Download IBM Equal Access Accessibility tool
3. Run tool against ic-select

## 🧐 Expected behaviour
No violations appearing

## 📸 Screenshots or code
![Capture](https://user-images.githubusercontent.com/116016939/225564070-794cbc9e-77ec-4e54-adb0-cc8f64eb0adb.PNG)

## 🖥 Desktop
 - OS: Windows
 - Browser: Chrome
 - Version: 110

## Additional info
This violation did not appear in the Storybook accessibility checker or the WAVE Evaluation tool in Chrome. Further investigation is required on why the IBM tool is displaying this error but other tools are not.

jd239:
Initial findings so far:
- This fix involves making structural updates on `<ic-menu>`.
- Moving the `id` from the `ul` which is in the shadow DOM onto the `<Host>`. 
- Removing the `<ul>` completely
- Changing the `li` into `div` and applying the `role="option"` to them

This may cause issues with the focus state as it was previously on the `ul`.

PR #554, opened 2023-04-05
fix(web-components): removed shadow DOM from ic-menu for a11y fixes

jd239:
## Summary of the changes
aria-owns and aria-controls was unable to link to the UL element within ic-menu due to shadow DOM encapsulation. Instead of restructuring the component, scoped worked.

Checked against accessibility scanner 'Equal Access Accessibility Checker' chrome plugin

## Related issue
#484 

## Checklist


GCHQ-Developer-847 (approved):
I don't think I can test this with a screen reader as I'm not able to use the ones for which aria-owns / aria-controls are supported, but I'm happy to approve as you've checked with that Chrome plugin

PR #556, opened 2023-04-05
fix(web-components): add flex-wrap CSS styling to interaction controls slot in ic-card

GCHQ-Developer-112:
## Summary of the changes
- Add `flex-wrap` CSS styling to interaction controls slot in `ic-card`
- Update package-lock.json in `react` package following a new release of ICDS components

## Related issue
#529

## Checklist


packages/web-components/src/components/ic-card/ic-card.stories.mdx
```diff
@@ -454,10 +456,12 @@ import Button from "../ic-button/readme.md";
           d="M17.65 6.35C16.2 4.9 14.21 4 12 4c-4.42 0-7.99 3.58-7.99 8s3.57 8 7.99 8c3.73 0 6.84-2.55 7.73-6h-2.08c-.82 2.33-3.04 4-5.65 4-3.31 0-6-2.69-6-6s2.69-6 6-6c1.66 0 3.14.69 4.22 1.78L13 11h7V4l-2.35 2.35z"
         />
       </svg>
-      <div slot="interaction-controls" style="display:flex;gap:16px;">
-        <ic-button variant="primary">Learn more</ic-button>
-        <ic-button variant="secondary">Hide</ic-button>
-      </div>
```
jd239, line 460: I would imagine consumers have been using it with a 'interaction-controls' wrapper. How does it look with your CSS changes because when this update is released, it might unexpectedly change the styling on external apps.

jd239 (changes requested):
Only a few comments. 
Is it possible to get add a story with either the viewport small or card with a reduced width so we can get a regression test with the buttons stacked..

MI6-451 (dismissed):

GCHQ-Developer-530 (approved):

MI6-451 (approved):

packages/web-components/src/components/ic-card/ic-card.stories.mdx
jd239: Were you able to test the previous implementation above but with the new styling?

packages/web-components/src/components/ic-card/ic-card.tsx
```diff
@@ -284,6 +286,7 @@ export class Card {
                   ["toggle-button-expanded"]: this.areaExpanded,
                 }}
                 variant="icon"
+                size="large"
```
jd239, line 289: This makes the button 40x40 but designs are 32x32.

GCHQ-Developer-112: I spoke to the design team on this and it was agreed that we use the large variant of the icon button so the height of the buttons (chevron & slotted buttons) are the same. Design plans to update size variants so that the default size will be 40px across all components.

packages/web-components/src/components/ic-card/ic-card.css
```diff
@@ -145,10 +145,18 @@ button {
   justify-content: center;
 }
 
-.interaction-controls {
+.interaction-area {
+  display: flex;
+  gap: var(--ic-space-md);
   margin-top: var(--ic-space-md);
+  align-items: flex-end;
```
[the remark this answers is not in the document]
GCHQ-Developer-112: Thank you 😀

GCHQ-Developer-530 (dismissed):

packages/web-components/src/components/ic-card/ic-card.stories.mdx
GCHQ-Developer-112: If an `interaction-controls` wrapper is used, then the user will mostly see the effects of the new CSS if they have more than one element with the slot attribute: `slot="interaction-controls"`. Below is how it looks if I were to use a wrapper. I believe the only difference a user using an `interaction-controls` wrapper would see is the positioning of the chevron. It will now appear at the bottom right of the card. 

The rest of the new CSS won't be applied inside the wrapper. As there's only one I wouldn't want to add styling a slotted div (`::slotted(div)`) as that could cause unintended side effects. 
Screenshot 2023-04-18 at 14 20 54

Issue #568, opened 2023-04-12
IcTabContext cannot detect new Tabs being added after inital render


GCHQ-Developer-741:
## Summary of the bug
Tabs that are added programatically after initial render do not have tabId's applied and therefore do not work.

## 🪜 How to reproduce
- Copy the IcTabs example
- Change the Tabs to be mapped over with a variable
- Add to that variable after a time delay
- The new tab will not work

## 🧐 Expected behaviour
When a new tab is added, the corrosponding ids should be added behind the scenes

## 📸 Screenshots or code

  ```
const [x, setX] = useState(['One', 'Two', 'Three'])
  setTimeout(() => setX([...x, 'Four']), 3000)
  return (
    <IcTabContext>
      <IcTabGroup label="Example tab group">
        {x.map((val, index) => (
          <IcTab key={index}>{val}</IcTab>
        ))}
      </IcTabGroup>
      {x.map((val, index) => (
        <IcTabPanel key={index}>{val}</IcTabPanel>
      ))}
    </IcTabContext>
  )
```

## Additional info
tabIds are currently only set during componentDidLoad (when IcTabContext first enters the DOM). Could maybe add a function to update new tabs on a rerender?

PR #574, opened 2023-04-13
fix(web-components): added functionality to link tabs when added/removed prog…

GCHQ-Developer-741:
…ramatically

tabs will now work even when updated programatically and new ones are added to the context after initial render

Edit: PR has been updated to include tabs which have been removed dynamically. 

## Related issue
#568

### DIFF
--- a/packages/web-components/src/components/ic-tab/ic-tab.tsx
+++ b/packages/web-components/src/components/ic-tab/ic-tab.tsx
@@ -56,6 +56,8 @@ export class Tab {
    */
   @Event() tabFocus: EventEmitter<IcTabClickEventDetail>;
 
+  @Event() tabCreated: EventEmitter<HTMLIcTabElement>;
+
   private isInitialRender: boolean = true;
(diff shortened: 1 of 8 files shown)


packages/web-components/src/components/ic-tab/ic-tab.tsx
```diff
@@ -56,6 +56,8 @@ export class Tab {
    */
   @Event() tabFocus: EventEmitter<IcTabClickEventDetail>;
 
+  @Event() tabCreated: EventEmitter<HTMLIcTabElement>;
```
jd239, line 59: Internal?

jd239 (changes requested):
Should we do something if a tab gets removed as well?

jd239 (commented):
I made updates to the ticket so don't think I'm best to review after doing the first round of reviews.

GCHQ-Developer-530 (approved):

GCHQ-Developer-847 (approved):

Issue #503, opened 2023-03-22
IcTextField should pass all other props available to input components and check for other components


MI6-651:
## Summary
IcTextField only passes down some properties, it should pass down all other properties as well into the input component.

Specifically this ticket will add any properties to TextField not currently available, but will not retrospectively add the hidden aria props (that are available, just hidden). 

## 💰 Use value
This is useful as it provides more options for developers, as it adds functionality e.g. validation on browser date pickers with an ICDS text field.

ad9242:
Rather than exposing as props, could potentially be implemented the way that aria attributes are inherited currently, i.e. have a list of the supported attributes. 
Technically would not be a breaking change if we wanted to change the existing props to this method, as they could still be set on the component, just not documented as props

ad9242:
Following on from previous comment - the decision is to use props

MI6-255:
https://github.com/mi6/ic-ui-kit/issues/515

PR #579, opened 2023-04-14
503 ictextfield should pass all other props available to input components and check for other components

MI6-255:
## Summary of the changes
Added min/max values to TextField, alongside checking if any other attributes were relevant to pass down.

## Additional information
There will be a guidance ticket that suggests the developer specifies the min/max in the `label` or `helperText`. For that reason, I decided not to add any extra information like maximum characters does where it adds `Field can contain a maximum of {maxNumChars} characters.` to an `aria-describedby`. I'm certain if it's in the `label`, that would be fine, but less certain about if that information was conveyed in `helperText`. Let me know if anyone disagrees with this approach, or thinks it could be an issue. 

## Related issue
#503 

## Checklist


MI6-255:
This is not working as intended since being pushed up, so converted to a draft.

GCHQ-Developer-530 (approved):

MI6-451 (approved):

PR #626, opened 2023-04-27
fix(web-components): ic-select-autofill-fix

GCHQ-Developer-575:
Stop autofill from browser in ic-select component

## Summary of the changes
Added search type to input tag. Removed default "x" button. Turned autocomplete to "off" which stops chrome adding its own dropdown.

## Related issue
677


packages/web-components/src/components/ic-select/ic-select.tsx
```diff
@@ -850,6 +850,8 @@ export class Select {
                 <input
                   class="select-input"
                   role="combobox"
+                  type="search"
```
jd239, line 853: Is this required when we autocomplete is set to off below?

GCHQ-Developer-575: It unusually is, weird caveat of the input element.
I considered other solutions such as setting autoComplete="new-password" etc. But this made the most sense in the context.
I believe it also adds to the semantics since it is a search box (although annoyingly it does need that CSS to remove the default "x")

GCHQ-Developer-575: Agreed, tested again and autocomplete="off" worked by itself.

GCHQ-Developer-530 (approved):

jd239 (approved):

Issue #464, opened 2023-03-07
[ic-checkbox]: Update event.detail from `icChange` and `icCheck`


jd239:
## Summary
`icChange` which is emitted from `ic-checkbox-group` returns an array of `ic-checkbox` values. When checking and unchecking the checkboxes within the group, the same array is returned. Further work is required by the consumer to get the checked state of the checkboxes which, may not be necessary if more detail is returned from the event. This ticket is to add a checked object to the `icChange` event so that `ic-checkbox-group` returns an object which can be used directly. 

## 💬 Description
When logging the value detail from `icChange` on `IcCheckboxGroup`, an array of the enabled checkbox values are returned `['foo','bar']`. When an checkbox is unchecked, the `icChange` still emitts `['foo', 'bar']`. This information is emitted from the value `event.detail.value`. No detail about what has been unchecked/checked is provided by the event.

It's worthwhile returning an object which provides the value, label, checked, other (if checkbox contains `ic-textfield` and disabled state).

Something like (but not exactly):
```jsx
[
    { label: 'item-1', value: 'item-1', disabled: true, other: null, checked: false },
    { label: 'item-2', value: 'item-2', disabled: false, other: null, checked: true },
    { label: 'item-3', value: 'item-3', disabled: false, other: null, checked: true },
]
```

This should be possible by getting the `ic-checkbox-group` to querySelectorAll of the `ic-checkbox` and when an `icCheck` is triggered from one of the checkboxes, `IcCheckboxGroup` is able to retrieve the status of the updated checkbox and update the `icChange` object. 

## 💰 Use value
Providing this object will make it easier for consumers to filter out the checked items. 

## Additional info
N/A

GCHQ-Developer-112:
#581 duplicates this ticket. This is the body of the #581 ticket.

## Summary of the bug
When you have multiple `ic-checkbox` components in `ic-checkbox group` and select a checkbox, the value emitted by the `onIcChange` event updates with the value of the selected checkbox. However, if you unselect the same checkbox, the value emitted by the `onIcChange` event. This is particularly an issue if you try to filter using checkboxes. As the filter works based on the value (event.detail.value) from `onIcChange`.

## 🪜 How to reproduce
For Core Team: see the ICDS test app, run the code and view the following path: "/drinks/about-you"

## 🧐 Expected behaviour
Expect the value in onIcChange when a checkbox is checked AND unchecked i.e. the value should always update and reflect the checkboxes that are currently checked

[...]

PR #628, opened 2023-04-27
feat(web-components): changed event.detail for ic-checkbox icChange

GCHQ-Developer-741:
## Summary of the changes
Changed icChange in `ic-checkbox-group` to return an array of checked `ic-checkbox`'s, so that users can easily retrieve values from the components when they are updated

## Related issue
#464 

## Checklist

### DIFF
--- a/packages/react/src/stories/ic-checkbox.stories.mdx
+++ b/packages/react/src/stories/ic-checkbox.stories.mdx
@@ -113,7 +113,7 @@ import { useForm } from 'react-hook-form';
 
 <Canvas>
   <Story name="Conditional dynamic">
-        <IcCheckboxGroup label="Conditional dynamic" name="1">
+        <IcCheckboxGroup label="Conditional dynamic" name="1" onIcChange={(ev) => console.log(ev.detail.selectedOption.label, ev.detail.selectedOption.checked)}>
           <IcCheckbox
             additionalFieldDisplay="dynamic"
             value="valueName1"
--- a/packages/web-components/src/components/ic-checkbox-group/ic-checkbox-group.e2e.ts
+++ b/packages/web-components/src/components/ic-checkbox-group/ic-checkbox-group.e2e.ts
@@ -127,8 +127,7 @@ describe("ic-checkbox-group component", () => {
     );
     expect(icChange).toHaveReceivedEventDetail({
       value: ["test2"],
-      checkedOptions: [checkboxElement],
-      textFieldValues: [null],
+      checkedOptions: [{ checkbox: checkboxElement }],
       selectedOption: checkboxElement,
     });
 
@@ -138,7 +137,6 @@ describe("ic-checkbox-group component", () => {
     expect(icChange).toHaveReceivedEventDetail({
       value: [],
       checkedOptions: [],
-      textFieldValues: [],
       selectedOption: checkboxElement,
     });
   });
--- a/packages/web-components/src/components/ic-checkbox-group/ic-checkbox-group.e2e.ts
+++ b/packages/web-components/src/components/ic-checkbox-group/ic-checkbox-group.e2e.ts
@@ -119,8 +119,27 @@ describe("ic-checkbox-group component", () => {
     await checkbox.click();
     await page.waitForChanges();
 
+    const checkboxElement = await page.evaluate(
+      () =>
+        document.querySelector(
+          "ic-checkbox[value='test2']"
+        ) as HTMLIcCheckboxElement
+    );
     expect(icChange).toHaveReceivedEventDetail({
       value: ["test2"],
+      checkedOptions: [checkboxElement],
+      textFieldValues: [null],
+      selectedOption: checkboxElement,
+    });
+
+    await checkbox.click();
+    await page.waitForChanges();
+
+    expect(icChange).toHaveReceivedEventDetail({
+      value: [],
+      checkedOptions: [],
+      textFieldValues: [],
+      selectedOption: checkboxElement,
     });
   });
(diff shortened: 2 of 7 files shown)


GCHQ-Developer-112 (dismissed):

packages/web-components/src/components/ic-checkbox-group/ic-checkbox-group.e2e.ts
jd239, line 1: Is there a test with 2/3 checked boxes, removing 1/2 and then inspecting the checkboxGroup value?

packages/web-components/src/components/ic-checkbox-group/ic-checkbox-group.types.ts
```diff
@@ -1,3 +1,3 @@
 export interface IcChangeEventDetail {
-  value: string[];
+  checkedOptions: HTMLIcCheckboxElement[];
```
jd239, line 2: This is a breaking change as someone _might_ be using `value`. Appreciate it needs updating so the unchecked value is removed from the array.
Could we not add checkedOptions as a new object emitted from the event?

packages/web-components/src/components/ic-checkbox/ic-checkbox.tsx
```diff
@@ -127,6 +128,9 @@ export class Checkbox {
 
   componentWillLoad(): void {
     addFormResetListener(this.host, this.handleFormReset);
+    this.host
+      .querySelector("ic-text-field")
+      ?.addEventListener("icChange", (e) => e.stopImmediatePropagation());
```
jd239, line 133: How does a consumer grab the value from the text-field and know it's related to a particular checkbox?

jd239 (changes requested):
Just a couple comments.

packages/web-components/src/components/ic-checkbox/ic-checkbox.tsx
GCHQ-Developer-741: I wasn't aware that was previously a feature. This code only prevents `ic-text-field`'s onChange from bubbling up to where it was being caught by the onIcChange of `ic-checkbox-group`. They could still put an onIcChange on the text-field itself and then check if the element's parent is a checkbox or not.

jd239: I guess that is where the idea for the object below comes into play....
```
[
    { label: 'item-1', value: 'item-1', disabled: true, other: null, checked: false },
    { label: 'item-2', value: 'item-2', disabled: false, other: null, checked: true },
    { label: 'item-3', value: 'item-3', disabled: false, other: null, checked: true },
]
```

Is it possible to add something to the checkedOptions array to include the textfield value?

GCHQ-Developer-112: Would be good for the icCheck event to emit the checked value - struggling to get the checked value of a checkbox through the event

GCHQ-Developer-741: Have updated the stories with an example. Also added another attribute to show which checkbox was pressed

packages/react/src/stories/ic-checkbox.stories.mdx
```diff
@@ -113,7 +113,7 @@ import { useForm } from 'react-hook-form';
 
 <Canvas>
   <Story name="Conditional dynamic">
-        <IcCheckboxGroup label="Conditional dynamic" name="1">
+        <IcCheckboxGroup label="Conditional dynamic" name="1" onIcChange={(ev) => console.log(ev.detail.selectedOption.checked)}>
```
ad9242, line 116:
```suggestion
        <IcCheckboxGroup label="Conditional dynamic" name="1" onIcChange={(ev) => console.log(ev.detail.selectedOption.label, ev.detail.selectedOption.checked)}>
```
Just to show a bit more info on what changed

packages/web-components/src/components/ic-checkbox/ic-checkbox.tsx
ad9242, line 133: we should also remove the event listener in the disconnected callback. see similar code in navigation group

ad9242 (changes requested):
just a note re tidying up event listener

packages/web-components/src/components/ic-checkbox/ic-checkbox.tsx
```diff
@@ -141,6 +145,9 @@ export class Checkbox {
 
   disconnectedCallback(): void {
     removeFormResetListener(this.host, this.handleFormReset);
+    this.host
+      .querySelector("ic-text-field")
+      ?.removeEventListener("icChange", (e) => e.stopImmediatePropagation());
```
ad9242, line 150: I don't think this will tidyup correctly as it is not removing the exact function that is bound in the addEventlistener, it is being passed a different function instance. 
Think it needs to be more like the example here: https://medium.com/@bigcatplichta/javascript-use-bind-to-dynamically-add-and-remove-event-listeners-d6b443877a73

ad9242 (approved):

GCHQ-Developer-112 (dismissed):

packages/web-components/src/components/ic-checkbox/ic-checkbox.tsx
jd239: Can see that a `textfieldValues` nd `checkedOptions` array has been added to the `icChange` event and the order of textfield values match. 
Could we merge the two together as they should line up based on the index?

jd239 (approved):

ad9242 (approved):

PR #644, opened 2023-04-29
fix(web-components): ic-card interaction section set to bottom

### LINKED ISSUE #528: [ic-card]: Interaction section is not positioned to the bottom if a height or a lot of content has been set.
## Summary of the bug
When using `ic-card` in a row, there will be instances where they have an equal height but not all cards will have content to push the interaction controls to the bottom. This behaviour does not match the anatomy of the card specified in the guidance. 

## 🪜 How to reproduce
1. Add an `ic-card` which includes heading, message and interaction controls.
2. Set a height on the card.
3. View the interaction controls underneath the message and not at the bottom of the card.

## 🧐 Expected behaviour
Interactions controls (and expandable) should appear at the bottom of the card.

## 📸 Screenshots or code
Capture

## 🖥 Desktop
 - OS: Windows
 - Browser: Chrome
 - Version: 110

## 📱 Mobile
 - Device: N/A
 - OS: N/A
 - Browser: N/A
 - Version: N/A

## Additional info
Since the content area is placed above the interaction controls/expand button, adding `flex grow` should allow the section to take up the remaining space and push the interaction controls to the bottom.

joshmu:
## Summary of the changes
**ic-card** styles updated to guarantee _interaction section_ is always located at the bottom of the card, regardless if there is additional height defined.

Additional storybook variant added for **ic-card** called _Additional height_ to show this scenario.

Solution was to apply flex grow to the interactive section itself to preserve the flat DOM structure of the possible card content variations since there is no definition of a card-body wrapper div.

## Related issue
Fix #528 

## Checklist


joshmu:
Just a heads up I believe visual regression tests are failing since your repo loki config will point to the default chromium instance on the operating system.

I would assume using docker chromium would guarantee a match between the developer machine and the pipeline.

If I'm wrong I'd love to know why as the pipeline is failing on a snapshot I have created in this PR.
I have passing tests if I run the ci test commands locally. 😅

ad9242:
> Just a heads up I believe visual regression tests are failing since your repo loki config will point to the default chromium instance on the operating system.
> 
> I would assume using docker chromium would guarantee a match between the developer machine and the pipeline.
> 
> If I'm wrong I'd love to know why as the pipeline is failing on a snapshot I have created in this PR. I have passing tests if I run the ci test commands locally. 😅

Hi @joshmu. Yes you are correct that it is because the default os chromium instance is used and there can be subtle rendering differences. We have investigated use of docker but hit some other issues. The current solution is as follows:

If the tests are passing for you locally, you can run the `Update visual regression test images` action on your branch. This will run the tests & update the reference images, so that the tests should pass in the pipeline. Unfortunately, if you then pull those reference images down, the tests will start failing locally though.

If you are not happy to run that action locally though, we can quickly resolve it once merged

joshmu:
Hey @ad9242 
I'm happy to manually trigger the action, unfortunately I do not have access to trigger the update workflow on this repo, and on my forked repo none of the ic-ui-kit actions are visible.
I have pulled latest develop & main and rebased my own branch with the latest.
Let me know if I'm missing something as it would be good to contribute and get all green

ad9242:
That's strange. I can see on your forked repo that there is no "update visual regression test images" workflow. In fact quite a few are missing. I tried creating a fork & they all appeared for me (once I enabled workflows).
Maybe we do look at getting it resolved once merged

ad9242 (approved):

joshmu:
Yep sounds good, once this goes through I'll try fork the repo again and see if the actions then become available.

GCHQ-Developer-112 (approved):

ad9242:
visual tests will be fixed by https://github.com/mi6/ic-ui-kit/pull/652

Issue #612, opened 2023-04-25
Text field error validation without icon or text


MI6-451:
## Summary
The text field should allow for 'inline' error validation.

## 💬 Description
The text field currently allows for inline validation but only for 'success' validation. There is a requirement to allow similar functionality for 'error' validation but instead of having the 'error' icon appear in the text field input, as is the case with the 'success' validation, there should be no icon or validation text. Instead, the validation would only be displayed through the red outline of the text field. Guidance should be suggest that where this is used additional means are used to convey the error so it does not solely rely on different colours, e.g. a tooltip displaying the error.

## 💰 Use value
This would allow for a greater range of 'error' validation and is necessary for a pattern on the Pagination bar.

## Additional info
This is linked to/blocks #460

MI6-255:
We discussed in this ticket:
- An event to expose the validation message that pagination can use
- Removing logic from pagination to check min/max values and use min/max from text field 
- Internal prop for turning off the validation text for own use only)

PR #669, opened 2023-05-10
feat(web-components): add internal prop to allow for validation without text or icon

GCHQ-Developer-112:
## Summary of the changes
To allow the use of text field in pagination bar, as it was designed, an internal prop has been introduced. This prop allows for inline validation on all validation statuses.

## Related issue
#612 

## Checklist


packages/web-components/src/components/ic-text-field/ic-text-field.tsx
```diff
@@ -338,6 +343,15 @@ export class TextField {
     return this.rows > 1;
   };
 
+  /**
+   * @internal Emitted when the validationInlineInternal is `true`
+   */
+  @Event() getValidationText: EventEmitter<IcValueEventDetail>;
```
GCHQ-Developer-741, line 349: How is this intended to be used? Is it for announcing the validation text to screen readers? I can't see it referenced anywhere else.

GCHQ-Developer-112: The intention is that this event will be used in the pagination bar so that the validation text can be used in the tooltip above the go to page part of the pagination bar (see screenshot 1).

Screenshot 2 shows an example of a listen event that can be placed in pagination bar. I think this approach will be particularly useful as I had issues trying to get the validation text using querySelector when I was initially testing the pagination bar.

**Screenshot 1:**
Screenshot 2023-05-12 at 12 17 36 

**Screenshot 2:**
Screenshot 2023-05-12 at 09 51 56

GCHQ-Developer-112:
> I think in the ticket it mentions adding an event which would expose the validation message that would've shown up so that pagination can access it. Please could you add this in?

This has now been added

GCHQ-Developer-530 (approved):

MI6-451 (approved):

Issue #659, opened 2023-05-10
[ic-checkbox] Checkbox label wrap causes checkbox to shrink


GCHQ-Developer-530:
## Summary of the bug
When a checkbox label is long enough that it wraps onto a new line, the checkbox width decreases.

## 🧐 Expected behaviour
The checkbox width should always stay the same.

## 📸 Screenshots or code
<img width="273" alt="Screenshot 2023-05-10 at 07 41 08" src="https://github.com/mi6/ic-ui-kit/assets/111882034/925d6d62-4e39-426f-8d6d-b9cad49f6669">

ad9242:
think the fix is to set flex-shrink: 0 on both the input & the svg elements

PR #678, opened 2023-05-12
fix(web-components): add flex-shrink 0 to checkbox and svgs

GCHQ-Developer-530:
## Summary of the changes
Add flex-shrink 0 to checkbox so it's width doesn't change when the label wraps

## Related issue
#659

## Checklist


GCHQ-Developer-741:
Is probably a question for the designers, but is it correct for the text to not line up like in this screenshot? I would expect it to be aligned on the left.
Screenshot from 2023-05-12 14-43-40

GCHQ-Developer-530:
> Is probably a question for the designers, but is it correct for the text to not line up like in this screenshot? I would expect it to be aligned on the left. Screenshot from 2023-05-12 14-43-40

I've updated it so that the text lines up and aligns centre so that it's consistent with radio 😊

GCHQ-Developer-741 (approved):

ad9242 (approved):

PR #686, opened 2023-05-16
fix(web-components): fixed failing test

GCHQ-Developer-741:
hotfix to prevent a value being updated in search-bar when it has no reason to be


MI6-451 (approved):

GCHQ-Developer-112:
Please could you write a unit test to cover the new if statement. The changes seems to brought the test coverage for `ic-menu` below 80%

Issue #455, opened 2023-03-03
[ic-select] Focus is not correctly applied on keyboard interaction


MI6-451:
## Summary of the bug
When the select is interacted with via the keyboard to select an option (via enter or space) focus should move back to the select itself when the menu has closed but this does not happen. Similarly when an option in the select menu is highlighted/navigated to and the escape key is pressed, the menu closes but focus is not returned to the select component as expected.

## 🪜 How to reproduce
Tell us the steps to reproduce the problem:
1. Focus a select component
2. Navigate to a select option via the up or down arrow key.
3. Select that option with the enter key or space key (or close the menu with the escape key)

## 🧐 Expected behaviour
When interacting with a select menu option via the keyboard (enter or space) or when closing the menu with an option selected the menu should close and focus should return to the select (with the selected option as the new value in the enter or space key case)

## 🖥 Desktop
 - OS: macOS
 - Browser: Chrome

PR #695, opened 2023-05-17
fix(web-components): fixed focus behaviour on blurring of menu for ic…

GCHQ-Developer-741:
…-select

changed focus behaviour to focus ic-select onKeyPress when the ic-menu is focused, as well as writing a test for code coverage

## Related issue
#455

### DIFF
--- a/packages/web-components/src/components/ic-select/ic-select.spec.tsx
+++ b/packages/web-components/src/components/ic-select/ic-select.spec.tsx
@@ -1420,4 +1420,52 @@ describe("ic-select", () => {
       "No results found"
     );
   });
+
+  it("should focus the input when escape is pressed whilst the menu is focused", async () => {
+    const spy = jest.spyOn(HTMLButtonElement.prototype, "focus");
+    const page = await newSpecPage({
+      components: [Select, Menu, InputComponentContainer],
+      html: `<ic-select label="IC Select Test"></ic-select>`,
+    });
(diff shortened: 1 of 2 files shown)


GCHQ-Developer-530 (dismissed):

packages/web-components/src/components/ic-select/ic-select.spec.tsx
```diff
@@ -1420,4 +1420,28 @@ describe("ic-select", () => {
       "No results found"
     );
   });
+
+  it("should focus the input when escape is pressed whilst the menu is focused", async () => {
```
jd239, line 1424: Is there a test for searchable as well?

jd239 (changes requested):
Looks good. I wonder if this behaviour should be added to search bar as well. I appreciate this may open a can of worms though.

GCHQ-Developer-741:
> Looks good. I wonder if this behaviour should be added to search bar as well. I appreciate this may open a can of worms though.

It certainly does. For searchbar this behaviour already works when Enter is pressed as this selects an option which focuses back to the parentEl. However, a similar approach for Escape causes the menu to open again before closing (flickering) which causes an e2e test to fail. I believe its because ic-search-bar opens the menu when it is focused. If its deemed to be necessary then I can investigate further otherwise could just raise a low priority ticket.

GCHQ-Developer-530 (approved):

Issue #664, opened 2023-05-10
Documentation needs improving for events on IcSelect


MI6-255:
Ticket opened on behalf of: @MI6-651
## Summary
 <br /> 
Unclear when certain events are fired: e.g. does icOptionSelect fire icChange? (to be reviewed in all components)
What ‘value’ means in the icSelect is ambiguous, and makes the component difficult to control
More documentation would be good to describe how the icSelect should be used, especially where results are being fetched externally  <br /> 
 <br />

MI6-255:
Make it clear which is for searchable select and which is for the standard select

PR #701, opened 2023-05-19
docs(web-components): update prop and event descriptions for select to be less ambiguous

GCHQ-Developer-530:
## Summary of the changes
Update prop and event descriptions for select to be less ambiguous. Updated boolean props to follow "If true" pattern. Also specified any events that are specific to searchable variant.

## Related issue
#664

## Checklist


packages/web-components/src/components/ic-select/ic-select.tsx
```diff
@@ -134,7 +134,7 @@ export class Select {
   @Prop() validationText?: string = "";
 
   /**
-   * The value of the currently selected option.
+   * The value of the select, reflected by the value of the currently selected option.
```
jd239, line 137: I'm not sure if this covers all bases because a searchable value may not come a selected option.

jd239 (changes requested):
Just a single comment. :)

GCHQ-Developer-112 (approved):

jd239 (approved):

Issue #707, opened 2023-05-22
[ic-pagination]: Add prop to programmatically control current page


jd239:
## Summary
It is not possible to programmatically set the current page the pagination is on.

## 💬 Description
Pagination sets the current page internally. This cannot be controlled externally. Suggestions from the community is that it should be. The user would pass a number to a prop such as `currentPage` or `page` which will update the state of the pagination component.

## 💰 Use value
There is an additional use case for the pagination which has been suggested by the community. 

## Additional info
N/A

PR #708, opened 2023-05-22
feat(web-components): adds method to set current page to pagination

MI6-451:
## Summary of the changes
Adds a public method to the pagination component to allow current page to be set.

## Related issue
#707 

## Checklist

### DIFF
--- a/packages/web-components/src/components/ic-pagination/ic-pagination.tsx
+++ b/packages/web-components/src/components/ic-pagination/ic-pagination.tsx
@@ -9,6 +9,7 @@ import {
   Watch,
   Event,
   EventEmitter,
+  Method,
 } from "@stencil/core";
 import paginationNextPrevious from "../../assets/pagination-next-previous.svg";
 import paginationFirstLast from "../../assets/pagination-first-last.svg";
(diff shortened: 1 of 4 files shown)


GCHQ-Developer-530 (approved):

GCHQ-Developer-112 (dismissed):

jd239:
Is it possible to make this a prop as well? Then get the public method to adjust the prop when using with pagination bar?

packages/web-components/src/components/ic-pagination/ic-pagination.tsx
```diff
@@ -84,6 +88,17 @@ export class Pagination {
     this.icPageChange.emit({ value: this.currentPage });
   }
 
+  @Method()
```
jd239, line 91: Is this a public method or will it be internal? If it is public, there needs to be a description so it appears within the guidance site props table...

jd239:
Can use this ticket https://github.com/mi6/ic-design-system/issues/464 to add `currentPage` example

jd239 (approved):

GCHQ-Developer-112 (approved):

Issue #653, opened 2023-05-05
Z-index is not correct for sticky ic-page-header


GCHQ-Developer-741:
## Summary of the bug
Spotted on the guidance site, the page header is sticky and should be shown above any page content. However, the ic-menu is displayed over it when it is opened and the page is scrolled.

## 🪜 How to reproduce
1. Go to https://design.sis.gov.uk/components/select.
2. Click the select example on the guidance tab so that the menu is open.
3. Scroll down whilst keeping the menu open.
4. The menu will appear above the sticky page header.

## 🧐 Expected behaviour
A sticky page header should cover all page content.

## 📸 Screenshots or code
![Screenshot from 2023-05-05 16-01-41](https://user-images.githubusercontent.com/117368893/236495128-db1b3293-2500-4533-b74b-970f877f13ae.png)

## Additional info
Will likely need a new design token for a sticky page header z-index. Can then add that to page header when the sticky prop is applied.

samezzz:
I'll like to work on this issue

PR #783, opened 2023-06-09
fix(web-components): added Z index token for sticky header

MI6-806:
## Summary of the changes
Added Z index token for sticky header, select menu no longer overlaps.

## Related issue
Issue #653

## Checklist
- Created new Z index token in variables.css
- Added token to sticky headers in ic-page-header.css


GCHQ-Developer-530 (dismissed):

packages/web-components/src/global/variables.css
```diff
@@ -293,6 +293,7 @@
   /* z-index */
   --ic-z-index-base-value: 0;
   --ic-z-index-page-header: calc(var(--ic-z-index-base-value) + 10);
+  --ic-z-index-page-header-sticky: calc(var(--ic-z-index-base-value) + 120);
```
ad9242, line 296: minor one: could it be moved lower down, so that the they values are in ascending order?

ad9242 (changes requested):
Has this been tested with a dialog displayed? I just wonder if the sticky page header will show overlayed above the dialog content?

MI6-806:
Not sure what actually happened here, think something went wrong?

Issue #674, opened 2023-05-11
[ic-top-navigation]: Update to grouped navigation behaviour


jd239:
## Summary
Update grouped navigation behaviour based on https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html

## 💬 Description
At the moment, grouped navigation items display on hover. The following article from W3 suggests:
- Adding a delay before showing the grouped navigations
- Adding a dismissable action via the keyboard if mouse pointer is still over navigation items or group navigation item

## 💰 Use value
This provides an accessibility enhancement to the top navigation.

## Additional info
- https://www.nngroup.com/articles/mega-menus-work-well/
- https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html

At the moment, the grouped navigation items appear when:
- Mouse hovers over group navigation item (without delay)
- Focusing on group navigation item and toggling

PR #787, opened 2023-06-12
Group nav changes

gd2910:
## Summary of the changes
Group-navigation has been changed to apply new WCAG features;

.5 second delay when hovering a grouped-nav-item (when mouse is idle) before dropdown.
delay not applied when mouse enters sibling grouped-nav-item.
dropdown can be cancelled with escape, space or enter (without focus).
.5 second delay after mouse leaves grouped-nav-item before dropdown is closed

## Related issue
#674 

## Checklist

### DIFF
--- a/packages/web-components/src/components/ic-navigation-group/ic-navigation-group.spec.ts
+++ b/packages/web-components/src/components/ic-navigation-group/ic-navigation-group.spec.ts
@@ -6,6 +6,18 @@ import { TopNavigation } from "../ic-top-navigation/ic-top-navigation";
 import { DEVICE_SIZES } from "../../utils/helpers";
 import { NavigationGroup } from "./ic-navigation-group";
 import * as helpers from "../../utils/helpers";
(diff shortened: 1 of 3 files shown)


ad9242:
getting some console errors in a couple of functions: triggerShowDropdown & handleMouseLeave

image

GCHQ-Developer-530 (dismissed):

ad9242:
Just a little comment on the behaviour. It seems a bit strange to me that the menu pops up after 0.5 seconds if I accidentally mouse over the nav group, or if I move my mouse onto a nav group item (when the menu\subitems are not visible) then quickly off. Just want to check with @DBD324 that is the intended behaviour.

gd2910:
That's not intended behaviour, the mouseIdle function should be catching it and only opening once the mouse has been idle on the element for 0.5, I'll take a look and see what's happening

GCHQ-Developer-530 (dismissed):

GCHQ-Developer-530 (approved):

Issue #799, opened 2023-06-19
Disabled=false causes screen readers to read out dimmed


GCHQ-Developer-530:
## Summary of the bug
When disabled is set to false, VoiceOver still reads out dimmed.

## 🪜 How to reproduce
More details internally, but investigate button disabled, text field disabled and select disabled.

## 🧐 Expected behaviour
Should not read out dimmed.

## Additional info
Check full info internally if you can't reproduce the issue

PR #808, opened 2023-06-22
fix(web-components): remove disabled attribute from components when disabled is false

GCHQ-Developer-530:
## Summary of the changes
Remove disabled attribute from components when disabled is false so that it doesn't get passed down to the native element causing dimmed to be read by screen readers

## Related issue
#799

## Checklist


GCHQ-Developer-847:
I wonder if you could put the if statement into a helper function and use that instead? It would only save a couple of lines of code for each component but could be worth it

MI6-451 (approved):

GCHQ-Developer-847 (approved):

Issue #735, opened 2023-06-06
Component border colours don't all meet contrast requirements and are inconsistent


DBD324:
## Summary of the bug
Many components use different border colours which makes the set inconsistent. Some also use border colours that don't meet the 3:1 contrast requirement that aims to differentiate UI component bounding boxes from adjacent background colours. 

WCAG 2.1 1.4.1 Non-text Contrast extract:
"For active controls any visual information provided that is necessary for a user to identify that a control is present and how to operate it must have a minimum 3:1 contrast ratio with the adjacent colors. Also, any visual information necessary to indicate state, such as whether a component is selected or focused must also ensure that the information used to identify the control in that state has a minimum 3:1 contrast ratio.

This Success Criterion does not require that changes in color that differentiate between states of an individual component meet the 3:1 contrast ratio when they do not appear next to each other. For example, there is not a new requirement that visited links contrast with the default color, or that mouse hover indicators contrast with the default state. However, the component must not lose contrast with the adjacent colors, and non-text indicators such as the check in a checkbox, or an arrow graphic indicating a menu is selected or open must have sufficient contrast to the adjacent colors."

Affected components are:
- Text fields
- Text areas
- Select
- Searchable select
- Multi-select
- Search bar
- Cards
- Dialogs
- Page header
- Popover menu
- Tabs

## 🧐 Expected behaviour
All components that use a border to indicate the size and shape of the component should use a border of Architectural 400 as this provides enough contrast to the default white background. 

Input components that use the component border to indicate interaction state should be update to use Action Dark Hover and Action Dark Pressed colours on the respective interaction states. 

Input components that show validation status using border colour should be updated to include hover and pressed states that maintain the validation colour. 

**New border tokens:**
Border default = Arch 400
Border Hover = Arch 500
Border Pressed = Arch 600
Border disabled = Arch 200

Border error default = Error red
Border error hover = Error red + 20% dark. 
Border error pressed = Error red + 40% dark

Border warning default = Warning amber
Border warning hover = Warning amber + 20% dark
Border warning pressed = Warning amber + 40% dark

Border success default = Success green
Border success hover = Success green + 20% dark
Border success pressed = Success green + 40% dark

Changes required:
- Cards - Current border 300 - Change to 400
- Dialog - Current border 200 - Change to 400
- Page header - Current border bottom 300 - Change to 400
- Popover menu - Current border 300 - Change to 400
- Tabs - Current border bottom 300 - Change to 400
[...]

PR #813, opened 2023-06-23
Feature/accordion component

j212122:
## Summary of the changes
Creates accordion and accordion-group

## Related issue
#240 

## Checklist
- [ x] I have added relevant unit and visual regression tests.
- [ x] I have manually accessibility tested any changes, if relevant.
- [ x] I have ensured any changes match the Figma component library.


packages/web-components/src/utils/types.ts
```diff
@@ -28,6 +28,8 @@ export enum IcInformationStatus {
   Success = "success",
 }
 
+export type IcSizes = "default" | "large" | "small";
```
GCHQ-Developer-741, line 31: I believe this type is also used in IcButton aswell. Would be worth updating it there as well

gd2910:
New components should adhere to the Stencil style guide;
Tests in a 'test' sub-directory
Each test in its own semantically named directory

Component files should also follow this pattern;
Variables (alphabetised)
Element
State (alphabetised)
Prop (alphabetised)
Watch (with watched props behind them)
Event (alphabetised)
Lifecycle (folowing lfiecycle process: connection, load, update, render)
Listen
Method
Local private methods
Render

packages/web-components/src/components/ic-accordion-group/ic-accordion-group.tsx
```diff
@@ -0,0 +1,113 @@
...
+  /**
+   * The size of the accordion.
+   */
+  @Prop() size?: IcSizes = "default";
+  /**
+   * If `true`, the side navigation will load in an expanded state.
+   */
+  @Prop({ mutable: true }) expanded: boolean = false;
+  /**
+   * If `true`,  only one accordion will open at a time.
```
GCHQ-Developer-530, line 42: Only small, but looks like there's more than one space between the comma and only

packages/web-components/src/components/ic-accordion/ic-accordion.css
```diff
@@ -0,0 +1,100 @@
...
+}
+
+button:focus {
+  box-shadow: var(--ic-border-focus);
+  border-radius: var(--ic-border-radius);
+  transition: var(--ic-transition-duration-fast);
+}
+
+.icon-container {
+  margin: 0 8px 0 0;
```
GCHQ-Developer-530, line 71: Can this 8px be changed to a variable?

packages/web-components/src/components/ic-accordion/ic-accordion.css
```diff
@@ -0,0 +1,100 @@
...
+}
+
+.expand-chevron {
+  width: var(--ic-space-lg);
+  height: var(--ic-space-lg);
+  margin-left: 40px;
```
GCHQ-Developer-530, line 77: Can this 40px be changed to a calculation with variables?

packages/web-components/src/components/ic-accordion/ic-accordion.css
```diff
@@ -0,0 +1,100 @@
...
+  transition: height 0.3s ease-out;
+}
+
+.expanded-content-inner {
+  padding: var(--ic-space-xs);
+}
+
+.expanded-content-open {
+  display: flex;
+  border-bottom: 1px solid var(--ic-architectural-300);
```
GCHQ-Developer-530, line 99:
```suggestion
  border-bottom: var(--ic-space-1px) solid var(--ic-architectural-300);
```

packages/web-components/src/components/ic-accordion/ic-accordion.tsx
```diff
@@ -0,0 +1,130 @@
...
+  /**
+   * The section header outlining section content.
+   */
+  @Prop() heading?: string = "";
+  /**
+   * The main body message of the card.
+   */
+  @Prop() message?: string = "";
+  /**
+   * If true accordion appears expanded.
```
GCHQ-Developer-530, line 33:
```suggestion
   * If `true`, the accordion appears expanded.
```

packages/web-components/src/components/ic-accordion/ic-accordion.tsx
```diff
@@ -0,0 +1,130 @@
...
+  /**
+   * The size of the accordion.
+   */
+  @Prop() size?: IcSizes = "default";
+  /**
+   * If `true`, the card will be disabled if it is clickable.
+   */
+  @Prop() disabled?: boolean = false;
+
+  @Event() accordionClicked: EventEmitter<{ id: string }>;
```
GCHQ-Developer-530, line 49: Is this an event we want people to be able to access? If so, it will need to be renamed to icAccordionClicked. If not, it will need an @internal.

GCHQ-Developer-530 (changes requested):
Looks really good, just a couple of small comments 😊

packages/web-components/src/components/ic-accordion/ic-accordion.css
GCHQ-Developer-530: Actually might want to check with @DBD324 that this is correct since the recent changes for updating borders to architectural 400. If the case is that it should be updated to 400, then you can use the --ic-border-default variable

packages/web-components/src/components/ic-back-to-top/readme.md
```diff
@@ -12,13 +12,6 @@
 | `target` _(required)_ | `target`  | The ID of the element to jump back to when the link is clicked. | `string` | `undefined` |
 
 
-## CSS Custom Properties
-
-| Name                       | Description            |
-| -------------------------- | ---------------------- |
-| `--ic-z-index-back-to-top` | z-index of back to top |
-
-
```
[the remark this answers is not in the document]
j212122: This pops in and out sometimes, not sure why. I've re-added

GCHQ-Developer-530 (changes requested):
Found a small bug where if you click see all, but then manually close the accordions, it still says hide all even though they are all hidden. 

https://github.com/mi6/ic-ui-kit/assets/111882034/fe00866a-777c-42c8-afb9-574f29790958

packages/web-components/src/components/ic-accordion/ic-accordion.css
```diff
@@ -0,0 +1,95 @@
+@import "../../global/normalize.css";
+
+:host {
+  display: block;
+  border-bottom: 1px solid var(--ic-architectural-300);
```
GCHQ-Developer-530, line 5:
```suggestion
  border-bottom: var(--ic-border-default);
```

packages/web-components/src/components/ic-accordion/ic-accordion.css
```diff
@@ -0,0 +1,95 @@
...
+.section-button.small {
+  padding: var(--ic-space-xxs) var(--ic-space-xs);
+}
+
+.section-button.large {
+  padding: var(--ic-space-sm) var(--ic-space-xs);
+}
+
+:host(:first-child) {
+  border-top: 1px solid var(--ic-architectural-300);
```
GCHQ-Developer-530, line 30:
```suggestion
  border-top: var(--ic-border-default);
```

GCHQ-Developer-530 (changes requested):
Spoke to designers and the borders should use the border default variable set up in https://github.com/mi6/ic-ui-kit/issues/735. Other than that, looks great!

packages/web-components/src/components/ic-accordion-group/ic-accordion-group.tsx
```diff
@@ -0,0 +1,143 @@
...
+  @Listen("accordionClicked")
+  handleAccordionClicked(event: CustomEvent) {
+    if (this.singleExpansion) {
+      this.accordions.forEach((acc) => {
+        acc.expanded = false;
+      });
+      const singleAccordion = this.accordions.find(
+        (acc) => acc.id === event.detail.id
+      );
+      singleAccordion.expanded = true;
```
GCHQ-Developer-741, line 83: I may have missed this part of the AC, but when singleExpansion is enabled, I'm seeing behaviour where I can't close the one i've opened once it has been opened. Is this correct? My understanding was that only one could be open at once, but all could be closed at the same time. Also noticing some flickering when clicking the other accordion when it closes them all.
[Screencast from 18-07-23 10:47:54.webm](https://github.com/mi6/ic-ui-kit/assets/117368893/dc91fd63-6237-4a98-947c-ef6b727e7978)

GCHQ-Developer-741: I think if this is the correct behaviour maybe hiding the closing animation that appears before it immediately opens again, as it may throw people off and have them thinking it didn't register their clicks.

packages/web-components/src/components/ic-accordion/ic-accordion.stories.mdx
```diff
@@ -0,0 +1,374 @@
...
+      </ic-accordion>`}
+  </Story>
+</Canvas>
+
+### Icon
+
+<Canvas>
+  <Story name="Icon">
+    {html` <ic-accordion heading="Accordion 1">
+        <svg
```
GCHQ-Developer-847, line 37: The icon isn't quite aligned to the centre of the text vertically:



Also, just thinking if someone tries to pass in an icon which has a random height and width - should there be a set height and width applied to the icon so whatever size icon they pass in, it always appears at the correct size?

Thinking maybe an 18px height + width on the SVG and 24px on the `icon-container` div to match the designs? :)

GCHQ-Developer-847 (changes requested):
Added a few comments - everything else looks great though!

gd2910:
Just a small one, looks like the background colour for 'light text' and 'parent' stories are mixed up.

gd2910:
There are a couple of private variables that can be brought up to the top of the class in `ic-accordion.tsx`
```
 private animationTimer: ReturnType<typeof setTimeout>;
  private expandedContentEl: HTMLDivElement;
```

gd2910 (approved):

GCHQ-Developer-530 (changes requested):
The disabled accordion seems to still have a click state, I think the behaviour should be that it shouldn't change background colour when clicked.

https://github.com/mi6/ic-ui-kit/assets/111882034/564c4850-e47e-4234-a5a7-ac983fb41496

GCHQ-Developer-530 (approved):

PR #855, opened 2023-06-30
Data table test

MI6-451:
## Summary of the changes
Test branch for data table accessibility testing

## Related issue
Tell us the issue number. If suggesting an improvement or component, please discuss it with us in an issue first.

## Checklist


jd239:
I'm wondering if it worth updating the `cellAlignment` and `alignment` depending on where it is being used. For example:
- `cellAlignment` is being used for aligning a column so should that ne updated to `columnAlignment`,
- `cellAlignment` is used within the row header, so should that be `rowAlignment`
- `alignment` is used for cell alignment so should that be `cellAlignment`

I guess we can use the parent key (i.e. if parent key is `header` and has `rowAlignment` then add CSS. If there is no key (or is passed through the column prop), do something with columnAlignment) get the value and apply the CSS class.

packages/web-components/src/components/ic-pagination-bar/ic-pagination-bar.tsx
```diff
@@ -0,0 +1,515 @@
...
+      this.TOOLTIP
+    ) as HTMLIcTooltipElement;
+    const textField = this.el.shadowRoot.querySelector(
+      this.TEXT_FIELD
+    ) as HTMLIcTextFieldElement;
+
+    if (ev.key === "Enter") {
+      if (textField.validationStatus === "error") {
+        tooltip.displayTooltip(false, false);
+        setTimeout(() => tooltip.displayTooltip(true, true), 50);
```
jd239, line 258: Why is this hiding and then displaying the tooltip again?

packages/web-components/src/components/ic-pagination-bar/ic-pagination-bar.tsx
```diff
@@ -0,0 +1,515 @@
...
+      } else {
+        this.goToPage();
+      }
+    } else if (ev.key === "Backspace") {
+      tooltip.displayTooltip(false, true);
```
jd239, line 263: Does this need to persist if you are amending the value while it is not longer on show? I might not be testing this correctly.

packages/web-components/src/components/ic-pagination-bar/ic-pagination-bar.tsx
```diff
@@ -0,0 +1,515 @@
...
+                  small
+                  label="items-per-page-input"
+                  class="items-per-page-input"
+                  hideLabel
+                  options={itemsPerPageOptions}
+                  value={this.itemsPerPage}
+                  onIcChange={() => this.changeItemsPerPage()}
+                ></ic-select>
+              </div>
+            ) : null}
```
jd239, line 428: No need to return `null` if nothing is going to be rendered, just use `{showItemsPerPageControl && (<element>)`. There are a couple instances where this can be applied in pagination and data-table

packages/web-components/src/components/ic-pagination-bar/ic-pagination-bar.tsx
```diff
@@ -0,0 +1,515 @@
...
+              <ic-tooltip
+                label={this.inputError}
+                target="#go-to-page-input"
+                disableHover
+                disableClick
+              >
+                <ic-text-field
+                  type="number"
+                  small
+                  label="go-to-page-input"
```
jd239, line 486: 'go-to-page-input' string is used in a few places to can be private variable at the top of the page.

jd239 (changes requested):
There is a lot of repeat coded which can be moved into functions just to make it less daunting. I've added some comments.

jd239:
What should happens when looping through the sort options and returning back to `unsorted`? Should the data go back to it's original order?

packages/web-components/src/components/ic-data-table/ic-data-table.tsx
```diff
@@ -0,0 +1,651 @@
...
+  /**
+   * Column headers will remain at the top of the table when scrolling vertically if set to `true`.
+   */
+  @Prop() stickyColumn?: boolean = false;
+
+  /**
+   * Row headers will remain to the left when scrolling horizontally if set to `true`.
+   */
+  @Prop() stickyRow?: boolean = false;
+
```
gd2910, line 112: As above but stickyRowHeaders?

packages/web-components/src/components/ic-data-table/ic-data-table.css
```diff
@@ -0,0 +1,211 @@
...
+  position: relative;
+  height: 100%;
+  overflow: auto;
+  transition: var(--ic-easing-transition-fast);
+  overflow-anchor: none;
+}
+
+:host([show-pagination="true"]) .table-row-container {
+  height: calc(100% - 57px);
+}
```
gd2910, line 27: Is it worth adding a global pagination-spacer variable?

GCHQ-Developer-530:
I think the sort icon should appear below the column header for mobile
Screenshot 2023-08-01 at 10 19 52

packages/web-components/src/components/ic-pagination-bar/ic-pagination-bar.tsx
```diff
@@ -0,0 +1,517 @@
...
+              <div class="items-per-page-holder">
+                <ic-typography
+                  class={{
+                    [`pagination-text-${appearance}`]: true,
+                    ["items-per-page-control-label"]: true,
+                  }}
+                  variant="label"
+                >
+                  {this.itemLabel}s per {this.pageLabel.toLowerCase()}
+                </ic-typography>
```
jd239, line 417: Would expect clicking on the label focuses on the select

jd239 (changes requested):
Second review complete. Some new comments. Nice update 👍

packages/web-components/src/components/ic-data-table/ic-data-table.tsx
```diff
@@ -0,0 +1,475 @@
...
+    } = this;
+
+    return (
+      <div class="table-container">
+        <div
+          class={{
+            ["table-row-container"]: true,
+            ["scrollable"]: scrollable,
+          }}
+          tabIndex={scrollable ? 0 : null}
```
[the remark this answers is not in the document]
GCHQ-Developer-741: Is still focusable with -1. Setting it to null prevents this

packages/web-components/.loki/reference/chrome_laptop_Web_Components_Data_table_Complex_example.png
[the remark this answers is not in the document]
GCHQ-Developer-741: No because they are disabled. If they could be interacted with it would constitute a failure

GCHQ-Developer-530:
More of a question really, but should the row highlighting be removed when you click away from the table?

packages/web-components/src/components/ic-data-table/ic-data-table.css
```diff
@@ -0,0 +1,222 @@
...
+.table-row-container:focus {
+  outline: none;
+  box-shadow: var(--ic-border-focus);
+  border-radius: var(--ic-border-radius);
+  transition: var(--ic-easing-transition-fast);
+  z-index: 1;
+}
+
+:host([embedded="true"]) .table-container {
+  border: solid var(--ic-architectural-200) var(--ic-space-1px);
```
GCHQ-Developer-530, line 40: Is it worth creating a new border variable for this? I think it's used a few times in navigation menu etc.

packages/web-components/src/components/ic-data-table/ic-data-table.css
```diff
@@ -0,0 +1,222 @@
...
+  font-weight: var(--ic-font-weight-bold);
+}
+
+.column-header,
+.row-header {
+  text-align: left;
+  padding: var(--ic-space-xs);
+  background-color: var(--ic-architectural-40);
+  border-right: solid var(--ic-architectural-200) var(--ic-space-1px);
+  border-bottom: solid var(--ic-architectural-200) var(--ic-space-1px);
```
GCHQ-Developer-530, line 117: These could both use the new border variable too

packages/web-components/src/components/ic-data-table/ic-data-table.css
```diff
@@ -0,0 +1,222 @@
...
+  text-align: left;
+  vertical-align: top;
+}
+
+.data-type-number {
+  text-align: right;
+  vertical-align: top;
+}
+
+.cell-alignment-left {
```
GCHQ-Developer-530, line 167: Can this be moved to under line 134? They both only contain text-align: left

packages/web-components/src/components/ic-data-table/ic-data-table.css
```diff
@@ -0,0 +1,222 @@
...
+  text-align: left;
+}
+
+.cell-alignment-right {
```
GCHQ-Developer-530, line 171: Can this be moved to under line 139? They both only contain text-align: right

packages/web-components/src/components/ic-data-table/ic-data-table.css
```diff
@@ -0,0 +1,222 @@
...
+  text-align: right;
+}
+
+.cell-alignment-center {
```
GCHQ-Developer-530, line 175: Can this be moved to under line 144? They both only contain text-align: center

packages/web-components/src/components/ic-data-table/ic-data-table.tsx
```diff
@@ -0,0 +1,493 @@
...
+   */
+  @Prop() data: { [key: string]: any }[];
+
+  /**
+   * Set the density of the table including font and padding.
+   */
+  @Prop() density?: IcDataTableDensityOptions = "default";
+
+  /**
+   * Applies a border
```
GCHQ-Developer-530, line 69:
```suggestion
   * Applies a border to the table.
```

packages/web-components/src/components/ic-data-table/ic-data-table.tsx
```diff
@@ -0,0 +1,493 @@
...
+
+  /**
+   * The row content for the table.
+   */
+  @Prop() data: { [key: string]: any }[];
+
+  /**
+   * Set the density of the table including font and padding.
+   */
+  @Prop() density?: IcDataTableDensityOptions = "default";
```
GCHQ-Developer-530, line 66: Might be something to discuss with design, but does this need to be renamed to size? It seems a bit inconsistent as we don't have a density prop on any other components, and the size prop tends to only affect the padding/font too. We could also then use the IcSizes type as it would need to be small, default, large instead of dense, default, spacious.

GCHQ-Developer-530 (changes requested):

packages/web-components/src/components/ic-data-table/ic-data-table.tsx
GCHQ-Developer-741: Have previously discussed this with design. This has deemed to be more reflective of the actual changes being applied, whereas 'sizes' is quite misleading

packages/web-components/src/components/ic-data-table/ic-data-table.css
```diff
@@ -0,0 +1,213 @@
...
+  color: var(--ic-color-secondary-text);
+}
+
+.cell-emphasis-high {
+  font-weight: 700;
+}
+
+.pagination-container {
+  background-color: var(--ic-architectural-40);
+  border-top: solid var(--ic-architectural-200) var(--ic-space-1px);
```
GCHQ-Developer-530, line 192: This can use the new border light variable

packages/web-components/src/components/ic-data-table/ic-data-table.tsx
```diff
@@ -0,0 +1,498 @@
...
+
+  /**
+   * Applies a border to the table container
```
GCHQ-Developer-530, line 69:
```suggestion
   * Applies a border to the table container.
```

packages/web-components/src/components/ic-data-table/ic-data-table.tsx
```diff
@@ -0,0 +1,498 @@
...
+        rowValues[headerIndex],
+        "rowAlignment"
+      );
+      rowEmphasis = this.getObjectValue(rowValues[headerIndex], "emphasis");
+    }
+    return rowValues.map((cell, index) => {
+      const column = this.columns[index];
+      const cellValue = (key: string) => this.getObjectValue(cell, key);
+
+      const getAlignment = (alignment: "horizontal" | "vertical") => {
```
jd239, line 200: Feel as though this can be outside of the `map` and pass through `cell` in addition to `alignment`. This will also make it possible to unit test

jd239 (changes requested):
Only 1 comment.

packages/web-components/src/components/ic-data-table/ic-data-table.tsx
GCHQ-Developer-741: Not sure this is necessary. This code is only repeated in this specific map function, and contains values that are only within the scope of the map function. Keeping it here would prevent unnecessarily passing through the same value multiple times.

jd239: How would you unit test the function?

jd239 (approved):

GCHQ-Developer-530 (approved):

Issue #751, opened 2023-06-07
Badge


ASM995:
## Summary
Small component to add additional meaning to other elements such as notifications or sensitive info.

## 💰 User value
Describe why this is important and how it will help users (either developers, designers or the end-users). Think about how it makes people's lives easier, more accessible or quicker to build stuff.

## 📚 User Stories
If relevant, describe the high-level functionality as user stories.

*As an* ICDS user: 
*I need* a badge component
*So that* I can attach additional meaning to other components, such as notification number or a lock symbol. 

*As an* ICDS user
*I need* a badge component that can show text, icon, or simple dot
*So that* I can cater for different situations. 

*An an* ICDS user
*I need* to be able to chose from a number of different colours
*So that* I can attach it to different components. 

## 📝 Acceptance Criteria 
If relevant, describe in full detail the different interactions and edge cases that the component or patterns needs to fulfil.

*Given*
*When* 
*Then*

**Type**
*Given* that a badge is being implemented
*When* setting its properties
*Then* it's type can be chosen from the following:
- Dot - No contents, a simple small shape that can indicate the presence of something. 
- Text (default) - A text label within a normal sized badge. This can show the number of things present. 
- Icon - An icon within a normal sized badge. This can be used to indicate something visually, such as a padlock icon. 

**Colour**
*Given* that a badge is being implemented
*When* setting its props
*Then* an option is available to set its colour to one of the following:
- Success green
- Warning amber
- Error red
- Info blue
- Neutral - arch 500
- Light grey - arch 40
- Custom (see next criteria)
*And* the foreground colour will adjust to either white or primary text to ensure colour contrast ratios (see Figma). 

*Given* that a badge is being implement with a custom colour
*When* entering a custom colour
*Then* the foreground colour should automatically determine whether to use white text or primary text colours, where the higher contrast option is selected. 

**Size**
*Given* that a badge is being implemented
*When* settings its props
*Then* an option is available to set the size of the badge to one of the following:
- Default (default)
- Small
- Large

**Alignment/Position**
*Given* that a badge is being implemented
*When* setting its props
*Then* an option is available to set the type of positioning to one of the following:
- Square cornered (default) - far
- Rounded corner - near
- Inline with text. 

*Given* that a badge is being implemented with 'squared corner' positioning, 
*When* rendered
*Then* the badge should appear to the top right of the parent component and overlap the square corner. (See Figma)

*Given* that a badge is being implemented with 'rounded corner' positioning, 
[...]

PR #856, opened 2023-06-30
751 badge component

GCHQ-Developer-530:
## Summary of the changes
Create badge component, and modify button, chip and tab components to have badge slot and "position: relative" to show position far, near and inline usage.

## Related issue
#751 

## Checklist


packages/web-components/src/components/ic-badge/ic-badge.tsx
```diff
@@ -0,0 +1,225 @@
...
+  private foregroundColour: IcThemeForeground;
+
+  @Element() el: HTMLIcBadgeElement;
+
+  @State() visible: boolean = true;
+
+  /**
+   * The accessible label of the badge component to provide context for screen reader users.
+   */
+  @Prop() accessibleLabel?: string;
```
GCHQ-Developer-847, line 46: This causes the screen reader to read out "badge displaying undefined" if it's not provided - should it be a required prop?

packages/web-components/src/components/ic-badge/ic-badge.tsx
```diff
@@ -0,0 +1,225 @@
...
+        label = this.textLabel;
+      }
+    }
+    return label;
+  };
+
+  private setAccessibleLabel = () => {
+    if (this.accessibleLabel !== null && getParentElement(this.el) !== null) {
+      getParentElement(this.el).ariaLabel =
+        "badge displaying " + this.accessibleLabel;
```
GCHQ-Developer-847, line 197: On tabs, this aria label replaces the label of the tab. E.g. in the "Position inline" story, it just says "badge displaying undefined" and doesn't say the word "Small" which is the visible label of the tab

packages/web-components/src/components/ic-badge/ic-badge.stories.mdx
```diff
@@ -0,0 +1,982 @@
...
+          ><ic-badge
+            text-label="1"
+            slot="badge"
+            position="near"
+            size="large"
+          ></ic-badge
+        ></ic-chip>
+      </div>
+      <div style="padding:16px;">
+        <ic-chip label="Small Dot" style="margin-right:16px;"
```
GCHQ-Developer-847, line 853: I'm not sure if the positioning of the badge on the chips matches the Figma designs - e.g. the badge isn't touching the chip on this one

Also - don't know if it's possible to do this, but does the positioning also need to depend on the size of the chip? If you change the size, the badges moves closer / further away

GCHQ-Developer-847 (changes requested):
Looks really good :) just a few comments

packages/web-components/src/components/ic-badge/ic-badge.tsx
```diff
@@ -0,0 +1,225 @@
...
+   * The custom badge colour. This will only style the badge component if variant="custom".
+   * Can be a hex value e.g. "#ff0000", RGB e.g. "rgb(255, 0, 0)", or RGBA e.g. "rgba(255, 0, 0, 1)".
+   */
+  @Prop() customColor?: string = null;
+
+  /**
+   * The maximum number shown on your badge appended with a +.
+   * This will only be displayed if type="text" and textLabel is not empty.
+   */
+  @Prop() maxNumber?: number;
```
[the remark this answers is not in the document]
GCHQ-Developer-530: This was discussed with @DBD324 and it was decided that we would implement it like this and add into the guidance to ensure users implement this prop correctly, e.g not using it for 7+

packages/web-components/src/components/ic-badge/ic-badge.tsx
```diff
@@ -0,0 +1,225 @@
...
+
+  /**
+   * The size of the badge to be displayed.
+   */
+  @Prop() size?: IcSizes = "default";
+
+  /**
+   * The text displayed in the badge. This will only be displayed if type="text".
+   */
+  @Prop() textLabel?: number;
```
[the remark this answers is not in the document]
GCHQ-Developer-530: Not sure, the AC was:
**Given that a text badge is being implemented
When setting it properties
Then the text label can be set to any number.**
I took that to be that it could only be set to a number as there was no ACs around other text.
Might be worth @DBD324 weighing in on this.

packages/web-components/src/components/ic-badge/ic-badge.stories.mdx
GCHQ-Developer-530: I think it's correct that the badge isn't touching the chip on this. From the Figma designs, it's based on the focus around the element and so from the screenshot below, it should be in the white part of the focus border into the light blue, not sitting on the element itself.
Screenshot 2023-07-07 at 13 17 29

Not sure about the positioning depending on the size of the chip, I'd think this was up to the developer to decide which size badge to use dependent on the size of the chip, but happy to try and look into changing the positioning based on size if you think that's what should happen.

packages/web-components/src/components/ic-badge/ic-badge.tsx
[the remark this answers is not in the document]
DBD324: Good question, I think the primary use case is for numbers, but can see how something like new could be useful too. We'll include in guidance detail about avoiding long labels. Maybe we should keep it flexible and allow text to be entered too. Would you suggest keeping its type 'string' with numbers being able to be added, or would we need to allow the change of type from number to string?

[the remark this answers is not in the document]
GCHQ-Developer-847: I think change the type of the prop to 'string' as that will allow numbers and letters (I think it would bring up an error otherwise when trying to pass in a letter). Then in the code it can be converted to a number where necessary - to carry out calculations for the max number etc. (if the value of the string is actually a number)

packages/web-components/src/components/ic-badge/ic-badge.tsx
```diff
@@ -0,0 +1,237 @@
...
+  private foregroundColour: IcThemeForeground;
+
+  @Element() el: HTMLIcBadgeElement;
+
+  @State() visible: boolean = true;
+
+  /**
+   * The accessible label of the badge component to provide context for screen reader users.
+   */
+  @Prop() accessibleLabel?: string;
```
GCHQ-Developer-847, line 47: I'm still getting 'badge displaying undefined' on the buttons when prop this isn't provided. 

And on the tabs, it reads out the tab label now but then reads out just the number on the badge e.g. "Small 1" for the first tab

packages/web-components/src/components/ic-badge/ic-badge.tsx
```diff
@@ -0,0 +1,238 @@
...
+    return label;
+  };
+
+  private setAccessibleLabel = () => {
+    if (getParentElement(this.el) !== null) {
+      if (getParentElementType(this.el) === "IC-TAB") {
+        getParentElement(this.el).setAttribute("aria-describedby", "badge");
+      } else {
+        getParentElement(this.el).ariaLabel =
+          "badge displaying " + this.accessibleLabel;
```
jd239, line 203: `badge displaying ${this.accessibleLabel}`

packages/web-components/src/components/ic-badge/ic-badge.tsx
```diff
@@ -0,0 +1,238 @@
...
+          [`${type}`]: true,
+          [`foreground-${foregroundColour}`]: foregroundColour !== null,
+          ["show"]: visible,
+          ["hide"]: !visible,
+        }}
+        id="badge"
+        aria-label={
+          this.accessibleLabel !== null &&
+          this.accessibleLabel !== undefined &&
+          "badge displaying " + this.accessibleLabel
```
jd239, line 226: `badge displaying ${this.accessibleLabel}`

packages/web-components/src/components/ic-badge/test/a11y/ic-badge.test.a11y.ts
jd239, line 1: Can we add a test where we set the accessibleLabel prop

packages/web-components/src/components/ic-badge/test/basic/ic-badge.e2e.ts
jd239, line 1: I don't think this is required and is probably covered by the snapshots in the Unit tests

packages/web-components/src/components/ic-badge/test/basic/ic-badge.spec.tsx
jd239, line 1: I feel as thought we can add more detailed tests to this for example:
- setBadgeColour:
Check the correct values are being stored within `this.customColorRGBA`

- getBadgeForeground
Check the value within `this.foregroundColor`.

- getTextLabel
Check the correct label is returned depending on maxNumber is set or not

packages/web-components/src/components/ic-button/ic-button.tsx
jd239, line 1: Add badge slot description to top

packages/web-components/src/components/ic-chip/ic-chip.tsx
jd239, line 1: Add badge slot description to top

packages/web-components/src/components/ic-tab/ic-tab.tsx
jd239, line 1: Add badge slot description to top

jd239 (changes requested):
Great job! 👍  I have added a few comments.

packages/web-components/src/components/ic-badge/ic-badge.tsx
```diff
@@ -0,0 +1,243 @@
...
+      } else {
+        getParentElement(
+          this.el
+        ).ariaLabel = `badge displaying ${this.accessibleLabel}`;
+      }
+    }
+  };
+
+  private isAccessibleLabelDefined = () => {
+    return this.accessibleLabel !== undefined && this.accessibleLabel !== null
```
jd239, line 209: `true` or `false` is returned by the statement above so no need for line 210 & 211

jd239 (changes requested):
Two final comments,
- There is an update to the audit-ci file (root) within the web-components commit scope
- One comment on the isAccessibleLabelDefined method

jd239 (approved):

GCHQ-Developer-847 (approved):

Issue #791, opened 2023-06-14
Event listeners not removed


ad9242:
## Summary of the bug
Some components add event listeners, which are not correctly removed. The following list may not be exhaustive, so all components should be reviewed:

**Ic-side-navigation:** 
"transitionend" event is never removed

**ic-breadcrumb-group:**
collapsedBreadcrumbEl "click" event is never removed 

"transitionend" event is never removed

**ic-checkbox:**
"icChange" on associated textfield does not remove same bound function instance

**Ic-horizontal-scroll**
"focus", "mouseup", "mouseleave" events are never removed 

**ic-tab-context:**
"keydown" listeners for tabGroup not removing same bound function

## 🧐 Expected behaviour
Any event listeners added should also be removed correctly, to prevent memory leaks

## Additional Info
basically this is bad:
```
document.addEventListener("keydown", () => foo());
document.removeEventListener("keydown", () => foo());
```
as it does not remove the exact same function instance
 
instead you need to do the following:

```
keydownHandler() {
  foo();
}
document.addEventListener("keydown", this.keydownHandler);
document.removeEventListener("keydown", this.keydownHandler);
```

PR #858, opened 2023-07-03
fix(web-components): remove event listeners

GCHQ-Developer-530:
## Summary of the changes
Remove event listeners correctly for breadcrumb group, checkbox, horizontal scroll, side navigation and tab context

## Related issue
#791

## Checklist

### DIFF
--- a/packages/web-components/src/components/ic-breadcrumb-group/ic-breadcrumb-group.tsx
+++ b/packages/web-components/src/components/ic-breadcrumb-group/ic-breadcrumb-group.tsx
@@ -16,6 +16,7 @@ export class BreadcrumbGroup {
   private ADD_CLASS_DELAY = 50;
   private breadcrumb: HTMLIcBreadcrumbElement;
   private breadcrumbs: HTMLIcBreadcrumbElement[];
+  private collapsedBreadcrumbEl: any;
   private collapsedBreadcrumbs: HTMLIcBreadcrumbElement[];
   private collapsedBreadcrumbWrapper: HTMLIcBreadcrumbElement;
   private IC_BREADCRUMB: string = "ic-breadcrumb";
@@ -58,6 +59,14 @@ export class BreadcrumbGroup {
         }
       }
     }
+  }
+
+  disconnectedCallback(): void {
+    this.breadcrumb.removeEventListener(
+      "transitionend",
+      this.transitionendHandler
+    );
+    this.collapsedBreadcrumbEl.removeEventListener("click", this.clickHandler);
   }
 
   private setBackBreadcrumb = () => {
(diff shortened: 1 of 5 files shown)


packages/web-components/src/components/ic-side-navigation/ic-side-navigation.tsx
```diff
@@ -351,7 +356,7 @@ export class SideNavigation {
       secondaryNavigationWrapper.classList.add(collapsedIconLabelsStart);
     }
 
-    this.el.addEventListener("transitionend", () => {
+    const transitionEndHandler = () => {
```
ad9242, line 359: again, i think this handler needs to be moved outside of the scope of the `animateCollapsedIconLabels `function, as it will be creating a new instance of `transitionEndHandler`, every time `animateCollapsedIconLabels ` is called

ad9242 (changes requested):
i think just need to modify horizontal scroll & side navigation to be more like the way breadcrumb-group is implemented, to ensure the bound function\event is removed

packages/web-components/src/components/ic-breadcrumb-group/ic-breadcrumb-group.tsx
```diff
@@ -156,9 +171,7 @@ export class BreadcrumbGroup {
     collapsedBreadcrumbEl.id = "collapsed-ellipsis";
     collapsedBreadcrumbEl.innerText = "...";
     collapsedBreadcrumbEl.classList.add("collapsed-breadcrumb");
-    collapsedBreadcrumbEl.addEventListener("click", () => {
-      this.handleHiddenCollapsedBreadcrumbs(this.collapsedBreadcrumbWrapper);
-    });
+    collapsedBreadcrumbEl[method]("click", this.clickHandler);
```
ad9242, line 174: does this need to be applied in this way - can we just use addEventListener instead?

packages/web-components/src/components/ic-breadcrumb-group/ic-breadcrumb-group.tsx
```diff
@@ -128,7 +136,14 @@ export class BreadcrumbGroup {
     }
   };
 
-  private renderCollapsedBreadcrumb = () => {
+  private clickHandler = () => {
+    this.handleHiddenCollapsedBreadcrumbs(this.collapsedBreadcrumbWrapper);
+  };
+
+  private renderCollapsedBreadcrumb = (action: "add" | "remove") => {
```
ad9242, line 143: when run with `remove` passed as param, won't this still create the content in the DOM?

packages/web-components/src/components/ic-breadcrumb-group/ic-breadcrumb-group.tsx
```diff
@@ -60,6 +60,14 @@ export class BreadcrumbGroup {
     }
   }
 
+  disconnectedCallback(): void {
+    this.breadcrumb.removeEventListener(
+      "transitionend",
+      this.transitionendHandler
+    );
+    this.renderCollapsedBreadcrumb("remove");
```
ad9242, line 68: think you could just remove the bound `clickHandler` here rather than call the above func, as calling that will creating element in the DOM

packages/web-components/src/components/ic-side-navigation/ic-side-navigation.tsx
```diff
@@ -330,38 +334,34 @@ export class SideNavigation {
...
       ".bottom-wrapper > .secondary-navigation"
     );
 
-    const collapsedIconLabelsEnd = "collapsed-icon-labels-end";
-    const collapsedIconLabelsStart = "collapsed-icon-labels-start";
-
     if (primaryNavigationWrapper) {
-      primaryNavigationWrapper.classList.remove(collapsedIconLabelsEnd);
-      primaryNavigationWrapper.classList.add(collapsedIconLabelsStart);
+      primaryNavigationWrapper.classList.remove(
```
ad9242, line 347: just checking the right classes have been removed\added??? looks like they have been switched, is that intended?

ad9242 (changes requested):

GCHQ-Developer-847 (approved):

Issue #726, opened 2023-06-03
Side navigation item label still displayed when no href and collapsed


GCHQ-Developer-847:
## Summary of the bug
When there is no href on a navigation item in the side navigation component, the label is still visible when it is collapsed.

## 🪜 How to reproduce
1. Render a side navigation component with nav items that don't have an href.
2. See that the labels are always visible even when it is collapsed.

One example of when a developer may have a side navigation component without an href is if they are using React Router and click handlers - e.g. using the `useNavigate` hook and `onClick` on the navigation items, like https://ultimatecourses.com/blog/programmatically-navigate-react-router. Having tested this set-up, the side navigation all works fine (no page reloads) - this bug with the label is the only issue.

## 🧐 Expected behaviour
The labels should not be visible when the side navigation is collapsed.

## 📸 Screenshots or code
<img src="https://github.com/mi6/ic-ui-kit/assets/111882109/27594940-6f19-4968-ad48-8b2a7cbf1740" width="50">

## 🖥 Desktop
 - OS: MacOS
 - Browser: Chrome, Safari, Firefox

PR #861, opened 2023-07-07
fix(web-components): duplicate a.link styles for div.link on nav items

GCHQ-Developer-530:
## Summary of the changes
Duplicate a.link styles for div.link on nav items so that when there is no href the label doesn't show when collapsed

## Related issue
#726

## Checklist

### DIFF
--- a/packages/web-components/src/components/ic-navigation-item/ic-navigation-item.css
+++ b/packages/web-components/src/components/ic-navigation-item/ic-navigation-item.css
@@ -411,6 +411,10 @@ svg {
 :host(.navigation-item-side-nav.navigation-item-top-nav-child)
   a.link
   ic-typography,
+:host(.navigation-item-side-nav) div.link ic-typography,
+:host(.navigation-item-side-nav.navigation-item-top-nav-child)
+  div.link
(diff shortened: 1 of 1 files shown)


packages/web-components/src/components/ic-side-navigation/ic-side-navigation.stories.mdx
```diff
@@ -66,7 +66,7 @@ import readme from "./readme.md";
             </svg>
           </ic-navigation-item>
           <ic-divider slot="primary-navigation"></ic-divider>
-          <ic-navigation-item slot="primary-navigation" href="/" label="Trends">
+          <ic-navigation-item slot="primary-navigation" label="Trends">
```
GCHQ-Developer-847, line 69: Not sure if we need an example showing it works without an href? I think it might be ok if we just test this locally - just thinking if we come back to this story in future we might wonder why it doesn't have an href, or if we're demoing a change to the side nav and then this link doesn't work or something

But only a tiny thing 🤷‍♀️

GCHQ-Developer-847 (approved):

Issue #814, opened 2023-06-26
Include stencil-community eslint plugins


gd2910:
## Summary
Stencil-community has a set of es-lint rules that should be included in the ui-kit eslint rules.
 
## 💬 Description
This ticket includes some investigation into the breadth of the rules in [stencil-community](https://www.npmjs.com/package/@stencil-community/eslint-plugin) and choosing which rules are suitable for the ui-kit

## 💰 Use value
The rules will help maintain the new style guide recently applied to the ui-kit.

## Additional info
Some larger components are unable to strictly follow the style guide (variables set by props etc) so this ticket may include finding these exceptions and adding exclusion lines.

PR #878, opened 2023-07-11
814 stencil eslint plugins

GCHQ-Developer-741:
## Summary of the changes
Added stencil-community eslint plugin, and fixed outstanding linting issues

## Related issue
#814 

## Checklist


.eslintrc.js
```diff
@@ -22,7 +22,33 @@ module.exports = {
...
+            "listen": "multiline"
+            }
+        ],
+        "@stencil-community/element-type": "error",
+        "@stencil-community/methods-must-be-public": "error",
+        "@stencil-community/no-unused-watch": "error",
+        "@stencil-community/own-methods-must-be-private": "error",
+        "@stencil-community/own-props-must-be-private": "error",
+        "@stencil-community/prefer-vdom-listener": "error",
+        "@stencil-community/props-must-be-public": "error",
```
jd239, line 44: I think we have some internal props like: https://github.com/mi6/ic-ui-kit/blob/main/packages/web-components/src/components/ic-breadcrumb/ic-breadcrumb.tsx#L37

I'm assuming I've misunderstood this lint rule.

package.json
```diff
@@ -54,7 +54,7 @@
     "remark-lint": "^9.1.0",
     "remark-preset-lint-recommended": "^6.1.1",
     "rimraf": "^3.0.2",
-    "stylelint": "^15.10.1",
+    "stylelint": "^14.16.1",
```
jd239, line 57: Is it possible to use v15 or does that break the stencil lint?

jd239 (changes requested):
Nice update to the linting. A few comments

package.json
GCHQ-Developer-741: Anything above v14 introduces issues with peer dependencies. Stenci-community requires <v15 otherwise we end up with a situation like the design-system where we need to add --legacy-peer-deps everytime we do an install.

.eslintrc.js
GCHQ-Developer-741: So that lint rule checks whether props have the `private` or `protected` keyword next to it. Seems to be a strange rule however because I just tried it and I couldn't even build the ui-kit because it raised an error, so I feel as though you'd spot this before running the linter.

jd239 (approved):

Issue #897, opened 2023-07-14
Update default search variant to use 'go' button


DBD324:
## Description
The current site search on the ICDS uses a droplist to show search results. It allows the auto-focus of one of the search results, and then the navigation to that page on 'enter' keypress, or click of the search icon button. 

The use of the search icon button is a bit misleading, as this variant of search doesn't allow the search term to be collected and then used on a separate search results page. 

To combat this, the design for this variant of search should be altered to use a search icon on the left hand side, and a 'go' or '->' icon button on this right hand side which is more accurate to the function it performs. 

When using the other search variant, which allows search terms to be entered and then emitted for a developer to use to feed a search results page, then the existing styling should be used, with the search icon button performing a true search. 

## 💰 Use value
Providing the two sets of styling enhances the understanding of the different modes of operation when using the two different search variants. 

## Additional info
Tell us anything else useful to help us understand your suggestion.

mi6-577:
Is this still an issue? The components have moved on considerably since 2023 as has the site.

Issue #798, opened 2023-06-19
[ic-search-bar]: Allow input custom value to be submitted when search bar uses option list


jd239:
## Summary
At the moment, it is not possible to submit a phrase from ic-search bar unless it is in the options list.

## 💬 Description
When using an options list with search bar, when the user starts typing out a phrase, the first option in the menu is highlighted. When enter is pressed, that value is submitted. A developer within the community would like to:
1. Have the options list appear
2. Submit the phrase within the input rather than have to select an item in the option menu

## 💰 Use value
This allows the search bar to direct users to a search results page if required or directly to a page suggested within the options menu. This will enhance the search bar functionality.

## Additional info
Current implementation: https://design.sis.gov.uk/components/search-bar

PR #899, opened 2023-07-14
fix(web-components): added prop to prevent auto highlighting first op…

GCHQ-Developer-741:
…tion

this prop gives people the ability to submit terms not included in the options list, since it won't automatically populate the field with the first option

## Summary of the changes
In order to allow users to submit their own custom search terms, the automatic highlighting of the first option needed to be removed, so that the field doesn't get populated on submission of search. However, there are a few scenarios where this functionality is needed, for example on our guidance site, which functions more as a quick navigation tool and therefore relies on having the options highlighted to avoid 404 errors.

To add flexibility, I have added a new prop called `searchMode`, which is `navigation` by default but users can set this to `query` if they want the search bar to function more traditionally.

A new ticket has been raised in design (https://github.com/mi6/ic-ui-kit/issues/897) to add some visual changes to the search bar when this prop is enabled, to make clear the difference in behaviour when this prop is used.

## Related issue
#798 

## Checklist


DBD324 (changes requested):
Think we need to review the terminology here. The issue isn't as simple as just disabling auto highlighting. It actually creates a totally different mode of search interaction. Think the work is sound, we just need to be really careful about the prop naming and the prop description as it changes the search mode drastically. 

Perhaps we could use something like:
- goToSearchResult: false

or conversely
-searchForInputTerm: true. 

I would like the team to think about wording this so that its impact and function is understandable.

jd239 (dismissed):
Nice work. I will approve as the code update is good. :)

packages/web-components/src/components/ic-search-bar/ic-search-bar.tsx
GCHQ-Developer-847, line 1: Tried adding the prop set to "query" on the "With Options" story - if I type something in and then start deleting the text I've typed in, the first option gets highlighted

GCHQ-Developer-847 (changes requested):

GCHQ-Developer-847:
Just noticed another thing actually - on the React "With options" story, I've tried setting searchMode to "query" and then added `onIcSubmitSearch={(ev) => console.log(ev.detail.value)}`. When I type something in the input box and press enter, nothing gets logged to the console. 

So when someone sets the search mode to "query", should it actually allow them to submit it i.e. should it emit onIcSubmitSearch with the value they've typed in when they press enter? That means they can either submit the value of one of their options, or the custom value they've typed in (a bit like how it works in the "Default" story - whatever you've typed in gets submitted)

GCHQ-Developer-847:
> Just noticed another thing actually - on the React "With options" story, I've tried setting searchMode to "query" and then added `onIcSubmitSearch={(ev) => console.log(ev.detail.value)}`. When I type something in the input box and press enter, nothing gets logged to the console.
> 
> So when someone sets the search mode to "query", should it actually allow them to submit it i.e. should it emit onIcSubmitSearch with the value they've typed in when they press enter? That means they can either submit the value of one of their options, or the custom value they've typed in (a bit like how it works in the "Default" story - whatever you've typed in gets submitted)

^ Actually this does work - it does get submitted, but only when there options. E.g. on the "With options" story, I can submit "ca" or "am" for example, as that brings up options that match. But when I type in a random string which doesn't match any option (when it says "No results found"), I can't submit it

GCHQ-Developer-847 (approved):
Approved (we have spoken about my previous comment and I am happy that it is all working as it should - was possibly fixed by another change)

DBD324 (approved):
Terminology seems much more understandable now, thank you.

PR #909, opened 2023-07-17
fix(web-components): #653 adds z-index for sticky page header

j212122:
adds new z-index var for sticky page header so select box does not overlap

## Related issue
[#653 ](https://github.com/orgs/mi6/projects/2/views/1?pane=issue&itemId=27454988)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,30 +3,51 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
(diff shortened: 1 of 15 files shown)


GCHQ-Developer-530 (approved):

ad9242 (changes requested):
I think also need to add the new variable in for the .sticky-desktop class on line 131. It didn't solve the issue currently testing on guidance site

GCHQ-Developer-847 (approved):

MI6-255 (approved):

Issue #896, opened 2023-07-14
Collapsed with labels side navigation variant stops showing navigation items after it is expanded


GCHQ-Developer-847:
## Summary of the bug
When you open a side navigation component which has labels visible when it is collapsed (i.e. `collapsed-icon-labels` set to true), the navigation items disappear when you expand it.

## 🪜 How to reproduce
Tell us the steps to reproduce the problem:
1. Go to the 'Collapsed labels' example in the 'Code' tab on the [side navigation page](https://design.sis.gov.uk/components/side-nav).
2. Expand the side navigation.
3. The navigation items disappear, and they are no longer visible when you collapse it again.

## 🧐 Expected behaviour
The navigation items should not disappear.

## 📸 Screenshots or code
<img src="https://github.com/mi6/ic-ui-kit/assets/111882109/6137f201-3fad-4f7f-a03a-3f9f097add7c" height="400">
<img src="https://github.com/mi6/ic-ui-kit/assets/111882109/9bec813f-48c7-48f9-8b15-a838ca20b45b" height="400">

## 🖥 Desktop
 - Browser: Chrome, Safari, Firefox

Issue #885, opened 2023-07-12
[ic-side-navigation]: static variant applies padding-left with pushes the content too far


jd239:
## Summary of the bug
The `static` variant removes the expand button on the side navigation. If the side navigation is not expanded and static is set, the side navigation stays in it's collapsed behaviour. However, a `padding-left: calc(var(--ic-space-xl) * 10)` is applied to the wrapper which forces the content to the right. 

## 🪜 How to reproduce
Tell us the steps to reproduce the problem:
1. Clone repo, run setup and boot up storybook
2. Go to side navigation and update the static example by removing the `expanded` prop
3. Set the viewport/browser window to approx 845px.
4. See the error

## 🧐 Expected behaviour
The padding on the right should set the content next to the side navigation even when it is closed.

## 📸 Screenshots or code
![Capture](https://github.com/mi6/ic-ui-kit/assets/116016939/74dabd00-3167-4928-b65d-78accb4e9109)

## 🖥 Desktop
 - OS:  Windows
 - Browser: Chrome
 - Version: 114

## 📱 Mobile
N/A

## Additional info
Tell us anything else useful to help us fix or understand the problem.

Issue #739, opened 2023-06-06
Side nav displays on small screen as empty after opening and changing screen width


GCHQ-Developer-847:
## Summary of the bug
When there is a side navigation component on a small screen (so that it appears across the top of the screen), if you open the menu and then increase and decrease the screen width, an empty menu is displayed.

## 🪜 How to reproduce
1. Go to a page with a side nav, e.g. [Storybook](https://mi6.github.io/ic-ui-kit/branches/develop/iframe.html?args=&id=web-components-side-navigation--default-story&viewMode=story)
2. Decrease the screen width so that the side nav moves to the top.
3. Open the side nav menu.
4. Increase the screen width so that it moves back to the side.
5. Decrease the screen width so it moves to the top again.
6. See that the menu is open and is not displaying anything - it is empty.

## 🧐 Expected behaviour
The menu should display the navigation items as normal. (Or should it just be collapsed again? Worth checking with a designer maybe).

## 📸 Screenshots or code
![image](https://github.com/mi6/ic-ui-kit/assets/111882109/91d601c2-121c-4617-97d3-3586b0a96c71)

## 🖥 Desktop
 - OS: MacOS
 - Browser: Chrome, Safari, Firefox

On Safari and Firefox, the nav items appear for a second and then disappear.

## Additional info
Make sure to check that the fix works with slotted links as well.

PR #910, opened 2023-07-17
Side navigation bug fixes

GCHQ-Developer-847:
## Summary of the changes
For #739: Updated CSS so that the visibility is handled by just the 'menu-visibility-visible' class instead of in the media query, and remove unused classes in ic-navigation-group CSS.

For #885: Fixed padding so content is pushed to the side the correct amount with static variant.

For #896: Updated transition handlers - re-added functionality for adding / removing classes on transition start.

## Related issue
#739, #885 and #896.

## Checklist

### DIFF
--- a/packages/web-components/src/components/ic-side-navigation/ic-side-navigation.tsx
+++ b/packages/web-components/src/components/ic-side-navigation/ic-side-navigation.tsx
@@ -336,7 +336,7 @@ export class SideNavigation {
     }
   };
 
-  private transitionEndHandler = () => {
+  private transitionHandler = (type: string) => {
     const primaryNavigationWrapper = this.el.shadowRoot.querySelector(
       ".primary-navigation"
     );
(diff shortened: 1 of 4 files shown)


GCHQ-Developer-530 (dismissed):

GCHQ-Developer-741 (approved):

packages/web-components/src/components/ic-side-navigation/ic-side-navigation.tsx
```diff
@@ -345,23 +345,34 @@ export class SideNavigation {
...
   };
 
+  private transitionEndHandler = () => {
+    this.transitionHandler("end");
+  };
+
   private animateCollapsedIconLabels = () => {
-    this.transitionEndHandler();
+    this.transitionHandler("start");
+    this.transitionHandler("end");
```
jd239, line 375: Can use your new `transitionEndHandler` method here..or if it's only being used once, remove it.

jd239 (changes requested):
Only one comment. Good work on the side navigation! :)

Issue #866, opened 2023-07-07
Introduce a PR checklist


MI6-255:
## Summary
Introduce a PR checklist, to allow those suggesting code to check it meets all our criteria and not merge code that is not "done".

## 💬 Description
The following should be introduced:
- A checklist discussed with the team on every PR in both ic-ui-kit and ic-design-system (this may be the definition of done)
- A process to prevent merge if an item is not checked (making optional points clear)

Suggested PR checklist:

- [ ] Relevant unit tests and visual regression tests added. 
- [ ] Visual testing against Figma component specification completed. 
- [ ] All acceptance criteria reviewed and met. 
- [ ] Accessibility Insights FastPass performed.
- [ ] A11y unit test added and yields no issues.
- [ ] A11y plug-in on Storybook yields no issues. 
- [ ] Manual screen reader testing performed using NVDA and VoiceOver. 
- [ ] Page can be zoomed to 400% with no loss of content. 
- [ ] Screen magnifier used with no issues. 
- [ ] Text resized to 200% with no loss of content.
- [ ] Text spacing increased as per the [WCAG 1.4.12 success criterion](https://www.w3.org/TR/WCAG21/#text-spacing) with no loss of content.
- [ ] Browser setting 'prefers reduced motion' tested. No animations or motion visible whilst this setting is on. 
- [ ] Windows High Contrast mode tested with no loss of content.
- [ ] System light and dark mode tested with no loss of content. 
- [ ] Manual keyboard testing for keyboard controls and logical focus order. 
- [ ] Min/max content examples tested with no loss of content or overflow. 
- [ ] Browser support tested (Chrome, Safari, Firefox and Edge).
- [ ] Correct roles used and ARIA attributes used correctly where required. 
- [ ] Logical heading structure is maintained, and the HTML elements used for headings can be changed to fit within the wider page structure. 
- [ ] All prop combinations work without issue. 

## 💰 Use value
Improve product quality and reduce bugs.

GCHQ-Developer-847:
PR created for action to prevent merge if there are uncompleted tasks: https://github.com/mi6/ic-ui-kit/pull/930

GCHQ-Developer-847:
PR created to add checklist to PR template: #1087

PR #930, opened 2023-07-20
feat(root): add action to check for tasks in PR description not compl…

GCHQ-Developer-847:
## Summary of the changes
Added 'check-tasks-completed' action to check there are no tasks in PR description which are not completed.

I have tested this by adding an uncompleted task to a PR description. The action fails, but then when you click on the checkbox on the task to mark it as done, it runs again and passes (and vice versa).

## Related issue
#866

### DIFF
--- a/.github/workflows/check-tasks-completed.yml
+++ b/.github/workflows/check-tasks-completed.yml
@@ -1,0 +1,12 @@
+name: Prevent PRs with tasks not completed
+
+on:
+  pull_request:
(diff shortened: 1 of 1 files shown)


GCHQ-Developer-530 (approved):

MI6-255:
Approved. Presume we won't close this issue until we've added the DoD?

MI6-255 (approved):

Issue #867, opened 2023-07-07
Updates to Date input component for use with Date Picker


ad9242:
## Summary
Some enhancements to date input to allow for it's use with Date Picker

- Add a clear button inside the input
- Add a launch picker button inside the input
- Add size variants - default, small & large
- Add max\min support for when typing a date

## Acceptance Criteria
Given that there is a Date input component
When using the component,
Then the user can choose between the default, small and large variants.

Given that a Date input component is being used 
When the input field is focused,
Then a button is visible in the input field to allow clearing the value

Given that a Date input component is being used as a date picker
When the input field is focused,
Then a button is visible in the input field to allow opening of a date picker

Given that a Date input has the `max` prop set
When a user types in the field
Then it should not be possible to enter a date which is greater than the max

Given that a Date input has the `min` prop set
When a user types in the field
Then it should not be possible to enter a date which is less than the min

jd239:
Update to disabledDates prop. Split into:
- disableFromNow / disableFuture
- disableUntilNow / disablePast

PR #945, opened 2023-07-26
add clear/calendar button, new sizes, min/max and disabled days to date input

GCHQ-Developer-530:
## Summary of the changes
Add clear button, new sizes (including update to input component container), internal calendar open button for date picker, disabled days, and min/max to date input. Change disabledDates prop to split out into disableFromNow and disableUntilNow.

## Related issue
#867 

## Checklist


packages/react/src/stories/ic-date-input.stories.mdx
```diff
@@ -7,111 +7,134 @@ import {
...
-        helperText='We will have your order ready for you on this day'
-      />
+    <Story name='Sizes'>
+        <IcDateInput
+          label='When would you like to collect your coffee?'
+          size="small"
+        />
+        <br/>
+        <IcDateInput
+          label='When would you like to collect your coffee?'
```
GCHQ-Developer-465, line 96: Should there be a size="medium" or for it to be called size="default" I enquire this due to consistency with other components such as [Loading indicator](https://design.sis.gov.uk/components/loading-indicators).

packages/react/src/stories/ic-date-input.stories.mdx
```diff
@@ -7,111 +7,134 @@ import {
...
 </Canvas>
 
+
+### Custom helper text
+
+<Canvas>
+  <Story name="Custom helper text">
+    <IcDateInput
+      label="When would you like to collect your coffee?"
+      helperText="We will have your order ready for you on this day"
```
GCHQ-Developer-465, line 113: Should it be date instead of day here? 
```suggestion
      helperText="We will have your order ready for you on this date"
```

packages/web-components/src/components.d.ts
```diff
@@ -424,21 +420,33 @@ export namespace Components {
...
          */
         "dateFromNowMessage"?: string;
         /**
-          * The text to display as the validation message when `disabledDates` is `until-now` and a disabled date is entered.
+          * The text to display as the validation message when `disableUntilNow` is true and a disabled date is entered.
          */
         "dateUntilNowMessage"?: string;
         /**
-          * If `true`, the disabled state will be set.
+          * The days of the week to disable.
```
GCHQ-Developer-465, line 431:
```suggestion
          * If `true`, the days of the week will be set to disable.
```

GCHQ-Developer-465 (commented):
Great work! Just a few comments :)

packages/react/src/stories/ic-date-input.stories.mdx
GCHQ-Developer-530: I think this is right as it aligns with the input component container

packages/web-components/src/components/ic-date-input/ic-date-input.stories.mdx
```diff
@@ -269,12 +328,42 @@ import readme from "./readme.md";
...
+</Canvas>
+
+### Min and max
+
+<Canvas>
+  <Story name="Min and max" parameters={{ loki: { skip: true } }}>
+    {html`<ic-date-input
+        label="When would you like to collect your coffee?"
+        min="2023-01-02"
+        max="2023-07-20"
```
jd239, line 358: Should max/min also accept the same date formats as `.value` to keep things consistent?

packages/web-components/src/components.d.ts
GCHQ-Developer-530: This isn't a boolean value so needs to be worded like this. This prop takes an array of days that the user would like to be disabled, so for example you could disable every monday.

packages/web-components/src/components/ic-date-input/ic-date-input.stories.mdx
```diff
@@ -269,12 +328,42 @@ import readme from "./readme.md";
...
+### Disabled days
+
+<Canvas>
+  <Story name="Disabled days" parameters={{ loki: { skip: true } }}>
+    {html`<ic-date-input
+        label="When would you like to collect your coffee?"
+      ></ic-date-input>
+      <script>
+        var dateInput = document.querySelector("ic-date-input");
+        dateInput.disableDays = [0, 1];
```
jd239, line 346: Assuming these are coming from the date picker, would it make more sense if the `disableDays` were an array of actual days rather than indexes and we convert them to indexes internally?
Or....could accept both?

packages/react/src/stories/ic-date-input.stories.mdx
GCHQ-Developer-465: Maybe thoughts should be drawn out for size consistency between components.

packages/react/src/stories/ic-date-input.stories.mdx
GCHQ-Developer-465: Awesome! Thanks 😊

packages/web-components/src/components/ic-date-input/ic-date-input.tsx
```diff
@@ -695,11 +848,28 @@ export class DateInput {
...
+        } else {
+          this.invalidDateText = `Please enter a date after ${this.min}.`;
+        }
+      } else if (this.isAfterMax) {
+        if (this.disableFromNow) {
+          this.invalidDateText = this.dateFromNowMessage;
+        } else {
+          this.invalidDateText = `Please enter a date before ${this.max}.`;
+        }
+      } else if (this.disableDays.length != 0) {
```
jd239, line 863:
```suggestion
      } else if (this.disableDays.length !== 0) {
```

packages/web-components/src/components/ic-date-input/ic-date-input.tsx
```diff
@@ -1189,8 +1400,37 @@ export class DateInput {
...
+      this.setPreventInput(input, false);
+    });
+    this.isDateSetFromKeyboardEvent = false;
+    this.value = null;
+    this.setValidationMessage();
+    this.handleDateChange(true);
+
+    setTimeout(() => {
+      this.dayInputEl.focus();
+    }, 100);
```
jd239, line 1414: Is `setTimout` still required? Also, what if the date format is YYYY/MM/DD or MM/DD/YYYY? The focus will go to the incorrect input. Can use `this.inputsInOrder()[0]` or something like that

packages/web-components/src/components/ic-date-input/ic-date-input.tsx
```diff
@@ -1234,9 +1474,50 @@ export class DateInput {
...
+                    ["hidden"]:
+                      isEmptyString(this.day) &&
+                      isEmptyString(this.month) &&
+                      isEmptyString(this.year),
+                  }}
+                  innerHTML={Clear}
+                  onClick={this.handleClear}
+                  variant="icon"
+                  appearance={IcThemeForegroundEnum.Dark}
+                  size={size === "large" ? "default" : "small"}
```
jd239, line 1496: Put `size === 'large'` into a variable as it's used more than once

jd239 (changes requested):
👋 A few comments

packages/web-components/src/components/ic-date-input/ic-date-input.css
```diff
@@ -50,3 +50,63 @@ ic-input-component-container:hover {
...
+  right: 2rem;
+  margin-right: var(--ic-space-xs);
+}
+
+.clear-button.hidden {
+  display: none;
+}
+
+.clear-button:focus,
+.calendar-button:focus {
```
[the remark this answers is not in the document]
GCHQ-Developer-530: I think they need to stay separate to get the css from ic-button that's used on every clear-button

packages/web-components/src/components/ic-date-input/ic-date-input.css
```diff
@@ -50,3 +50,63 @@ ic-input-component-container:hover {
...
+  fill: white;
+}
+
+ic-input-component-container.small {
+  --input-width: 16.5rem;
+}
+
+ic-input-component-container.large {
+  --input-width: 22rem;
+}
```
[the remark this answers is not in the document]
GCHQ-Developer-530: It looks to me like it updates the width of the date input as the padding value on the side remains unchanged and gap is always 8px e.g 
Screenshot 2023-08-01 at 07 45 45
Screenshot 2023-08-01 at 07 45 51
Not sure if I'm looking in the wrong place

packages/web-components/src/components/ic-date-input/ic-date-input.stories.mdx
GCHQ-Developer-530: I thought it did as it accepts Date object or string

packages/web-components/src/components/ic-date-input/ic-date-input.css
```diff
@@ -50,3 +50,63 @@ ic-input-component-container:hover {
...
+.input-container {
+  width: 100%;
+  display: flex;
+  align-items: center;
+  position: relative;
+}
+
+.clear-button,
+.calendar-button {
+  position: absolute;
```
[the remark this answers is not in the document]
GCHQ-Developer-530: I tried wrapping in flex but it doesn't ever apply flex-end so the calendar and clear just remain at the side of the inputs, until I add position:absolute

packages/web-components/src/components/ic-date-input/ic-date-input.stories.mdx
GCHQ-Developer-530: Yeah I'm not sure, I presume there was a reason @ad9242 implemented it like this...

packages/web-components/src/components/ic-date-input/ic-date-input.css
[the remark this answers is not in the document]
jd239: Was thinking more in the lines of 
```jsx
            <div class="input-container"> /* display: flex; justify-content: space-between; */
              <div class="date-inputs">
                { /* Date inputs */ }
              </div>
              <div class="action-buttons">
                {showClearButton && (
                  // Clear button
                )}
                {(this.day || this.month || this.year) &&
                  showClearButton &&
                  showCalendarButton && (
                   // Divider
                  )}
                {showCalendarButton && (
                  // Calendar button
                )}
              </div>
            </div>

```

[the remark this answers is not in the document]
jd239: To be honest, I'm not sure if it improves the code anyway. Just always think absolute should be used if last resort.

packages/web-components/src/components/ic-date-input/ic-date-input.css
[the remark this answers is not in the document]
jd239: Sorry, think I've made this confusing. Was trying to work out if the `size` prop should only affect the padding of the input-container rather than increase it's width in addition to the padding. Since we are using `size` as our 'density', density is more to do with how close the input values/textContent is to the border surrounding it rather than the width.

packages/web-components/src/components/ic-date-input/ic-date-input.stories.mdx
jd239: If I pass through a max date as '20-07-2023', 'Invalid Date' is returned from `this.maxDate' within 'watchMaxHandler'.

packages/web-components/src/components/ic-date-input/ic-date-input.stories.mdx
jd239: Let's park this for now. Assuming this is being passed down from `Date Picker` so maybe this question is more related to that component.

packages/web-components/src/components/ic-date-input/ic-date-input.tsx
```diff
@@ -123,6 +151,36 @@ export class DateInput {
...
+   * The value of this prop is ignored if `disableFromNow` is set to `true`.
+   */
+  @Prop() max?: string | Date = "";
+
+  @Watch("max")
+  watchMaxHandler(): void {
+    if (this.disableFromNow) {
+      this.maxDate = new Date();
+    } else {
+      this.maxDate = new Date(this.max);
```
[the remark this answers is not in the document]
jd239: This is related to https://github.com/mi6/ic-ui-kit/pull/945#discussion_r1280414472

packages/web-components/src/components/ic-date-input/ic-date-input.css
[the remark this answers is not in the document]
GCHQ-Developer-530: Ahh yeah I think this is a weird one because it's meant to match the width of the calendar for date picker which obviously does change width for the different size variants, but then if this was just used on its own would we necessarily want it to change width? Hopefully a designer can weigh in on this

packages/web-components/src/components/ic-date-input/ic-date-input.tsx
```diff
@@ -104,9 +118,25 @@ export class DateInput {
...
+   * The days of the week to disable.
    */
-  @Prop() disabledDates?: IcDisabledDateTypes;
+  @Prop() disableDays?: IcWeekDays[] = [];
+
+  /**
+   * The text to display as the validation message when `disableDays` is set and a disabled date is entered.
+   */
+  @Prop() disableDaysMessage?: string =
+    "The date you have selected is on a weekday that is not allowed. Please select another date.";
```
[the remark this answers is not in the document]
GCHQ-Developer-530: I've changed the default for the prop to day of the week. I think IcWeekDays is fine as it's not all one word which I think implies we mean days of the week and not weekday.

GCHQ-Developer-847 (approved):

packages/web-components/src/components/ic-date-input/ic-date-input.css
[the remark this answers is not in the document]
GCHQ-Developer-530: Designers agree that it should only affect the padding and the width should only change for date picker to match calendar sizes. This has now been updated.

jd239 (approved):
Nice work!

GCHQ-Developer-465 (approved):

PR #961, opened 2023-07-31
Z index not correct for sticky ic page

j212122:
## Summary of the changes
adds new z-index value for sticky page header to prevent select box overlapping

## Related issue
#653


jd239 (changes requested):
Looks like the two commits can be merged into one. Also added another comment.

jd239 (approved):

GCHQ-Developer-465 (approved):

Issue #831, opened 2023-06-27
Incorrect focus border radius


GCHQ-Developer-530:
## Summary of the bug
Whilst developing the badge component it has been noticed that the border radius that comes from --ic-border-focus is incorrect and the focus glow curves at the wrong point.

## 🧐 Expected behaviour
New designs: White -> border radius of 4px, Blue -> border radius of 6px, Glow -> border radius of 10px. 

## 📸 Screenshots or code
Figma designs where the badge fits perfectly into the curved focus flow:
<img width="558" alt="Screenshot 2023-06-27 at 14 33 55" src="https://github.com/mi6/ic-ui-kit/assets/111882034/9480cc51-e2a1-40d0-8e01-3977de49755c">
How it currently looks:
<img width="404" alt="Screenshot 2023-06-27 at 14 33 42" src="https://github.com/mi6/ic-ui-kit/assets/111882034/0d3d3fd8-1dd9-4853-8a25-5dac63096789">

PR #964, opened 2023-08-02
style(web-components): changed sizing of --ic-border-focus

GCHQ-Developer-741:
## Summary of the changes
Updated `--ic-border-focus` to use new values

## Related issue
#831 

## Checklist


GCHQ-Developer-530 (changes requested):
Just tried with badge and it seems to be further off aligning due to the white part of the focus now being double the original size. Might be worth going back to @DBD324 about the values they suggested. Not sure if there's a way to fix this but the issue with the original border focus is just that the outer glow isn't as circular as the badge is so I don't know if there's a way to adjust how rounded the border is?
Screenshot 2023-08-02 at 14 33 50

GCHQ-Developer-530:
> Just tried with badge and it seems to be further off aligning due to the white part of the focus now being double the original size. Might be worth going back to @DBD324 about the values they suggested. Not sure if there's a way to fix this but the issue with the original border focus is just that the outer glow isn't as circular as the badge is so I don't know if there's a way to adjust how rounded the border is? Screenshot 2023-08-02 at 14 33 50

I was just playing around with this on the a11y/badge branch and weirdly keeping the original border focus values, but updating --ic-border-radius to 3px fixed it and all of the badges fit perfectly. But it seems a bit of a strange fix and it would update a lot of components, so probably not the best idea.

GCHQ-Developer-465 (dismissed):

packages/web-components/src/global/variables.css
```diff
@@ -2,7 +2,7 @@
   --ic-border-radius: 2px;
   --ic-border-radius-inset: 11px;
   --ic-border-focus: 0 0 0 2px var(--ic-architectural-white),
-    0 0 0 4px var(--ic-focus-blue), 0 0 0 8px var(--ic-focus-glow);
+    0 0 0 2px var(--ic-focus-blue), 0 0 0 8px var(--ic-focus-glow);
```
GCHQ-Developer-530, line 5:
```suggestion
    0 0 0 4px var(--ic-focus-blue), 0 0 0 8px var(--ic-focus-glow);
```

GCHQ-Developer-530 (approved):

GCHQ-Developer-299:
Would be nice to have an entry in Storybook for Button component with Badge - is that something to stick in another ticket though?

GCHQ-Developer-741:
> Would be nice to have an entry in Storybook for Button component with Badge - is that something to stick in another ticket though?

This exists on the [a11y/badge](https://github.com/mi6/ic-ui-kit/tree/a11y/badge) branch. Since this hasn't been merged into develop yet, it won't show, but once they both have it'll be there

GCHQ-Developer-465 (approved):

Issue #655, opened 2023-01-10
Empty state component


MI6-451:
## Summary
Research and design a pattern and guidance for the 'empty state'

The empty state pattern would describe a general pattern for any component, pattern or page that may have an empty state. It can be used when something fails to load, when there are no results matching search terms or filter options, or if other error situations happen.

## 💬 Description
We have use cases which require an 'empty state pattern' where a graphic or similar is used as a placholder for later information/data/components which are not yet present e.g. a landing/homepage where the user has no alerts yet.

The current pattern is simply a horizontally and vertically centered graphic or similar image with a short message in some cases e.g. "No messages to display"

We should research whether this approach is valid and formalise a pattern for implementing it.

## 💰 Use value
A pattern for a common use case will make developer/designer experience easier and ensure consistency across developed applications.

Empty states are important as they provide the user with feedback when there is no content to display. It provides a method of being able to try again and avoids blank or empty components and elements.

## Additional info
Considerations:
The graphic can be an icon, a vector graphic, or an image. 
We should recommend that the graphic itself is aria-hidden because it's decorative.
We may wish to offer a utility component to do this, consider during this ticket.
We should specify size(s) for the graphic element, including Typography of the text underneath.

## 📚 User Stories
If relevant, describe the high-level functionality as user stories.

*As an* ICDS user: 
*I need* an empty state pattern
*So that* I can apply it to components or patterns that may have no content. 

## 📝 Acceptance Criteria 
If relevant, describe in full detail the different interactions and edge cases that the component or patterns needs to fulfil.

*Given* that an empty state is being implemented
*When* setting the properties
*Then* the following elements can be toggled on/off:
- Image - toggleable - default: on. 
- Title - permanent - required. 
- Subtitle - toggleable - default: on. 
- Body - toggleable - default: on. 
- Interaction slot: toggleable - default: on. Action button - toggleable - default: on. OR - Link - toggleable - default: on. 

*Given* that an empty state with action button is being implmented
*When* setting its properties
*Then* the action button can be switched to either primary, secondary, tertiary, destructive, or icon button. 

*Given* that an empty state is being implemented, 
*When* setting the interaction area up
*Then* its contents should be able to be slotted to allow custom interactions to be included. 
(for example, instead of the default option of showing a single button, it may be customised to include 2 buttons.)

*Given* that an empty state has been defined with an image
*When* setting its properties
*Then* the image size can be selected:
- Default: 72px x 72px
- Small: 48px x 48px
[...]

PR #982, opened 2023-08-08
655 empty state component

GCHQ-Developer-530:
## Summary of the changes
Create empty state component, add relevant tests and stories.
 
## Related issue
#655 

## Checklist


DBD324 (changes requested):
Looks great, the only two things are:
- The padding between the link and button is quite tight. On the Figma designs, the link uses a larger line height in addition to the 8px padding. Could you take a look at this and see if you can match please? Perhaps we need to update 8px padding to somethink like 16px? Or is this something that is controlled within the slot?
- Are we able to change the image to an icon? I assume so given that it is a slot, but maybe good to show an example of this in storybook? 
- Finally, is it possible to show an example of an empty state component embedded in another div with a border and background? This would just show how it can be constrained to fill the contents of its parent.

Thanks

GCHQ-Developer-847 (changes requested):
Looks great! Only one very small comment :)

GCHQ-Developer-847 (approved):

GCHQ-Developer-299 (approved):

PR #987, opened 2023-08-09
fix(web-components): conditionally adding aria-owns to search-bar

GCHQ-Developer-741:
only adding aria-owns if the menu is being rendered at the same time

## Checklist


GCHQ-Developer-530 (approved):

ad9242:
Just one commment: should we add a test so that we can see that its starts without the aria-owns & then gets added?

GCHQ-Developer-530 (approved):

ad9242 (approved):

PR #1019, opened 2023-08-18
fix(web-components): ic-tooltip no longer interferes with previousSibling

GCHQ-Developer-299:
Ic Tooltip would, when the previous element is conditionally rendered in react lifecycle, replace the contents of it. Now we store ariaDescribedBy in a variable so we don't search for that previousElementSibling

## Summary of the changes
Changed how updateLabel() functions, instead of getting the previous element sibling on every call, it refers to an ariaDescribedBy variable created on load. 

However, i've been unable to add a test - because the issue occurs during react apps with conditional rendering. We only have web-component unit tests. I will enquire with the developer team :) 

## Related issue
#986 

## Checklist

### DIFF
--- a/packages/web-components/src/components/ic-tooltip/ic-tooltip.tsx
+++ b/packages/web-components/src/components/ic-tooltip/ic-tooltip.tsx
@@ -18,6 +18,7 @@ import { onComponentRequiredPropUndefined } from "../../utils/helpers";
 })
 export class Tooltip {
   private arrow: HTMLDivElement;
+  private ariaDescribedBy: HTMLElement;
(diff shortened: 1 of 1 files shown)


GCHQ-Developer-530:
Would it be possible to add that code snippet in your comment as a react story?

GCHQ-Developer-299:
> Would it be possible to add that code snippet in your comment as a react story?

I spoke to the tech lead about that, and we decided that since it was demonstrating a bugfix, rather than a feature of the component, it shouldn't remain as a story.

ad9242 (approved):

Issue #1006, opened 2023-08-15
[ic-dialog]: Review and update components which use popperJS within dialog


jd239:
**The details of this ticket demonstrate the issue with `ic-dialog` and `ic-tooltip`, however, this work should involve investigating all components which use the popperJs library**

## Summary of the bug
- I have a dialog that contains IcChips
- The chips have tooltips
- When I hover over the chips to make the tooltip visible
- The tooltips appear nowhere near the chips, and cause me to scroll in both directions to make the tooltips visible
Also, on a more restricted width, the tooltips do not cause the dialog to scroll, but appear outside the tooltips

## 🪜 How to reproduce
Tell us the steps to reproduce the problem:
1) Open the dialog
2) Hover over either/both of the chips
3) Scroll to find the tooltips
4) Reduce the width of the screen (so the dialog is still visible)
5) Repeat step 2

## 🧐 Expected behaviour
Tooltip should be on show without the scrollbars

## 📸 Screenshots or code
Example of bug: https://stackblitz.com/edit/stackblitz-starters-fvvwad?file=src%2FApp.tsx

## 🖥 Desktop
N/A

## 📱 Mobile
N/A

## Additional info
N/A

PR #1020, opened 2023-08-22
fix(web-components): update tooltip and popover menu to adjust placement on dialogs

GCHQ-Developer-530:
## Summary of the changes
Update tooltip and popover menu to adjust placement on dialogs. Update tooltip to only create popper on show

## Related issue
#1006

## Checklist

### DIFF
--- a/packages/web-components/src/components/ic-popover-menu/ic-popover-menu.tsx
+++ b/packages/web-components/src/components/ic-popover-menu/ic-popover-menu.tsx
@@ -10,7 +10,7 @@ import {
   Method,
 } from "@stencil/core";
 import { getSlotElements } from "../../utils/helpers";
-import { createPopper } from "@popperjs/core";
+import { createPopper, Instance as PopperInstance } from "@popperjs/core";
 
 @Component({
   tag: "ic-popover-menu",
@@ -24,7 +24,9 @@ export class PopoverMenu {
   private ARIA_LABEL: string = "aria-label";
   private backButton: HTMLIcMenuItemElement;
   private currentFocus: number;
+  private firstRender: boolean = true;
   private popoverMenuEls: HTMLIcMenuItemElement[] = [];
+  private popperInstance: PopperInstance;
 
   @Element() host: HTMLIcPopoverMenuElement;
 
@@ -77,6 +79,12 @@ export class PopoverMenu {
     }
   }
 
+  disconnectedCallback(): void {
+    if (this.popperInstance !== undefined) {
+      this.popperInstance.destroy();
+    }
+  }
+
   componentDidLoad(): void {
     const slotWrapper = this.host.shadowRoot.querySelector("ul.button");
     const popoverMenuElements = getSlotElements(slotWrapper);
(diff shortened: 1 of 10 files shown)


packages/web-components/src/components/ic-popover-menu/ic-popover-menu.tsx
```diff
@@ -100,18 +108,68 @@ export class PopoverMenu {
...
+      const dialogEl = this.host.closest("ic-dialog");
+      const onDialog = dialogEl !== null;
+
+      if (onDialog) {
+        this.host.classList.add("on-dialog");
+        const menu = this.host.getBoundingClientRect();
+        const dialogBottom = dialogEl.getBoundingClientRect().bottom;
+        const anchorHeight = this.anchorEl.getBoundingClientRect().height;
+        let offset;
+        console.log(dialogEl.scrollHeight, dialogEl.clientHeight);
```
GCHQ-Developer-299, line 124: Leftover console.log?

GCHQ-Developer-299 (approved):
Have tested in storybook and seems to fix the issue comprehensively :)

packages/web-components/src/components/ic-tooltip/ic-tooltip.tsx
```diff
@@ -123,9 +106,156 @@ export class Tooltip {
...
+          `${child.left - dialogEl.left + child.width}px`
+        );
+        this.toolTip.style.setProperty(
+          this.TOOLTIP_Y_VARIABLE,
+          `${child.top - dialogEl.bottom + child.height}px`
+        );
+        break;
+    }
+  };
+
```
ad9242, line 216: rather than lots of calls to this.toolTip.style.setProperty(....), could we just create local variables that get set inside the case: statements, then just have 2 calls to setProperty to set the x & y

ad9242:
There is an issue, when the dialog content scrolls: 
image

the tooltip text is hidden to the left and also appears in wrong position if you have scrolled the content

think we should not allow the x position to be less than 0 when dialog content overflows & may need to do additional calculation to remove the scroll height from the y pos

packages/web-components/src/components/ic-tooltip/ic-tooltip.tsx
```diff
@@ -123,9 +105,111 @@ export class Tooltip {
...
+    this.toolTip.style.setProperty("--tooltip-translate-x", `${tooltipX}px`);
+    this.toolTip.style.setProperty("--tooltip-translate-y", `${tooltipY}px`);
+  };
+
   private show = () => {
     this.toolTip.setAttribute("data-show", "");
-    this.popperInstance.update();
+
+    this.icDialogEl = this.el.closest("ic-dialog");
```
ad9242, line 177: think there is scope for tidying up this code, so it is not run every time the tooltip is shown. Could determine if it is on a dialog in the `componentWillLoad `or `componentDidLoad`.

the `on-dialog` class could then be applied conditionally in the render

packages/web-components/src/components/ic-popover-menu/ic-popover-menu.tsx
```diff
@@ -100,18 +108,67 @@ export class PopoverMenu {
...
-              offset: [0, 4],
+    if (this.firstRender && this.open) {
+      this.firstRender = false;
+      let adjust = false;
+
+      const dialogEl = this.host.closest("ic-dialog");
+      const onDialog = dialogEl !== null;
+
+      if (onDialog) {
+        this.host.classList.add("on-dialog");
```
ad9242, line 119: could set a flag to show it on a dialog & then conditionally apply class in the `render`

GCHQ-Developer-530: Conditionally rendering the class in the render causes issues and never applies the on-dialog class. We do classList.add in other places so I think this should be okay to stay like this in here and ic-tooltip.

GCHQ-Developer-299:
Screenshot 2023-08-29 at 08 16 52

just noticed that the select menu also renders 'inside' rather than on-top of a dialog (checking out the 'Scrollable' dialog in storybook)  - is this expected behaviour, and if not, would we consider it outside the scope of this ticket?

GCHQ-Developer-530:
> https://mi6.github.io/ic-ui-kit/branches/1006-ic-dialog-review-and-update-components-which-use-popperjs-within-dialog/web-components

I think this may be intentional behaviour for select as it only opens inside when you can scroll to view it. @ad9242 might know better

GCHQ-Developer-530:
> Screenshot 2023-08-29 at 08 16 52
> just noticed that the select menu also renders 'inside' rather than on-top of a dialog (checking out the 'Scrollable' dialog in storybook) - is this expected behaviour, and if not, would we consider it outside the scope of this ticket?

I think this may be intentional behaviour for select as it only opens inside when you can scroll to view it. @ad9242 might know better

ad9242:
> > Screenshot 2023-08-29 at 08 16 52
> > just noticed that the select menu also renders 'inside' rather than on-top of a dialog (checking out the 'Scrollable' dialog in storybook) - is this expected behaviour, and if not, would we consider it outside the scope of this ticket?
> 
> I think this may be intentional behaviour for select as it only opens inside when you can scroll to view it. @ad9242 might know better

yes, unfortunately we were unable to find a way to make the content overlay when we looked at select on a dialog and the dialog content is scrollable. This was the compromise: you should be able to scroll to the options

GCHQ-Developer-299 (approved):

ad9242 (approved):

Issue #595, opened 2023-04-20
IcSelect breaks with duplicate options


MI6-651:
## Summary of the bug
The IcSelect component breaks when there are two identical options. It highlights both options at the same time, and prevents the user from using arrow keys to navigate the menu.

MI6-255:
The specific issue here is not that two values with the same key and value are selected, it's more the issue the arrow keys fail to work to select the item in that case

MI6-255:
Potentially explore if we should show two values with duplicate keys, to prevent the end user from struggling to use it

MI6-255:
https://github.com/mi6/ic-ui-kit/issues/442

PR #1051, opened 2023-09-01
fix(web-components): deduplicate ic-select option values

GCHQ-Developer-299:
## Summary of the changes
Check and deduplicate IcMenuOptions and log a warning describing the removed item

## Related issue
#595 

## Checklist

### DIFF
--- a/packages/web-components/src/components/ic-select/ic-select.tsx
+++ b/packages/web-components/src/components/ic-select/ic-select.tsx
@@ -209,6 +209,7 @@ export class Select {
    */
   @Prop() options?: IcMenuOption[] = [];
   @State() filteredOptions: IcMenuOption[] = this.options;
+  @State() uniqueOptions: IcMenuOption[] = this.options;
 
   @Watch("options")
   watchOptionsHandler(): void {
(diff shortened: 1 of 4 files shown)


jd239 (approved):
Very nice!

jd239 (approved):

GCHQ-Developer-530 (approved):

packages/web-components/src/components/ic-select/ic-select.tsx
```diff
@@ -1123,7 +1168,11 @@ export class Select {
               small={small}
               menuId={menuId}
               open={this.open}
-              options={searchable ? this.filteredOptions : options}
+              options={
+                searchable
+                  ? this.filteredOptions
+                  : this.deduplicateOptions(options)
```
ad9242, line 1174: just a comment on the efficiency of this. Currently this function will run every time the component is re-rendered. eg just opening & closing the menu causes it to run. would it be better to just run the code when the options are changed (eg a watch on the options)?

GCHQ-Developer-299: I was unable to mutate the options object in the options watch function without causing a RangeError on the array and having to make this.options into a mutable prop - this might be a skill issue on my part. 
I guess maybe i'd need to create an internal "this.deduplidatedOptions" array that then gets filtered or rendered etc.?

ad9242: yes, that's what I was thinking, you could set\update this in the existing `watchOptionsHandler`

ad9242 (changes requested):
awaiting modification to improve efficiency of dedupe code

ad9242 (approved):

GCHQ-Developer-530 (approved):

Issue #224, opened 2023-01-06
[ic-select] Add ability to pass in custom select items to select component


GCHQ-Developer-530:
## Summary
Allow for users to customise their select items to not just be text, e.g. adding a status tag to some of the options.

## 💰 Use value
Would allow users to pass in their own component into each of the drop down items.

GCHQ-Developer-847:
Two potential solutions for this:

1. Add a property to the IcMenuOption object to which the user can pass custom HTML (a bit like the `actions` property of the data table)

2. Create a new component for the select options (e.g. something like `<IcMenuItem>`) and refactor ic-menu and ic-select to allow users to slot these options into the select (similar to [MUI](https://mui.com/material-ui/react-select/#basic-select)) instead of passing the options as an array to the `options` prop

Option 1 would of course be a lot quicker and easier to implement and the avoid the breaking change of Option 2, but it would allow less flexibility / customisation. Although with Option 1, we could allow for a bit of flexibility about where the custom HTML element is placed by maybe adding another property which determines whether the element is placed to the left or the right of the option text.

DBD324:
Figma updated to include new icon option in select droplist menu options. #1034

PR #1073, opened 2023-09-11
224 ic select add ability to pass in custom select items to select component

GCHQ-Developer-530:
## Summary of the changes
Add icon and element to menu options so that any icon/custom element can be added to select.

## Related issue
#224 

## Checklist


packages/web-components/src/components/ic-menu/test/basic/ic-menu.spec.tsx
```diff
@@ -20,8 +20,19 @@ const menuOptions = [
     value: "tea",
     children: [{ label: "green", value: "green" }],
   },
-  { label: "Aicano", value: "acano" },
-  { label: "Mocha", value: "mocha" },
+  {
```
GCHQ-Developer-299, line 23: Would it be suitable to add a test for custom components in an ic-select? Something to explicitly test that the contents of a component renders, beyond passing a custom option into the regular tests? 
Might be overkill

GCHQ-Developer-530: I think this is covered with the update to the snapshot test. If you look at the mocha/aicano part of the snapshot file, it shows the svg renders for aicano and the status tag renders for mocha

GCHQ-Developer-299: You're right - I think I hadn't considered that since this change doesn't involve logic changes, testing the render alone is sufficient

GCHQ-Developer-299 (approved):

GCHQ-Developer-465 (approved):

Issue #1069, opened 2023-09-07
Slots documentation missing for some components


ad9242:
## Summary of the bug 
The slots documentation is missing from some of the component docs. From an initial quick look, this applies to Dialog & Breadcrumb, but could apply to more, so all components should be checked.

The docs comment is in the component tsx files, but appears to need to be placed higher up. For a good example, see IC-Button code 

## 🪜 How to reproduce 
Tell us the steps to reproduce the problem: 
1. Go to page: https://design.sis.gov.uk/components/dialog/code
2. Scroll down & see "Props" are documented, but there is no info on slots

## 🧐 Expected behaviour 
All components with slots should have their information visible.

## 📝 Acceptance Criteria  

*Given* a component has slots
*When*  viewing the documentation generated for the component
*Then* there should be information about the slots

PR #1074, opened 2023-09-11
fix(web-components): moves slot to top of file so it shows in guidanc…

j212122:
## Summary of the changes

slot information was incorrectly placed in the file so slot guidance was not appearing on the code documentation on the site. Moved slots to top of file, below imports

## Related issue
#1069


ad9242 (commented):
change looks good, but just querying why the package-lock have changed?

PR #1088, opened 2023-09-15
fix(web-components): call functions from componentWillLoad to stop stepper flickering

GCHQ-Developer-530:
## Summary of the changes
call functions from componentWillLoad to stop stepper flickering

## Related issue
#911 

## Checklist


GCHQ-Developer-299 (changes requested):
Looks like the test snapshots need updating
Unfortunate that the stepper visual test is still flaking out

PR #1091, opened 2023-09-15
Bugfix/1059 accordion popover fix

jd239:
## Summary of the changes
Updated content-area on accordion so component with popover are not cutoff by overflow:hidden. Change includes switching to overflow:hidden during the expanding animation so the reveal animation is effective. Once expanded, the content-area is switched to overflow:visible so any popover appears as expected.

## Related issue
N/A

## Checklist


ad9242 (changes requested):
Spotted a bit of an issue using tab key. Focus seems to move to focusable items that are in the collapsed accordian. eg if focus is here:

image

pressing tab shows the tooltip for the button that is not currently visible:
image

Also, tab key does not focus accordian on dialog. I think dialog will need a change to allow this in the list of interactive controls

packages/react/package.json
```diff
@@ -24,8 +24,8 @@
     "dist/"
   ],
   "dependencies": {
-    "@ukic/fonts": "^2.3.0",
-    "@ukic/web-components": "^2.4.0"
```
GCHQ-Developer-299, line 28: This doesn't seem to be covered by the ticket changes, is this intentional?

jd239: The dependencies are locked down to prevent the web-components and react packages not being in tandem. For example, there are issues if the react package is 2.4.0 and the downloaded web-components package is 2.2.0 for example. This is to prevent that  from happening.

GCHQ-Developer-299 (approved):

ad9242 (approved):

Issue #1065, opened 2023-09-06
SlottedSVG Icons shifted in IcButton


MI6-613:
In 2.3.0 all slottedSvgs with MdiIcons the icons are no longer correctly positioned and are shifted so only the top left corner is shown.

MI6-255:
As part of this issue, could we add MDI Icons to our Visual Tests to prevent issues like this

MI6-255:
This is a React component which we don't use Loki with yet

PR #1093, opened 2023-09-18
1065 slotted svgs shifted in buttons

GCHQ-Developer-530:
## Summary of the changes
Add viewbox to icons in buttons so that there are no sizing/shifting issues when mdi icons or svgs without a viewbox are used.

## Related issue
#1065 

## Checklist


GCHQ-Developer-299 (approved):

packages/react/package.json
```diff
@@ -24,6 +24,7 @@
     "dist/"
   ],
   "dependencies": {
+    "@mdi/js": "^7.2.96",
```
ad9242, line 27: should it be a dev dependency? as it stands it's appearing as a dependency in our built package, so everyone will get it added

GCHQ-Developer-299 (approved):

ad9242 (approved):

Issue #1061, opened 2023-09-05
[@ukic/docs]: Does not get updated when publishing to NPM


jd239:
## Summary of the bug 
When releasing empty state, the docs did not get updated due to json file change not being detected. A work around was to manually update the docs version number and manually publish NPM. This should not be required. 

## 🪜 How to reproduce 
1. Make an update to the props for any component
2. Check if the json file within the docs package has been updated.
3. Check if a commit is required to push the docs update to Github

## 📸 Screenshots or code 
N/A

## 🖥 📱 Device 
- Type: N/A
- Device: N/A
- OS version: N/A
- Browser version: N/A

## 🧐 Expected behaviour 
Commit should be required when the JSON file has been updated. This update will be added to the docs JSON.

## 📝 Acceptance Criteria  
If relevant, describe in full detail the different interactions and edge cases that the component or patterns needs to fulfil. 

*Given that as a developer* 
*When I publish the latest version of components to NPM using automation*  
*Then the docs should also include a commit and published to NPM with the correct version number and changelog* 

## Additional info 
N/A

PR #1095, opened 2023-09-18
1061 ukicdocs does not get updated when publishing to npm

GCHQ-Developer-530:
## Summary of the changes
Remove docs.json from git ignore and update lint rules so that docs version gets updated correctly.

## Related issue
#1061 

## Checklist

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -26,13 +26,13 @@ module.exports = {
         "@stencil-community/decorators-context": "error",
         "@stencil-community/decorators-style": [
             "error", {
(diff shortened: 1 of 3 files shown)


ad9242 (commented):
when building locally, I get changes to docs.json. One is the timestamp, which I would expect, but also looks like dependencies are spreads over multiple lines instead of inline:

image

jd239 (approved):

GCHQ-Developer-299 (approved):

ad9242 (approved):

PR #1097, opened 2023-09-18
fix(web-components): fix for css calc issue in new chromium version

ad9242:
## Summary of the changes
Fixes issue where rgb alpha was not getting applied. this issue was introduced with Chromium v117, & seem to be when the individual r,g,b values exceeded 255
The fix sets upper limit for calculated values to be 255

## Related issue
N/A

## Checklist


MI6-255:
Code looks good, still waiting for the build. Do we need to apply this same logic to the logics JS counterpart? "getThemeColorBrightness"?

ad9242:
> getThemeColorBrightness

think it's ok as that just deals with setting\reading the individual css properties. The issue appeared to be with css calc()

MI6-255 (approved):

GCHQ-Developer-530 (approved):

Issue #860, opened 2023-07-05
[ic-top-navigation] Screen reader reads the navigation items out twice


GCHQ-Developer-530:
## Summary of the bug
When testing with a screen reader, navigation items in the top nav are read out twice.

## 🪜 How to reproduce
Tell us the steps to reproduce the problem:
1. Go to guidance site
2. Turn on screen reader
3. Navigate to navigation items
4. See the error

## 🧐 Expected behaviour
It should only read them out once.

## 📸 Screenshots or code
On our website:
<img width="439" alt="Screenshot 2023-07-05 at 14 55 37" src="https://github.com/mi6/ic-ui-kit/assets/111882034/a253dcea-078e-4433-8893-55eb44ecf84d">
In storybook:
<img width="624" alt="Screenshot 2023-07-05 at 14 49 04" src="https://github.com/mi6/ic-ui-kit/assets/111882034/fc07890f-f042-4a04-8045-a9cc495b3b83">

GCHQ-Developer-299:
From an initial glance, it seems that the issue is an overabundance of aria-labels. If the navigation-item component has a text label, it doesn't need an aria-label. 
`<a slot="navigation-item" href="/get-started" aria-label="Get started">Get started</a>`

Will look to update the navigation-item code

GCHQ-Developer-299:
Looks like it's maybe the aria-described-by tooltips.
ic-buttons, if they are Only Icon, have both a tooltip and an aria-label, which means they too have their names read out twice. Need to make it so that, if there's a tooltip then the aria-label isn't set.

GCHQ-Developer-299:
Re-opening because the issue is still present on navigation items. 
The `<ic-tooltip>` element is still rendered in the shadow-root on navigation-items and is being read out.

To replicate with screenreader: 
- visit ICDS guidance site
- Navigate through elements linearly, do not tab to the navigation items
- You'll find there's an invisible label (the ic-tooltip I think) that reads out, before you navigate to the link which also reads its label

GCHQ-Developer-299:
Reading the mozilla spec for the [tooltip role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role) and I think maybe i've been overcomplicating things. If I can add an aria-describedby to the button, that will meet the spec and maybe silence the double-reading?

GCHQ-Developer-299:
Superceded by https://github.com/mi6/ic-ui-kit/issues/1227

PR #1127, opened 2023-09-27
860 ic top navigation double reading

GCHQ-Developer-299:
## Summary of the changes
Added a silent prop to the ic-tooltip, for instances where the tooltip is duplicating the visible text of the child component. 

Set the tooltips for icon-variant buttons and navigation-items to be silent, as they already have accessible names. 

I've tested this thoroughly with Voiceover (both by tabbing through and moving the virtual cursor) on icon buttons, side/top navigation and ic-tooltip. I've also added a new "silent" story to the ic-tooltip. 

I am unsure how this gels with the [mozilla guidance on role=tooltip](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role) since this change ignores it. Functionally this does solve the double-reading bug and preserves functionality of the tooltip though.

## Related issue
#860

## Checklist


packages/web-components/src/components/ic-tooltip/ic-tooltip.stories.mdx
```diff
@@ -31,6 +31,43 @@ import readme from "./readme.md";
   </Story>
 </Canvas>
 
+### Silent
+
+<Canvas>
+  <Story name="Silent" height="100px">
+    {(args) =>
+      html`<ic-tooltip
```
jd239, line 39: Hovering over this button produces two tooltips. One generated from the browser and the other, an `ic-tooltip`

packages/web-components/src/components/ic-tooltip/ic-tooltip.tsx
```diff
@@ -271,11 +242,11 @@ export class Tooltip {
   render() {
     const { label } = this;
     return (
-      <Host class={{ "ic-tooltip": true }}>
+      <Host class={{ "ic-tooltip": true }} title={!this.silent ? label : null}>
```
jd239, line 245: Title will produce the browser tooltip as well as the `ic-tooltip`

jd239 (changes requested):
Wondering if another way to look at this ticket is to figure out whether aria-hidden on the tooltip should be set to true or false or would that prevent any of it's children such as button from being read out?

GCHQ-Developer-299:
PR has become stale and there's significant feedback to address. Will close and return with another branch.

Issue #1067, opened 2023-09-06
Data table row height sizing options


DBD324:
## Summary 
Within a data table, the row height needs to be able to customised in a few different ways:
- Need a way to specify a fixed height table row
- Need a way to specify an auto-height table row
- Need a way for text wrapping to occur
- Need a way for sizing overrides to occur

## 💬 Description 
A full description of what the component or pattern should be and how it should work. 

## 💰 User value 
Describe why this is important and how it will help users (either developers, designers or the end-users). Think about how it makes people's lives easier, more accessible or quicker to build stuff. 

## 📚 User Stories 
If relevant, describe the high-level functionality as user stories. 

*As an* ICDS user:  
*I need* a way to set the row height of every row in a table
*So that* they can all be the same size. 

*As an* ICDS user:  
*I need* a way to use a default row height for every row in a table
*So that*I don't have to manually set it. 

*As an* ICDS user:  
*I need* a way to set the row height of individual rows in a table
*So that* they can be different sizes.

*As an* ICDS user:  
*I need* a way to set the row height of a group of rows in a table
*So that* they can be different sizes.

*As an* ICDS user:  
*I need* a way to reset the row height of every row in a table
*So that* they can all be returned to the default.

*As an* ICDS user:  
*I need* a way to set a row height to be automatic
*So that* it contains the content within. 

*As an* ICDS user:  
*I need* a way to set the estimated row height 
*So that* row heights can be set before data loads. 

## 📝 Acceptance Criteria  
If relevant, describe in full detail the different interactions and edge cases that the component or patterns needs to fulfil. 

*Given* that a table is being implemented
*When* setting the row height to a single positive integer value,
*Then* it should apply to all rows in the table.

*Given* that a table is being implemented
*When* no custom row-height value has been set
*Then* a default of 40px should be used.

*Given* that a table is being implemented
*When* setting its properties
*Then* an option exists to set the row height differently for different rows,
*Or* to set the row height differently for different groupings of rows. 
(e.g perhaps every other row needs to be set differently, or rows that contain a specific value need to be set differently)

*Given* that a table is being implemented
*When* some rows have a specific row height set, but not all, 
*Then* the un-set rows use the default row height. 

*Given* that a table has been implemented with custom row heights
*When* using the reset row height function
*Then* row heights should be reset to the default row height. 

*Given* that a table's rows are being set
*When* the row heights change
*Then* an event should occur that can be used to trigger other functions. 

**Auto row-height**
[...]

PR #1171, opened 2023-10-11
1067 row height

GCHQ-Developer-741:
## Summary of the changes
Added two new props: `rowHeight` to override all row's heights, and `getRowHeight` to pass a custom function to apply different row heights based on different properties. `getRowHeight` is passed the data object from inside data tables, meaning that people will be able to filter based on the data properties they pass through.

## Related issue
#1067 

## Checklist


packages/web-components/src/components/ic-data-table/ic-data-table.tsx
```diff
@@ -90,6 +101,15 @@ export class DataTable {
    */
   @Prop() embedded?: boolean = false;
 
+  /**
+   * Allows for custom setting of row heights on individual rows based on an individual value from the `data` prop and the row index.
+   * If the function returns `null`, that row's height will be set to the `rowHeight` property.
+   */
+  @Prop({ mutable: true }) getRowHeight?: (parmas: {
```
GCHQ-Developer-299, line 108: Feels weird to me that a function called 'get' would be used for setting a value. Can you explain the logic here?

GCHQ-Developer-741: I think the logic is that calling the function returns (gets) the number to be used for the row height. The actual logic for setting it is done a bit further down (line 587)

packages/web-components/src/components.d.ts
```diff
@@ -498,6 +498,13 @@ export namespace Components {
           * Applies a border to the table container.
          */
         "embedded"?: boolean;
+        /**
+          * Allows for custom setting of row heights on individual rows based on an individual value from the `data` prop and the row index. If the function returns `null`, that row's height will be set to the `rowHeight` property.
+         */
+        "getRowHeight"?: (parmas: {
```
GCHQ-Developer-530, line 504: Is this meant to be params?

packages/web-components/src/components/ic-data-table/ic-data-table.tsx
GCHQ-Developer-299: Thanks for explaining. I'm still a bit confused since this is a mutable property (rather than a function, which I assume all "getX" methods would be) - I guess my next question is: could this prop be called just "rowHeight"?  

I've done a global search of the project in VSCode and there's no other prop that starts with "get"

GCHQ-Developer-741: rowHeight is already used by another property which sets the height for all rows. The fact that it is mutable is because this prop is overridden on line 309. I think rowHeight would be misleading for this as its about getting the height for rows that meet a specific criteria, rather than setting it more generally, if that makes sense

GCHQ-Developer-299: Then i'd suggest 'rowHeight' should be renamed to 'universalRowHeight' or 'allRowHeights' 
and 'getRowHeight' to 'rowHeight' to better differentiate the scopes between them.

Sorry for not understanding the full picture before now!

GCHQ-Developer-530 (approved):

packages/web-components/src/components/ic-data-table/ic-data-table.tsx
GCHQ-Developer-741: Have updated to globalRowHeight and variableRowHeight

GCHQ-Developer-299 (approved):

PR #1183, opened 2023-10-18
fix(web-components): change how ic-button handles tooltips (to solve screenreaders double-reading)

GCHQ-Developer-299:
## Summary of the changes
- removed the SR-only span in ic-tooltip that used to act as aria-describedby, so that parent elements can use it.
- ic-button now sets the id of ic-tooltip and applies that id to 'aria-describedby' to link them. Now Voiceover knows that the tooltip text belongs to the button and says "more content available" when you put the cursor on a button.
- "Tooltips" examples added to the Button storybook page to give an example of good tooltips (that provide additional information beyond the button's accessible label)

Our tooltips now perform the same as the example tooltip on [Mozilla's aria role:tooltip example](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role)

## Related issue
#860

## Checklist


packages/web-components/src/components/ic-button/ic-button.tsx
```diff
@@ -80,11 +79,6 @@ export class Button {
    */
   @Prop() disabled?: boolean = false;
 
-  /**
-   * If `true`, the ic-tooltip which is shown for icon variant will be disabled. Title or aria-label must be set if this prop is not applied.
-   */
-  @Prop() disableTooltip?: boolean = false;
```
GCHQ-Developer-530, line 86: I think instead of completely deleting this prop, we need to deprecate it in case people are currently using it. Then we can delete it when we come to do v3 work.

GCHQ-Developer-530 (changes requested):

packages/web-components/src/components/ic-button/ic-button.tsx
GCHQ-Developer-299: Done :)

GCHQ-Developer-530 (approved):

packages/docs/docs.json
```diff
@@ -1298,9 +1298,15 @@
           "mutable": false,
           "attr": "disable-tooltip",
           "reflectToAttr": false,
-          "docs": "If `true`, the ic-tooltip which is shown for icon variant will be disabled. Title or aria-label must be set if this prop is not applied.",
-          "docsTags": [],
+          "docs": "",
+          "docsTags": [
+            {
+              "name": "deprecated",
+              "text": "No longer functional, by default tooltips do not appear on buttons. use tooltipLabel to add a tooltip"
```
GCHQ-Developer-465, line 1305: Should this be a capital? And is punctuation needed at the end?
```suggestion
              "text": "No longer functional, by default tooltips do not appear on buttons. Use tooltipLabel to add a tooltip."
```

GCHQ-Developer-465, line 1305: Same here?
```suggestion
              "text": "No longer functional, by default tooltips do not appear on buttons. Use tooltipLabel to add a tooltip."
```

packages/docs/docs.json
```diff
@@ -1298,9 +1298,15 @@
...
+            }
+          ],
           "default": "false",
+          "deprecation": "No longer functional, by default tooltips do not appear on buttons. use tooltipLabel to add a tooltip",
```
GCHQ-Developer-465, line 1309:
```suggestion
          "deprecation": "No longer functional, by default tooltips do not appear on buttons. Use tooltipLabel to add a tooltip",
```

packages/web-components/src/components.d.ts
```diff
@@ -216,7 +216,7 @@ export namespace Components {
         "ariaControlsId": string | boolean;
         "ariaOwnsId": string | boolean;
         /**
-          * If `true`, the ic-tooltip which is shown for icon variant will be disabled. Title or aria-label must be set if this prop is not applied.
+          * @deprecated No longer functional, by default tooltips do not appear on buttons. use tooltipLabel to add a tooltip
```
GCHQ-Developer-465, line 219:
```suggestion
          * @deprecated No longer functional, by default tooltips do not appear on buttons. Use tooltipLabel to add a tooltip.
```

packages/web-components/src/components.d.ts
```diff
@@ -2821,7 +2825,7 @@ declare namespace LocalJSX {
         "ariaControlsId"?: string | boolean;
         "ariaOwnsId"?: string | boolean;
         /**
-          * If `true`, the ic-tooltip which is shown for icon variant will be disabled. Title or aria-label must be set if this prop is not applied.
+          * @deprecated No longer functional, by default tooltips do not appear on buttons. use tooltipLabel to add a tooltip
```
GCHQ-Developer-465, line 2828:
```suggestion
          * @deprecated No longer functional, by default tooltips do not appear on buttons. Use tooltipLabel to add a tooltip.
```

packages/web-components/src/components/ic-button/ic-button.tsx
```diff
@@ -81,7 +80,7 @@ export class Button {
   @Prop() disabled?: boolean = false;
 
   /**
-   * If `true`, the ic-tooltip which is shown for icon variant will be disabled. Title or aria-label must be set if this prop is not applied.
+   * @deprecated No longer functional, by default tooltips do not appear on buttons. use tooltipLabel to add a tooltip
```
GCHQ-Developer-465, line 83:
```suggestion
   * @deprecated No longer functional, by default tooltips do not appear on buttons. Use tooltipLabel to add a tooltip.
```

GCHQ-Developer-465 (commented):
Awesome work 👍 

Just small comments on documentation.

packages/web-components/src/components/ic-button/ic-button.tsx
GCHQ-Developer-847, line 1: Getting a lint warning about the `title` variable on line 361 not being used

packages/web-components/src/components/ic-button/ic-button.stories.mdx
```diff
@@ -876,6 +876,43 @@ export const defaultArgs = {
   </Story>
 </Canvas>
 
+### With tooltip
+
+<Canvas>
+  <Story name="With Tooltip">
```
GCHQ-Developer-847, line 882:
```suggestion
  <Story name="With tooltip">
```

packages/web-components/src/components/ic-button/ic-button.tsx
GCHQ-Developer-847, line 1: The fix works really well for the 'Icon only' story, but I'm getting some repetition with the 'With tooltip' story :(

If I tab to each of the buttons, it repeats the visible label of the button at the end, e.g.:
- "refresh refresh the current page **refresh**, button, group"
- "Submit Submitting the form will refresh the page **Submit**, button, group"

Maybe this isn't a huge problem though? The 'Icon only' ones are fine which I think was the main thing

GCHQ-Developer-847 (changes requested):
A few comments :)

packages/web-components/src/components/ic-button/ic-button.tsx
GCHQ-Developer-530: Since the issue on our guidance site is on the top navigation nav items, does that mean that it would repeat the label at the end of those still then?

GCHQ-Developer-847: I've had a look at the top navigation in Storybook and it's only reading the nav item labels out once, so seems like it's just an issue with the button

GCHQ-Developer-299: I've worked out that this is being caused by the tooltip slotting the button element, so the screen reader reads Button > Tooltip > Button again. 

It looks like this can be solved by aria-hiding the slot, but I am worried about further side effects

GCHQ-Developer-299: I think this might be a bug with Storybook (might be caused by the above, but maybe I was grasping at straws)
if you refresh the page, Voiceover no longer reads the button label for a second time after the tooltip.

GCHQ-Developer-465 (approved):

packages/web-components/src/components/ic-button/ic-button.tsx
GCHQ-Developer-847: I think you're probably right about it being a Storybook issue!

I think I properly understand it all now - that the aria-label value should be the label of the element and the title should be the description of element, so it does seem like it's repeating "refresh" on the "Icon only" story, but then if I add a title attribute, it reads the aria-label and then the title (well, after I've reloaded the page), which is equivalent to reading out the button label and then the tooltip label.

And it only reads out the navigation item labels once. So all looks really good to me! 👍 Happy to approve

Issue #844, opened 2023-06-28
Button pressed styling not applied on click


ad9242:
## Summary of the bug
When clicking on an ic-button, the pressed styling is not applied. This is due to previous work to make the focused styling take priority (and as soon as the mouse button is pressed down, the browser will give it focus)

## 🪜 How to reproduce
1) go to https://mi6.github.io/ic-ui-kit/branches/develop/?path=/story/web-components-button--secondary
2) hover on a button & it goes light blue but 
3) hold your mouse button down on the button & it goes to white (instead of a slightly darker blue as specified in Figma)

## 🧐 Expected behaviour
In order to match the Figma specs, the pressed state should be applied. If not, then the Figma examples\prototypes should be updated to match the behaviour, as it is currently confusing seeing different behaviour in design vs dev

## Additional info
This may require a bigger review of how the pressed state should apply on mouse click, to ensure consistency across all interactive components

PR #1203, opened 2023-10-19
fix(web-components): applies active button styling when button pressed

j212122:
## Summary of the changes

Applies active/ pressed styling to button

## Related issue
#844 

## Checklist
- [] Relevant unit tests and visual regression tests added. 

### DIFF
--- a/packages/web-components/package-lock.json
+++ b/packages/web-components/package-lock.json
@@ -22126,9 +22126,9 @@
       }
     },
     "node_modules/postcss": {
-      "version": "8.4.29",
-      "resolved": "https://registry.npmjs.org/postcss/-/postcss-8.4.29.tgz",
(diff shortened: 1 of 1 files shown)


GCHQ-Developer-299 (approved):
Tested with mouse and keyboard, compared against Mozilla's [active button demo](https://developer.mozilla.org/en-US/docs/Web/CSS/:active) and interactivity lines up

DBD324 (changes requested):
We still have an issue whereby when a button is focussed, the hover and pressed states are then not displayed when interacting via a mouse. 

The focus and interaction states should almost be considered separate, in that whilst focussed, all other states should still apply. It is possible to have 'hover + focussed' and 'active/pressed + focussed'. 

I think the order of operation should be as follows:
1. Button exists on page unfocussed or hovered showing default state. 
2. Mouse over causes hover state. 
3. Mouse down causes active state plus applies focus
4. Mouse up event causes complete click event (if still over button) and button remains hovered and focussed. 
5. Mouse leaves button area causing default state to return, with focus still displayed.

PR #1212, opened 2023-10-23
Audit fixes and postcss version bump

ad9242:
## Summary of the changes
fixes audit issue and bumps postcss to 8.4.31 as recommended by dependabot

## Related issue
N/A


GCHQ-Developer-530 (approved):

MI6-255:
Does it need updating at the root level too?

ad9242:
> Does it need updating at the root level too?

no, it's already 8.4.31 there

Issue #1248, opened 2023-11-07
Top Navigation: App Icon is not part of homepage link's clickable area + clickable area doesn't function on mobile


GCHQ-Developer-299:
## Summary of the bug 
Currently app icon is part of the highlighted area in the App Title, but is not part of the anchor element so when users try to click on it, they get an unexpected lack of behaviour.

## 🪜 How to reproduce 
Tell us the steps to reproduce the problem: 
1. Go to ICDS Guidance Site on Desktop
2. Hover over the ICDS logo in the top navigation, note how it's part of the highlighted area around the app name
3. Try clicking the icon, find that it doesn't actually count as part of the clickable area.

1. View the top navigation in mobile orientation (make your browser narrow, or use responsive mode in devtools)
2. try to click the app title, find that none of it triggers the link, despite appearing clickable

## 📸 Screenshots or code 
![Screenshot 2023-11-07 at 16 43 13](https://github.com/mi6/ic-ui-kit/assets/111292477/bc2c5781-1404-46d4-bd9e-1a12639ba0cb)
![Screenshot 2023-11-07 at 16 43 25](https://github.com/mi6/ic-ui-kit/assets/111292477/928662ae-85d6-4772-a4f7-cd4ad783ddd4)

## 🧐 Expected behaviour 
Clicking on the app icon will trigger the link behaviour of the app title, bringing it inline with the visible highlight

PR #1282, opened 2023-11-22
1248 top navigation app icon is not part of homepage links clickable area

GCHQ-Developer-530:
## Summary of the changes
Add short app title slot for top navigation to handle routing on small screen sizes.
Add styling to cover cases where non svgs are slotted into app-icon (e.g. a link)
Added tests and a story.

## Related issue
#1248 

## Checklist


packages/web-components/src/components/ic-top-navigation/ic-top-navigation.tsx
```diff
@@ -330,17 +331,24 @@ export class TopNavigation {
                   {(hasTitle || isSlotUsed(this.el, "app-title")) && (
                     <Component class="title-link" {...attrs}>
                       {this.hasAppIcon && (
-                        <div class="app-icon-container" aria-hidden="true">
+                        <div class="app-icon-container">
```
ad9242, line 334: just checking the reason for removing the aria-hidden. previously I think this was added as the icon does add any extra information. happy if it's been agreed though!

packages/docs/docs.json
```diff
@@ -16955,6 +16959,10 @@
           "name": "search",
           "docs": "Content will be rendered in search area to left of buttons."
         },
+        {
+          "name": "short-app-title",
+          "docs": "Handle routing by nesting a route in the short app title."
```
ad9242, line 16964: feel like we need to make clearer when to use this slot i.e. for use when you want a shorter title on smaller devices size. Also make it clear that you can use both. Not sure if this is the place though, or if it comes more under guidance on the site. Maybe we could add the new example to the site, with additional info on how\why it does what it does??

packages/web-components/src/components/ic-top-navigation/test/basic/__snapshots__/ic-top-navigation.spec.ts.snap
```diff
@@ -311,6 +311,64 @@ exports[`ic-top-navigation renders with slotted app-title content: renders-with-
...
+exports[`ic-top-navigation renders with slotted short-app-title: renders-with-slotted-short-title 1`] = `
+<ic-top-navigation status="alpha" version="v0.0.7">
+  <mock:shadow-root>
+    <div class="top-navigation">
+      <ic-section-container aligned="full-width" full-height="">
+        <header role="banner">
+          <div class="top-panel-container">
+            <div class="app-details-container">
+              <div class="title-link">
+                <ic-typography aria-label="undefined ()" variant="subtitle-small">
```
ad9242, line 323: is the "undefined ()" expected?

packages/web-components/src/components/ic-top-navigation/ic-top-navigation.tsx
GCHQ-Developer-299: the icon used to be aria-hidden because we assumed only an image would be put in there. (since the slot is 'app-icon') but we didn't tell developers this was the case and it tripped up one of our customer applications

GCHQ-Developer-530: Yeah we need to remove the aria-hidden now so that when people slot links in then they'll get read out. Like @GCHQ-Developer-299 mentioned, it tripped up a customer recently

packages/web-components/src/components/ic-top-navigation/test/basic/__snapshots__/ic-top-navigation.spec.ts.snap
GCHQ-Developer-530: Ooh no, I'll take a look

packages/docs/docs.json
GCHQ-Developer-530: Guidance site update is [here](https://github.com/mi6/ic-design-system/pull/694). It will need to be built locally to see it working, but hopefully it adds a bit more info

GCHQ-Developer-299 (approved):

ad9242 (approved):

Issue #1242, opened 2023-11-06
[ic-dialog] dismiss icon button does not show in safari


GCHQ-Developer-530:
## Summary of the bug 
View any of the dialog stories in storybook on safari, the dismiss icon is there as you can see the tooltip and focus, but it doesn't actually visually show anything.

## 🪜 How to reproduce 
Tell us the steps to reproduce the problem: 
1. Go to dialog stories on safari
2. Open a dialog
3. See no dismiss icon button

## 📸 Screenshots or code 
<img width="439" alt="Screenshot 2023-11-06 at 13 14 39" src="https://github.com/mi6/ic-ui-kit/assets/111882034/ac8404ed-7872-4ed3-adc3-707978a4a9d0">

## 🖥 📱 Device 
- Type: Desktop
- Device: MacBook
- Browser version: Safari 17.1

## 🧐 Expected behaviour 
Should be able to see the dismiss button.

ASM995:
Linked to #1263

PR #1284, opened 2023-11-23
fix(web-components): move close icon svg into button for dialog

GCHQ-Developer-530:
## Summary of the changes
Move close icon svg into button for dialog instead of its own span to ensure it works across all browsers

## Related issue
#1242

## Checklist


GCHQ-Developer-299 (approved):

packages/web-components/src/components/ic-dialog/ic-dialog.tsx
```diff
@@ -514,12 +514,11 @@ export class Dialog {
               <ic-button
                 class="close-icon"
                 variant="icon"
+                innerHTML={closeIcon}
                 aria-label={dismissLabel}
                 onClick={this.closeIconClick}
                 data-gets-focus={destructive || !buttons ? "" : null}
-              >
-                <span class="close-icon-svg" innerHTML={closeIcon} />
-              </ic-button>
```
jd239, line 522: Has this been updated because there is a problem in Safari or a problem with the implementation? Would expect this to work as it's just an SVG in a span.

jd239 (commented):
Single comment. More of a query.

packages/web-components/src/components/ic-dialog/ic-dialog.tsx
jd239: What happens with it's just the SVG not wrapped in a span?

GCHQ-Developer-530: This has been updated because of the Safari issue. In other places where close icons appear within icon buttons, we just use innerHTML to display them, this was the only place we used a span and I thought it'd be the most simple solution in order to get it in line with other components.

GCHQ-Developer-299 (approved):

packages/web-components/src/components/ic-dialog/ic-dialog.tsx
jd239: Makes sense for consistency. Will approve.

It's worth creating another ticket to look into why the SVG didn't appear (within the span) in Safari in case an update is required within `ic-button`.

jd239 (approved):

packages/web-components/src/components/ic-dialog/ic-dialog.tsx
GCHQ-Developer-530: I'll create a ticket now. It works within a span if you're slotting an icon into a regular button so potentially only an issue with the icon variant.

Issue #1273, opened 2023-11-17
[ic-select]: Allow `icOptionSelect` event to be fired on `Enter` instead of `ArrowDown`


GCHQ-Developer-112:
## Summary 
Currently, the events `icOptionSelect` and `icChange` are fired when a user moves through the menu. This new improvement to `ic-select` would allow me as a developer to choose whether I want the `icOptionSelect` event to be fired when the user moves through the menu or when they click on an option or select it using the `Enter` key. The assumption at the moment is that this will work using a prop. 

## 💰 User value 
This will allow me to swap out the MUI select component I'm using for the ICDS select component. As they will be equivalent in functionality. It will also allow the selection of an option using the `Enter` key to trigger a secondary action.

_Raised on behalf of a user_

PR #1297, opened 2023-12-04
1273 ic select allow icoptionselect event to be fired on enter instead of arrowdown

GCHQ-Developer-530:
## Summary of the changes
Add selectOnEnter prop to select component to determine when icOptionSelect is emitted (either on enter or arrow down)

## Related issue
#1273 

## Checklist


GCHQ-Developer-299:
I'm not sure this fulfils the requirement - while it does only emit the event on enter, the value that appears selected in the select field still changes when you arrow through the menu. 
This leads to unusual behaviour where if you arrow through to an option, then press Esc to close the menu, then open the menu again using spacebar, the event fires on the item that was in the field (even though the user never actually pressed enter on that item)

GCHQ-Developer-299 (dismissed):

ad9242:
found an issue with the async select default story if you set the new prop. if you make a selection, say "espresso", then use arrow down key, you can't see the current focused item/items above the selected at the top of the list. 

eg below the current focused item is not visible:
image

seems to be it doesn't scroll the item into view in this mode

GCHQ-Developer-299 (approved):

ad9242 (changes requested):
can we add a test to make sure the event is only emitted when enter is pressed?

ad9242 (approved):

GCHQ-Developer-299 (approved):

PR #1304, opened 2023-12-05
fix(web-components): open ic-dialog during initialisation using the open prop

dz333dz:
## Summary of the changes
A new open prop has recently been added to the IcDialog component, which replaces the deprecated showDialog and hideDialog methods.

The trouble is, when an instance of an IcDialog component is initialised using the new open prop, with open="true", the dialog remains hidden. This is because the dialog is only made visible when the value of the open prop changes, i.e. when it is assigned a different value after the component has been initialised (which happens in the watchOpenHandler method).

In particular, this causes a problem for testing IcDialog based components using Cypress component tests, as it is currently not possible to instantiate a component using a prop of open="true" and then just go ahead and run assertions in Cypress, as a change of state is needed for the open prop to make the dialog visible in the tests, which involves some hacky and unreliable workarounds to get it working.

The small change in this PR fixes the issue. Now, if users of the IcDialog component pass in open="true", then the dialog is made visible during initialisation within the componentDidLoad function, reusing the existing code that executes whenever the watchOpenHandler is executed.

## Related issue
N/A

## Checklist

### DIFF
--- a/packages/web-components/src/components/ic-dialog/ic-dialog.tsx
+++ b/packages/web-components/src/components/ic-dialog/ic-dialog.tsx
@@ -117,38 +117,7 @@ export class Dialog {
   @Watch("open")
   watchOpenHandler(): void {
     if (this.open) {
-      this.dialogRendered = true;
-
-      if (this.disableHeightConstraint) {
-        this.dialogEl.show();
-      } else {
-        this.dialogEl.showModal();
-      }
-
-      setTimeout(() => {
-        this.fadeIn = true;
-
-        /**
-         * This is required to set scroll back to top if:
-         * - dialog content goes below the fold
-         * - is closed using cancel or confirm and reopened.
-         *
-         * Without this, the scroll bar will start from the dialog's last scroll-x coordinate.
-         */
-        if (this.disableHeightConstraint && this.backdropEl.scrollTop !== 0) {
-          this.backdropEl.scrollTop = 0;
-        }
-      }, 10);
-
-      setTimeout(() => {
-        this.setInitialFocus();
-        checkResizeObserver(this.runResizeObserver);
-      }, 75);
-
-      setTimeout(() => {
-        this.getFocusedElementIndex();
-        this.icDialogOpened.emit();
-      }, 80);
+      this.dialogOpened();
     } else {
       this.fadeIn = false;
       if (this.resizeObserver !== null) {
@@ -225,6 +194,10 @@ export class Dialog {
     this.setAlertVariant();
 
     this.refreshInteractiveElementsOnSlotChange();
+
+    if (this.open) {
+      this.dialogOpened();
+    }
   }
 
   componentDidRender(): void {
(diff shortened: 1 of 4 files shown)


packages/web-components/src/components/ic-dialog/ic-dialog.tsx
```diff
@@ -117,46 +117,17 @@ export class Dialog {
...
+      this.dialogOpened();
     } else {
       this.fadeIn = false;
       if (this.resizeObserver !== null) {
         this.resizeObserver.disconnect();
       }
       setTimeout(() => {
         this.dialogRendered = false;
-        this.dialogEl.close();
+        if (this.dialogEl?.close) {
```
jd239, line 128: I do not think this if statement is required as `dialogEl` is always going to be a `<dialog>` element and the close method comes standard with it. 
ref: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog

jd239 (changes requested):
Nice addition to the dialog component. Thanks for the contribution. 🚀 
A few things:
- Could you please create the 'Auto Open' story for the Web Components storybook

- The commits need to be separated based on the update to the package so for example:
   - The web component commit scope should only contain the update to dialog.tsx and the new web component story
   - Docs commit for the update to docs.json
   - React commit for React storybook

Hopefully this helps: https://github.com/mi6/ic-ui-kit/blob/main/CONTRIBUTING.md#git-commit

Thanks.

packages/web-components/src/components/ic-dialog/ic-dialog.tsx
dz333dz: Agreed - this was left over from trying to getting this to work using a different lifecycle method. I will revert this change.

dz333dz:
> Nice addition to the dialog component. Thanks for the contribution. 🚀 A few things:
> 
> * Could you please create the 'Auto Open' story for the Web Components storybook
> * The commits need to be separated based on the update to the package so for example:
>   
>   * The web component commit scope should only contain the update to dialog.tsx and the new web component story
>   * Docs commit for the update to docs.json
>   * React commit for React storybook
> 
> Hopefully this helps: https://github.com/mi6/ic-ui-kit/blob/main/CONTRIBUTING.md#git-commit
> 
> Thanks.

Apologies, I missed the fact that there are 2 different storybooks to update! I will make sure that the "Auto opening" (as well as the "Hidden close button" story that I recently added) appears in both storybooks.

I will also look into making the changes in separate commits, as suggested.

dz333dz:
All comments addressed - ready for re-review.

GCHQ-Developer-530 (changes requested):
Hi @dz333dz thank you for your contribution! Looks really good, just one small change. Please could you update both storybook examples to have a button to reopen the dialog after it has been closed? At the moment it just leaves the blank page as there's nothing there to reopen it. Thank you

GCHQ-Developer-530 (approved):

PR #1338, opened 2023-12-21
Add IcAlert Cypress tests

mmb675:
## Summary of the changes
Add IcAlert Cypress tests: Component, Visual Regression, Accessibility


packages/react/src/component-tests/IcAlert/IcAlert.cy.tsx
```diff
@@ -0,0 +1,344 @@
...
+      .should("be.visible")
+      .contains("Title displayed above the message");
+    cy.findShadowEl("ic-alert", "ic-typography")
+      .contains(
+        "This alert has the title at the top of the alert and the message below it."
+      )
+      .should("be.visible");
+  });
+
+  it.s("Custom with Message and Icon", () => {
```
ad9242, line 168:
```suggestion
  it("Custom with Message and Icon", () => {
```
this is causing tests to fail to run for me

packages/react/src/component-tests/IcAlert/IcAlert.cy.tsx
```diff
@@ -0,0 +1,344 @@
+/// <reference types="Cypress" />
+import { mount } from "@cypress/react";
+import React from "react";
+import { IcAlert, IcButton, IcLink } from "../../components";
+import { CYPRESS_AXE_OPTIONS } from "../../../cypress/utils/a11y";
+
+describe("IcAlert", () => {
+  it("Neutral Alert with Message and Icon", () => {
```
jd239, line 8: We could add the visual regression and a11y tests for the Netural alert within this test. It doesn't need to be in a separate describe block. This could be the same for all the other variants as well. Add an visual regression/a11y test for each variant in the relevant tests below.

jd239 (changes requested):
Added a comment which involves moving some parts around.

GCHQ-Developer-847:
Think this needs rebasing as it's showing test files for other components rather than just the alert

GCHQ-Developer-299:
This PR was superceded by https://github.com/mi6/ic-ui-kit/pull/1383 - much of the code here was reused in it :) 
Will close this PR now

PR #1350, opened 2024-01-03
test(react): refactoring select tests

jd239:
## Summary of the changes
- Refactoring select / searchable tests
- Removed redundant screenshots. 
   - For example, removed 'required' screenshot from searchable as the same screenshot is available for select and this only affects the label. 
- Removed separate visual regression block
- Separated select and searchable

## Related issue
N/A

## Checklist


packages/react/src/component-tests/IcSelect/IcSelectSearchable.cy.tsx
```diff
@@ -0,0 +1,938 @@
...
+        searchable
+      />
+    );
+    cy.checkHydrated("ic-select");
+    cy.compareSnapshot(
+      "searchable-default-value",
+      DEFAULT_TEST_THRESHOLD + 0.01
+    );
+
+    cy.get("ic-select").shadow().find(IC_INPUT_CONTAINER).type("ca");
```
GCHQ-Developer-530, line 645: Do we need this line in this test? Instead of it showing that the default value shows results, it tries searching for "cappuccinoca" so the screenshot shows no results found.

jd239: Agreed. The test and the title is misleading. Will remove this action and update the screenshot below as well.

GCHQ-Developer-530 (approved):

GCHQ-Developer-465 (approved):

Issue #416, opened 2023-02-22
Add built-in events to website documentation


MI6-255:
## Summary
Add built-in events to website documentation, to make it clear about events such as onclick on a Button for example.

## 💬 Description
Navigating to Button for example, going to the "Code" tab and looking at events. As a consumer, I'd expect to see all events I can use, such as onclick. This ticket covers looking into if it is feasible and applying all missing events on all components.

## 💰 Use value
Make it easier for developers to see how to use the components.

GCHQ-Developer-465:
After further investigation with this ticket additional information is needed, thus I am adding this comment below to help anyone who looks into this ticket again. 

Focus on the following...

Explore what the ICDS should be doing about built in events (e.g. OnClick, change if relevant, etc). StencilJS just passes on anything built in like OnClick to the component. For instance, if you look at [Button](https://design.sis.gov.uk/components/button), you can see no event for “click” or “onclick” or anything similar. That’s because Stencil is passing on those DOM events, see more [here](https://stenciljs.com/docs/events). 

Therefore the issue is, as a consumer, there are no **code examples** or **events documented** to show how to add 'on click' to a button, which may be obvious to some, but maybe not new developers or those using ICDS for the first time. 

Thus, it’d be worth **exploring** how we make it obvious e.g. is it code examples:
- is it somehow getting the more obvious events (such as OnClick for button) in the docs somehow
- is it worth just linking from all the events to a page that shows what built in events may work? 

You should also consider looking at other libraries.

GCHQ-Developer-299:
Material UI has a section on their Button docs for [Handling Clicks](https://mui.com/material-ui/react-button/#handling-clicks) that explains that all their components accept the onClick handler. Below that they also state that:
> Note that the documentation [avoids](https://mui.com/material-ui/guides/api/#native-properties) mentioning native props (there are a lot) in the API section of the components.

I look at Vuetify as another example, and they don't explicitly mention the @click event at all. 

My recommendation would be that we add a "Native properties and events" page to the documentation that explains how our components allow native web component / react events. We could then link to that from relevant component pages (e.g. Button)

Thoughts?

PR #1353, opened 2024-01-08
416/built in events

GCHQ-Developer-465:
## Summary of the changes
Add built in events for components to display on the website. 

Updated the events from internal.

## Related issue
#416


ad9242:
Just a quick comment. I think any events that we are removing the "@internal" decorator from, we should prefix with "Ic" in order to be consistent in our naming convention of events

PR #1382, opened 2024-01-15
fix(web-components): updated manualInputKeyboardOpen function

jd239:
## Summary of the changes
Updated manSetInputValueKeyboardOpen function to prevent setFocus from being triggered every time the keyDown event was triggered. This bug was evident in Cypress tests as the onBlur event was being triggered on every input

## Related issue
N/A

## Checklist


packages/web-components/src/components/ic-menu/ic-menu.tsx
```diff
@@ -603,7 +603,10 @@ export class Menu {
         if (this.isSearchBar) {
```
ad9242, line 603: Is it worth moving this code into a separate function to avoid duplication? Seems to be exactly the same logic as in the default case when (event.key !== "Tab")

packages/web-components/src/components/ic-menu/ic-menu.tsx
```diff
@@ -600,24 +600,11 @@ export class Menu {
...
-        if (event.key !== "Tab") {
-          if (this.isSearchBar) {
-            (this.parentEl as HTMLIcSearchBarElement).setFocus();
-            if (this.searchMode === "navigation") this.setHighlightedOption(0);
-          } else if (this.isSearchableSelect) {
-            (this.parentEl as HTMLIcSelectElement).setFocus();
-          }
-          this.focusFromSearchKeypress = true;
-        }
+        if (event.key === "Tab") return;
```
GCHQ-Developer-299, line 606: shouldn't this be covered by the "Tab" case on line 596?

jd239: Yeah. That's right. The code has been updated. Since the default and backspace behaviour is the same, `case Backspace` has been removed as this should be captured by `default`

packages/web-components/src/components/ic-menu/test/basic/ic-menu.spec.tsx
```diff
@@ -608,13 +609,123 @@ describe("ic-menu in isolation", () => {
...
+
     page.rootInstance.manSetInputValueKeyboardOpen(keyboardEvent("Backspace"));
 
     await page.waitForChanges();
 
     expect(searchableSelect.setFocus).toHaveBeenCalled();
   });
-  it("tests manSetInputValueKeyboardOpen function when searchbar", async () => {
+
+  it("tests setFocus in not called when backSpace is pressed and no menu option is highlighted", async () => {
```
GCHQ-Developer-299, line 621: tiny typo, should be 'is not called' 
this won't stop me approving though

GCHQ-Developer-299 (approved):

packages/web-components/src/components/ic-menu/test/basic/ic-menu.spec.tsx
jd239: Happy to make the update

GCHQ-Developer-299 (approved):

ad9242 (approved):

Issue #1301, opened 2023-12-05
Migrate Alert testing over to Cypress


mmb675:
## Summary 
Migrate Alert testing over to Cypress

## 💬 Description 
As per #1166, Cypress testing needs to be migrated over to Cypress for this component. This involves migrating over every e2e test we currently have, adding accessibility and visual testing. This also involves removing the old frameworks code, so two of the same tests do not run (including Loki and e2e). 

## 💰 Use value 
Improve our testing toolset to allow us to test in more states and deliver a higher quality product.

PR #1383, opened 2024-01-15
Add IC Alert tests to cypress

GCHQ-Developer-299:
## Summary of the changes
Add Cypress IC Alert tests and remove the old IC Alert tests from the web components repo

Supercedes https://github.com/mi6/ic-ui-kit/pull/1338

## Related issue
#1301


GCHQ-Developer-530 (dismissed):

GCHQ-Developer-530 (approved):

packages/react/cypress-visual-screenshots/baseline/IcAlert.cy.tsx-title-displayed.png
GCHQ-Developer-741, line 1: Just an observation but there is a specific test for the title-above variant, but all the screenshots appear to feature the title above. Should there be a reference image of one that features the heading and message inline with each other, in case that behaviour is altered in the future without us noticing?

packages/react/cypress-visual-screenshots/baseline/IcAlert.cy.tsx-dismiss.png
GCHQ-Developer-741, line 1: I don't really see the purpose of having a blank image to cover a dismissable alert. I feel like it would be better to take the screenshot before the alert gets dismissed, to capture the necessary visual details.

GCHQ-Developer-299: I agree with you - and I don't think the current test is checking that the alert actually gets dismissed either...

packages/react/cypress-visual-screenshots/baseline/IcAlert.cy.tsx-title-displayed.png
GCHQ-Developer-299: Good thinking! I've changed the size of the viewport so the regular alerts display inline, which allows the title-above alerts to stand out :)

GCHQ-Developer-741 (approved):

GCHQ-Developer-530 (approved):

PR #1444, opened 2024-01-26
Fix Malformed Tooltips when label not provided

GCHQ-Developer-299:
## Summary of the changes
Before, if a label wasn't provided to a tooltip component (or icon variant button, which internally renders a tooltip):
Screenshot 2024-01-26 at 11 46 05
(a malformed blob would appear - the tail and body of an empty tooltip)

Now, the tooltip won't 'show' if a label isn't provided, and a console warning will be logged. This is in addition to the existing error logged by onComponentRequiredPropUndefined()

How to replicate:
1. run branch locally
2. Edit the icon-only button story, deleting the 'aria-label' on one of the buttons
3. Hover over/tab onto the button.
4. Note that no tooltip appears

## Related issue
#1346 

## Checklist

### DIFF
--- a/packages/web-components/src/components/ic-tooltip/ic-tooltip.tsx
+++ b/packages/web-components/src/components/ic-tooltip/ic-tooltip.tsx
@@ -172,38 +172,42 @@ export class Tooltip {
   };
 
   private show = () => {
-    this.toolTip.setAttribute("data-show", "");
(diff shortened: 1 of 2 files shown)


GCHQ-Developer-465 (approved):
Have ran this locally and tested as mentioned in the PR. Awesome 👍

Issue #604, opened 2023-04-24
Updated textField to allow use of 'maxlength' and 'minlength' attributes


DBD324:
## Summary
Usual html input components allow you to set a maxlength and minlength, which sets the number of characters that can be entered into a field. The ic-text-field component should also provide this functioanlity. 

This differs from the existing prop max-length which is used to display a character count. The terminology used here is a bit confusing, as max-length may cause misunderstanding of its function, as it doesn't actually limit the number of characters that can be entered into the field. 

## 💬 Description
New properties should be created for maxlength and minlength that would work similarly to the html defintion. 

**Maxlength**
An integer value that specifies the maximum allowable characters that can be entered into the field. This actually limits additional characters being entered into the field. This value must be greater than or equal to minlength. If maxlength is exceeded then an error message should be shown stating 'Maximum input is X characters'. 

**minlength**
An integer value that specifies the minimum allowable characters that can be entered into the field.  This value must be less than or equal to maxlength. If minlength is not achieved then an error message should be shown stating 'Minimum input is X characters'. 

**Existing max-length prop**
This property currently is used to show the character count. The terminology here is incorrect, as it it doesn't actually affect the maximum allowable input characters, but instead sets the value of the character count. The user can still enter more characters than the character count into the field. 

Change docs to more accurately describe what this prop does. 

This ticket avoids a breaking change as it won't actually change name of prop, it will just update docs. Breaking change ticket will be raised (#1064) to change the prop name in next major release. 

## 💰 Use value
This attribute can be useful when creating forms as it makes sure the correct input length is adhered to. 

## Additional info
Tell us anything else useful to help us understand your suggestion.

gd2910:
Validation needs clarification, should it display warning/error when limit reached (as maxlength should never be able to be exceeded)? Old max-length should perhaps be changed to 'show-character-count' and be a boolean value, only applicable when new maxlength value set.

PR #1472, opened 2024-02-05
feat(web-components): add min/maxCharacters to ic-text-field

gd2910:
provide native functionailty to ic-text-field with min-characters and max-characters props, includes warnings for exceeding and aria labels

## Summary of the changes
provide native functionality to ic-text-field with min-characters and max-characters props, includes warnings for exceeding and aria labels. Validation status for each, spec tests added for new methods.

## Related issue
#604 

## Checklist


packages/web-components/src/components/ic-text-field/ic-text-field.tsx
```diff
@@ -372,13 +388,32 @@ export class TextField {
...
+        this.numChars >= this.maxCharacters ? true : false;
+    }
+  };
+
+  private getMinCharactersUnattained = (value: string) => {
+    this.numChars = value.length;
+
+    if (this.minCharacters > 0) {
+      this.minCharactersUnattained =
+        this.numChars < this.minCharacters ? true : false;
```
GCHQ-Developer-741, line 405:
```suggestion
        this.numChars < this.minCharacters;
```

packages/web-components/src/components/ic-text-field/ic-text-field.tsx
```diff
@@ -372,13 +388,32 @@ export class TextField {
     }
   };
 
+  private getmaxCharactersReached = (value: string) => {
+    this.numChars = value.length;
+
+    if (this.maxCharacters > 0) {
+      this.maxCharactersReached =
+        this.numChars >= this.maxCharacters ? true : false;
```
GCHQ-Developer-741, line 396:
```suggestion
        this.numChars >= this.maxCharacters;
```

GCHQ-Developer-741 (approved):

packages/web-components/src/components/ic-text-field/ic-text-field.tsx
GCHQ-Developer-847, line 1: Should the `maxCharacters` prop prevent the user from actually typing any more characters than the maximum? A bit like the [HTML text field](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/maxlength). It sounds like that might be the required behaviour looking at the ticket (not 100% sure though) - but then if that is the case, the maximum can't be exceeded so not sure when the error would appear

GCHQ-Developer-847 (changes requested):
A few comments :)

packages/web-components/src/components/ic-text-field/ic-text-field.tsx
GCHQ-Developer-847: As discussed, ignore the first part of this - it does work, sorry 🤦‍♀️ There's just the issue of when the error should appear (if at all)

packages/web-components/src/components/ic-text-field/ic-text-field.tsx
```diff
@@ -158,13 +161,23 @@ export class TextField {
   /**
    * The maximum number of characters that can be entered in the field.
    */
+  @Prop() maxCharacters: number = 0;
+
+  /**
+   * The label for maximum number of characters that can be entered in the field.
```
GCHQ-Developer-299, line 167: This comment for MaxLength doesn't seem to line up - how can a label be a number?

packages/web-components/src/components/ic-text-field/ic-text-field.tsx
```diff
@@ -56,6 +56,9 @@ export class TextField {
   @Element() el: HTMLIcTextFieldElement;
 
   @State() numChars: number = 0;
+  @State() maxCharactersReached: boolean = false;
+  @State() maxCharactersError: boolean = false;
```
GCHQ-Developer-741, line 60: I'm a bit confused why these two state values appear to be doing a similar thing. They also appear to be based off one another. Is it necessary for both of these to be considered `state`?

gd2910: I needed separate states to control the error state occurring and the maxlength being applied to the input, without these the error state is applied once the maxCharacters is reached, when we only want it to be applied once the user attempts to exceed the maxLength.

GCHQ-Developer-299 (dismissed):

packages/web-components/src/components/ic-text-field/ic-text-field.tsx
```diff
@@ -345,6 +362,9 @@ export class TextField {
   @Listen("keydown", {})
   handleKeyDown(ev: KeyboardEvent): void {
     this.icKeydown.emit({ event: ev });
+    this.maxCharactersReached
```
GCHQ-Developer-741, line 365:
```suggestion
    this.maxCharactersError = this.maxCharactersReached
```

DBD324 (approved):
From a functional perspective this seems to work well. Thanks

GCHQ-Developer-847 (approved):

PR #1535, opened 2024-02-19
fix(web-components): remove role on ic-alert when "announced" prop is false.

GCHQ-Developer-847:
## Summary of the changes
Tiny change to remove `role` attribute on ic-alert when `announced` prop is "false". This was causing an Accessibility Insights error with the alerts used in the guidance on the ICDS website because it was causing `role` to be set to "false" which is an invalid value for that attribute.

## Related issue
N/A

## Checklist

### DIFF
--- a/packages/web-components/src/components/ic-alert/ic-alert.stories.mdx
+++ b/packages/web-components/src/components/ic-alert/ic-alert.stories.mdx
@@ -154,3 +154,34 @@ import Button from "../ic-button/readme.md";
     </ic-alert>`}
   </Story>
 </Canvas>
+
+### Announced
+
+<Canvas>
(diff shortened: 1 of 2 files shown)


GCHQ-Developer-847:
> Out of curiosity, was it returning `role=undefined` if announced was false? Just thinking for a testing point of view, as long as it didn't appear correctly, this change would not break any tests where developers are using the role as a selector.

It was returning `role="false"` (you can see this with the alert on the [Angular page](https://design.sis.gov.uk/get-started/install-components/angular), for example)

GCHQ-Developer-299 (approved):

GCHQ-Developer-741 (approved):

Issue #1560, opened 2024-02-22
Chip: Accessibility Improvements for Icons


GCHQ-Developer-299:
## Summary 

- [x] Storybook: Add alt-text or aria-hidden to icons used in the chips in our web-component/react storybook stories
- [x] Guidance site: Add alt text to icons on Chip examples on guidance site. 

## 💬 Description 
Based on some screenreader testing of the [Chip Web Component stories](https://mi6.github.io/ic-ui-kit/branches/develop/web-components/?path=/story/web-components-chip--static-story) - when read out by voiceover all the icons read as 'image' because we've not provided alternative text. The same goes for the icons on the examples on our website. 

## 🧾 Guidance 
Chip guidance already tells us what to do for [Icons](https://design.sis.gov.uk/components/chip#icons) - we should follow our own words here :)

PR #1573, opened 2024-02-27
1560 chip accessibility improvements for icons

GCHQ-Developer-299:
## Summary of the changes
Add aria labels or aria-hidden to storybook stories for IcChip. This will be replicated in the guidance site in a separate pull request.

## Related issue
#1560 

## Checklist


GCHQ-Developer-530 (approved):

packages/react/src/stories/ic-chip.stories.mdx
jd239, line 1: The changes look good. I'm wondering if it is worth changing the svg to a `SlottedSVG` and making sure the icon is still read out correctly. An SVG with slot in TS would cause an issue.

jd239 (changes requested):
Just a small update. Other than that, looks good.

jd239 (approved):

Issue #1465, opened 2024-02-02
[ic-data-entity] Accessibility issue for data entity


GCHQ-Developer-530:
## Summary of the bug 
When running accessibility tests in Cypress for data entity, there's an error to do with the an aria-parent not being set.

It can also be replicated running accessibility checks against our site.

It seems to be an issue with the `role="cell"` we've added to a `div` for the `end-component` slot. If that's removed then there's no errors, but the cell role must have been applied for a reason so worth looking into whether we need to replace it with a different role or if any of the screen reader functionality goes missing when it's removed.

## 📸 Screenshots or code 
<img width="1034" alt="Screenshot 2024-02-02 at 13 11 27" src="https://github.com/mi6/ic-ui-kit/assets/111882034/ce1e49fe-6e64-478f-940c-da3dfdee2f7e">

## 🧐 Expected behaviour 
It should pass Cypress accessibility tests.

ASM995:
Please consider accessibility guidance if required.

PR #1629, opened 2024-03-04
fix(web-components): removed role of cell on the end-adornment in ic-data-entity

GCHQ-Developer-741:
## Summary of the changes
Removed cell role from ic-data-entity

## Related issue
#1465 

## Checklist

### DIFF
--- a/packages/web-components/src/components/ic-data-row/ic-data-row.tsx
+++ b/packages/web-components/src/components/ic-data-row/ic-data-row.tsx
@@ -48,9 +48,7 @@ export class DataRow {
   @Prop() value: string;
 
   disconnectedCallback(): void {
-    if (this.resizeObserver !== null) {
-      this.resizeObserver.disconnect();
-    }
+    this.resizeObserver?.disconnect();
   }
 
   componentWillLoad(): void {
@@ -61,9 +59,7 @@ export class DataRow {
 
   componentDidLoad(): void {
     checkResizeObserver(this.runResizeObserver);
-    if (this.hasEndComponent) {
-      this.labelEndComponent();
-    }
+    if (this.hasEndComponent) this.labelEndComponent();
   }
 
   private runResizeObserver = () => {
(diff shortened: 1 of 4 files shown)


GCHQ-Developer-299 (approved):

GCHQ-Developer-299 (changes requested):
looking at [Mozilla's docs for <dl>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl)
they show a <dl> containing multiple <dt> and <dd> pairs. Wheras our data-rows are a <dl> with a single <dd><dl> pair inside. This leads to a small oddity in VO where it says every entity is a definition list with two items.

Can the <dl> element be moved up to ic-data-entity.tsx?

packages/web-components/src/components/ic-data-row/ic-data-row.tsx
```diff
@@ -75,82 +71,75 @@ export class DataRow {
...
-      "slot[name=end-component]"
-    );
-
-    component.forEach((child: HTMLElement) =>
-      child.setAttribute("aria-label", "for " + this.label + " row")
-    );
+    this.el.shadowRoot
+      .querySelectorAll("slot[name=end-component]")
+      .forEach((child) =>
+        child.setAttribute("aria-label", "for " + this.label + " row")
```
jd239, line 109: Can use a template literal here

packages/web-components/src/components/ic-data-row/ic-data-row.tsx
```diff
@@ -75,82 +71,75 @@ export class DataRow {
...
+            {(isSlotUsed(el, "value") || value) && renderCellContent("value")}
           </div>
-          {this.hasEndComponent && (
-            <div class="end-component">
-              <div role="cell">
-                <slot name="end-component"></slot>
-              </div>
-            </div>
+          {hasEndComponent && (
+            <dd class="end-component">
```
jd239, line 140: Didn't realise the role was only on the last 'cell'. If the role or div was deleted, what affect would that make? I might be missing something. Just trying to understand if `role=cell` was not set on the other label and value divs, is it required for the end cell?

jd239 (changes requested):
A couple comments

packages/web-components/src/components/ic-data-row/ic-data-row.tsx
GCHQ-Developer-741: I don't believe it was required for any of the cells, which is why I removed it. role="cell" is used in table setups, rather than lists of data like this, so it was being used incorrectly

GCHQ-Developer-299 (approved):

GCHQ-Developer-530 (changes requested):
Can we add any missing a11y checks into the data entity cypress tests so that we can make sure this fix doesn't get reverted at any point?

GCHQ-Developer-530 (approved):

GCHQ-Developer-299 (approved):

jd239 (approved):

PR #1631, opened 2024-03-05
Minor storybook improvement

GCHQ-Developer-299:
## Summary of the changes
Added a media query to set text color to white in Storybook when Dark Mode is enabled. 
More could be done - for example we hard code svg 'fill' so icons in cards are still black in dark mode. But for now, I think this is a nice QoL improvement to dark mode storybook! Thoughts?

## Related issue
No issue, it was bugging me

## Checklist


GCHQ-Developer-530 (approved):
Looks like this creates an issue with white text in alert as well, but think it's like the ic-input issue you've already highlighted so should be done in a separate ticket 😊 definitely looks much better now

zxh722 (approved):

Issue #1380, opened 2024-01-15
Allow top navigation breakpoint size to be customized


ad9242:
## Summary 
Requests have been made to allow the larger size top navigation variant be displayed on smaller screen sizes.

## 💬 Description 
Currently, when the viewport width drops below 992px, certain content is moved into a menu accessed via the hamburger. The search bar also changes to a button that opens a panel below

![image](https://github.com/mi6/ic-ui-kit/assets/116102637/b069dc96-42d6-46d6-8e33-f67a066c5b7b)

![image](https://github.com/mi6/ic-ui-kit/assets/116102637/1f91fe82-14ee-4f5d-bd34-22f8a748c197)

In some cases, eg when not displaying app version or status information, there is a lot of unused space which means there would still be space available to keep the search bar, and show the navigation items\mega menu, even at smaller viewport widths.

Whilst allowing someone to specify an exact px threshold would be tricky to implement as is, a proposal is to add a prop which would enable users to set the size at which the switch occurs. Currently it happens when the viewport size drops to what is defined as a medium sized device (below 992px). So by adding a prop, this could be set to change at size small, medium, or large (**are these 3 sufficient??? Is large really required?? **). this could potentially be achieved with jsut a boolean prop of something like "breakAtMediumInsteadOfLarge" (better name required)

NOTE: may need to consider how css is applied, in particular where media queries are currently used. One option (untested), may be to apply a particular class in the resizeObserver callback, & apply the css that way.

## 💰 Use value 
This would allow greater flexibility for users to show their full content in cases where it would fix appropriately on a smaller viewport.

## 📝 Acceptance Criteria  
If relevant, describe in full detail the different interactions and edge cases that the component or patterns needs to fulfil.  

*Given* a top navigation exists
*When* content is displayed on viewport less than 992px wide
*Then* all content can be visible without switching to the flyout menu view

## 🧾 Guidance 
May need to add guidance around ensuring that content will fit\display correctly when setting the prop to smaller size

## Additional Info
An alternative option (simpler?) option has been discussed, where maybe top nav is made more aware of it's content and, say, would not break so early if there were no status\version elements in the top nav.

PR #1704, opened 2024-03-25
1380 top nav custom breakpoint

GCHQ-Developer-741:
## Summary of the changes
- Added a new prop, called `customMobileBreakpoint`, which allows users to switch to mobile mode at different screen sizes.
- Added new type for this prop, called `IcDeviceSizes`, which only allows our device size values to be set.
- Added class to be applied to host element when it is in mobile mode, since CSS media queries won't work with this prop.
- Refactored some code in a few components.
- Added a CSS attribute to ic-navigation-group to prevent it from wrapping its text when being included in ic-horizontal-scroll. Also added a cypress visual-regression test for this, with generated screenshot 

## Related issue
#1380 

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes


GCHQ-Developer-465 (approved):
Looks great! Good use of cypress testing too 👍

gd2910 (approved):

PR #1744, opened 2024-04-09
Add tab label to the detail emitted after the onIcTabSelect event

GCHQ-Developer-112:
## Summary of the changes
Add tab label to the detail emitted by the `onIcTabSelect` and `icTabSelect` events

## Related issue
https://github.com/mi6/ic-design-system/issues/751


gd2910 (dismissed):

ad9242 (changes requested):
looks good, but it would be nice to see an example in storybook, so developers can see the type of info that is emitted in the event (which could then maybe be added as an example to the site in another ticket)

ad9242 (approved):

GCHQ-Developer-299 (approved):

Issue #1468, opened 2024-02-05
ic-chip: Allow background colour to be set. 


DBD324:
## Summary 
Allow customisation of chip background. 
Provide appearance property to change foreground colour so that an accessible pairing can be achieved. 

## 💬 Description 
Chips are often used to denote different categories of information, so the ability to change the background colour would enable categories to be assigned colours. 

## 💰 Use value 
This makes the chip component more flexible so it can cater to more use cases. 

## 📝 Acceptance Criteria  
If relevant, describe in full detail the different interactions and edge cases that the component or patterns needs to fulfil.  

*Given* that a chip is being implemented 
*When* setting it properties
*Then* a custom colour can be selected,
*And* the foreground appearance should be automatically set to ensure good contrast. 

## ✏ Designs 
If there's a Figma design file (or other mock-up), include it here. 

## 🧾 Guidance 
If there's written guidance or documentation, include a link to it here.  

## Additional info 
Tell us anything else useful to help us understand your suggestion.

PR #1748, opened 2024-04-09
1468 ic chip allow background colour to be set

GCHQ-Developer-530:
## Summary of the changes
Add custom colour prop to chip to alter the background/border colour. 
Move IcColor type from badge types to global types. 
Add unit tests to up coverage.
Add cypress test.
Add missing pagination test snapshots.

## Related issue
#1468 

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes


packages/web-components/src/components/ic-chip/ic-chip.tsx
```diff
@@ -40,6 +42,12 @@ export class Chip {
    */
   @Prop() appearance?: IcChipAppearance;
 
+  /**
+   * The custom chip colour. This will only style the chip component if dismissible=false.
+   * Can be a hex value e.g. "#ff0000", RGB e.g. "rgb(255, 0, 0)", or RGBA e.g. "rgba(255, 0, 0, 1)".
+   */
+  @Prop() customColor?: IcColor = null;
```
GCHQ-Developer-112, line 49: Would be good to add a sentence on how this overrides the transparentBackground prop/attribute.

Also good to add the reverse to the transparentBackground prop

packages/web-components/src/components/ic-chip/ic-chip.tsx
```diff
@@ -40,6 +42,12 @@ export class Chip {
    */
   @Prop() appearance?: IcChipAppearance;
 
+  /**
+   * The custom chip colour. This will only style the chip component if dismissible=false.
```
GCHQ-Developer-112, line 46:
```suggestion
   * The custom chip colour. This prop/attribute will be applied to the chip component if `dismissible=false`.
```

packages/web-components/src/components/ic-chip/ic-chip.tsx
```diff
@@ -127,6 +139,31 @@ export class Chip {
     this.isHovered = false;
   };
 
+  private setChipColour = () => {
+    let colorRGBA = null;
```
GCHQ-Developer-741, line 143: This bit of code (143-150) is repeated in both ic-badge and ic-theme. Suggest making it a helper function that can be used in all three.

packages/web-components/src/components/ic-chip/ic-chip.tsx
GCHQ-Developer-530: I don't think it does override the transparentBackground prop. The transparentBackground prop only affects the outlined variant, and for the outlined variant the customColor prop only changes the border colour.

GCHQ-Developer-112: Ahh sorry, I misunderstood what the `transparentBackground `prop is used for

packages/web-components/src/components/ic-chip/ic-chip.tsx
```diff
@@ -127,6 +138,24 @@ export class Chip {
     this.isHovered = false;
   };
 
+  private setChipColour = () => {
+    const colorRGBA = convertToRGBA(this.customColor);
```
GCHQ-Developer-741, line 142: What happens if convertToRGBA returns null?

gd2910:
Looks great! There's a couple of pagination baseline screenshots in the commit that I'm not sure should be there?

GCHQ-Developer-530:
> Looks great! There's a couple of pagination baseline screenshots in the commit that I'm not sure should be there?

I think they were left out of a previous commit so when I run the action, they get added in as they're missing. I think that's okay though because we do need them

gd2910 (dismissed):

packages/web-components/src/components/ic-chip/ic-chip.tsx
```diff
@@ -40,6 +41,12 @@ export class Chip {
    */
   @Prop() appearance?: IcChipAppearance;
 
+  /**
+   * The custom chip colour. This prop will be applied to the chip component if `dismissible=false`.
```
ad9242, line 45:
```suggestion
   * The custom chip colour. This prop will be applied to the chip component if `dismissible` is set to `false`.
```

just to match other places we mention props set to values

packages/web-components/src/components/ic-chip/ic-chip.tsx
```diff
@@ -83,6 +90,10 @@ export class Chip {
   componentWillLoad(): void {
     removeDisabledFalse(this.disabled, this.el);
 
+    if (!this.dismissible && this.customColor !== null) {
+      this.setChipColour();
```
ad9242, line 94: this will only happen when the component first loads. should there be a @watch on the customColour prop to run the same logic if the value changes?

ad9242: if so, should also add a test to check it works when the prop is changed

gd2910 (dismissed):

GCHQ-Developer-741 (approved):

packages/web-components/src/components/ic-chip/ic-chip.tsx
```diff
@@ -127,6 +146,25 @@ export class Chip {
...
   };
 
+  private setChipColour = () => {
+    if (convertToRGBA(this.customColor) !== null) {
+      const colorRGBA = convertToRGBA(this.customColor);
+      const rgbaValue = `rgba(${colorRGBA.r.toString()}, ${colorRGBA.g.toString()}, ${colorRGBA.b.toString()}, ${colorRGBA.a.toString()})`;
+
+      const brightness =
+        (colorRGBA.r * 299 + colorRGBA.g * 587 + colorRGBA.b * 114) / 1000;
+      const foregroundColour =
```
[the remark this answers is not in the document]
GCHQ-Developer-530: I think it's used slightly differently in other components like badge as it doesn't set a colour, it sets it to the IcForegroundEnum. I think if we implement anything else similar and this is duplicated, then we could create a util method.

GCHQ-Developer-741 (approved):

GCHQ-Developer-299 (approved):

PR #1760, opened 2024-04-12
Accordion text select

ad9242:
## Summary of the changes
Fixes issue selecting text in accordions\accordion group. The issue occurred when the group had the prop `singleExpansion` set to `false` as the see/hide all button would receive focus.

The fix removes delegatesFocus from the accordion group component but this required changes to ic-dialog to ensure that elements are correctly focussed, when used there

## Related issue
#1174

### General 

### Testing


packages/web-components/src/components/ic-dialog/ic-dialog.tsx
```diff
@@ -440,7 +443,10 @@ export class Dialog {
 
     let nextFocusEl = this.getNextFocusEl(this.focusedElementIndex);
 
-    const isHidden = getComputedStyle(nextFocusEl).visibility === "hidden";
+    const isHidden =
+      getComputedStyle(nextFocusEl).visibility === "hidden" ||
+      (nextFocusEl.tagName === this.IC_ACCORDION_GROUP &&
+        nextFocusEl.hasAttribute("single-expansion"));
```
jd239, line 449: The overall fix works as expected but this confused me for accordion. Is the check for the single-expansion so that it skips the see all/hide all button as that is not rendered when a single expansion?

jd239 (commented):
Looks good. Just a single comment to improve my understanding rather than a required update

packages/web-components/src/components/ic-dialog/ic-dialog.tsx
ad9242: Correct. When single expansion, there is no see\hide button so the first accordian element needs to be focussed instead (as the current focussed element is the accordian group which no longer delegates focus)

GCHQ-Developer-112:
Small comment/typo - please could change "accordian" to "accordion" in your commit messages

packages/web-components/src/components/ic-dialog/ic-dialog.tsx
jd239: Was trying to figure out why `ic-accordion-group` was registering as an interactive element even though there is no button but I can see it in the slottedInterativeElement array.

jd239 (approved):

packages/web-components/src/components.d.ts
```diff
@@ -104,6 +104,10 @@ export namespace Components {
           * The header for the accordion group.
          */
         "groupTitle": string;
+        /**
+          * Sets the focus on first focusable element in the accordian group. If the "See/Hide all" button is present, it will be focused. Otherwise, the first accordion will be focused.
```
GCHQ-Developer-299, line 108: Got a typo in that first 'accordian'

GCHQ-Developer-299 (approved):

GCHQ-Developer-465 (approved):

Issue #1768, opened 2024-04-16
[ic-text-field] Cypress migration


gd2910:
## Summary
Migrate current E2E and snapshot tests to Cypress, remove old tests and redundant snapshots.

Ensure current tests adequately cover the functionality of the component and add more where needed.

PR #1807, opened 2024-04-24
test(react): icTextField Cypress Tests

MI6-286:
icTextField Cypress Tests

#1768

## Summary of the changes
Completed migration of e2e tests to Cypress for IcTextField and removed old e2e tests.

## Related issue
#1768


packages/react/src/component-tests/IcTextField/IcTextFieldTestData.tsx
```diff
@@ -31,3 +34,362 @@ export const Uncontrolled = () => {
...
+  (): ReactElement => (
+    <div style={style}>
+      <IcTextField
+        maxLength={25}
+        label="What is your favourite coffee?"
+        required
+        placeholder="Please enter…"
+        helperText="Such as Arabica, Robusta or Liberica"
+      >
+        <SlottedSVG
```
GCHQ-Developer-299, line 54: Not at all necessary for approval on this PR, but I think you can extract the SlottedSVG into it's own variable and paste than it for each example in the file, to cut down on duplicated markup :)

GCHQ-Developer-299: Check out an example in IcSideNavigationTestData - the variable ReusableSlottedIcon

packages/react/src/component-tests/IcTextField/IcTextFieldTestData.tsx
```diff
@@ -31,3 +47,290 @@ export const Uncontrolled = () => {
...
+      helper-text="Such as Arabica, Robusta or Liberica"
+    >
+      <SlottedSVG
+        slot="icon"
+        xmlns="http://www.w3.org/2000/svg"
+        height="24px"
+        viewBox="0 0 24 24"
+        width="24px"
+        fill="#000000"
+        size="small"
```
ad9242, line 330: does this `size` on slottedsvg do anything? i don't think there is a "size" attribute. can this be removed & then just use the ReusableSlottedIcon?

ad9242 (changes requested):

packages/react/src/component-tests/IcTextField/IcTextFieldTestData.tsx
MI6-286: As agreed, removed the icon from CustomSizeTextField

GCHQ-Developer-530 (approved):

ad9242 (approved):

PR #1821, opened 2024-05-01
Fix audit errors

GCHQ-Developer-530:
## Summary of the changes
Fix audit errors in each package.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4675,9 +4675,9 @@
       "dev": true
     },
     "node_modules/ejs": {
-      "version": "3.1.9",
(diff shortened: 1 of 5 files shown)


GCHQ-Developer-847 (approved):

ad9242 (approved):

GCHQ-Developer-465:
Are there supposed to be 5 commits on this?

GCHQ-Developer-530:
> Are there supposed to be 5 commits on this?

Yeah we need each audit fix to be in the correct scope so all 5 are needed in this case

PR #1958, opened 2024-06-06
1153 server side pagination

GCHQ-Developer-741:
## Summary of the changes
- Added a new slot to ic-data-table, called pagination-bar, where a user can use their own pagination-bar to control pagination when interacting with backend APIs.
- Added react and web-component stories to show the new slot
- Did some minor refactoring and added a check to ic-pagination-bar, to prevent infinite state loops when the same array contents are being received by the pagination-bar, triggering the watch handler
- Added slot definition to canary-docs

## Related issue
#1153 

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes


packages/canary-web-components/src/components/ic-data-table/ic-data-table.tsx
```diff
@@ -795,23 +801,27 @@ export class DataTable {
             }
           ></ic-loading-indicator>
         )}
-        {showPagination && (
+        {(showPagination || isSlotUsed(this.el, "pagination-bar")) && (
```
ad9242, line 804: It would be good to check if there are any issues with not having any slot content initially, but then adding it later, as there have been issues with other components with this approach. A bit of an edge case, but I'm thinking of maybe you have an empty table initially & only put content in the bar slot once data has loaded

ad9242:
Should it be made clearer that the `paginationBarOptions` prop on data table is only used by the built in pagination bar and not if you use the slot?

ad9242 (dismissed):

GCHQ-Developer-299 (approved):

ad9242 (approved):

Issue #1970, opened 2024-06-10
Add props playground to each component and test all combinations


MI6-255:
## Summary 
Add props playground to each component and test all combinations

## 💬 Description 
This issue will:

1. Add a props playground to each component
2. Test all combinations of the component to make sure they work together, either fixing them quickly where they do not, or opening a ticket

Broken up per component, assign yourself to the relevant task:

- [x] Accordion @GCHQ-Developer-530 
- [x] Alert (Already has playground that has been checked in PR checklist review)
- [x] Back to top @ad9242 #2026 
- [x] Badge @GCHQ-Developer-530 
- [x] Breadcrumb @GCHQ-Developer-530 
- [ ] Button @MI6-255 
- [x] Card @GCHQ-Developer-530 
- [x] Checkbox @ad9242 
- [x] Chip @GCHQ-Developer-530 
- [x] Classification banner @GCHQ-Developer-530 
- [x] Data entity @GCHQ-Developer-530 
- [x] Data table @GCHQ-Developer-530 
- [x] Date input @GCHQ-Developer-530 
- [x] Date picker @GCHQ-Developer-530 
- [x] Dialog @GCHQ-Developer-530 
- [x] Empty state @GCHQ-Developer-530 
- [x] Footer @MI6-255 
- [x] Hero @MI6-255 
- [x] Link @GCHQ-Developer-530 
- [x] Loading indicator @GCHQ-Developer-847 
- [x] Multi select @GCHQ-Developer-847 
- [x] Page header @MI6-255 
- [x] #1988 @gd2910 
- [x] Popover menu @GCHQ-Developer-530 
- [x] Radio button @GCHQ-Developer-530 
- [x] Search bar @GCHQ-Developer-530 done in remaining checklist PR
- [x] Section container @GCHQ-Developer-299 
- [x] Select - has playground, check for point 2. @GCHQ-Developer-299  ('Size' toggle is a text field)
- [x] Side navigation @GCHQ-Developer-299 
- [x] Skeleton @GCHQ-Developer-530 
- [x] Status tag @GCHQ-Developer-530 
- [x] Stepper @gd2910 #2039 #1609 
- [x] Switch (Already has playground that has been checked in PR checklist review)
- [x] Tabs @GCHQ-Developer-530 
- [x] Text field @GCHQ-Developer-741 
- [x] Toast @GCHQ-Developer-530 
- [x] Toggle button @GCHQ-Developer-530 
[...]

PR #1989, opened 2024-06-13
1970 playground select, side nav and section container

GCHQ-Developer-299:
## Summary of the changes
Add Storybook playgrounds for Select, SideNav and Section Container
Small fix to Section Container for when you change 'aligned' prop to null after first render (as what happens in the storybook)

## Related issue
#1970

### DIFF
--- a/packages/react/src/stories/ic-side-navigation.stories.mdx
+++ b/packages/react/src/stories/ic-side-navigation.stories.mdx
@@ -26,6 +26,17 @@ import { NavLink, MemoryRouter, Route, Routes } from "react-router-dom";
 
 <Description markdown={readme + NavigationItem + NavigationGroup} />
 
+export const defaultArgs = {
+  appTitle: "Test App Title",
+  version: "v1.0.0",
+  status: "BETA",
+  href: "https://www.google.com",
+  collapsedIconLabels: false,
+  navOneLabel: "Navigation One",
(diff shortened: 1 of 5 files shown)


packages/react/src/stories/ic-side-navigation.stories.mdx
```diff
@@ -630,6 +641,106 @@ export const Controlled = () => {
   </Story>
 </Canvas>
 
+### Playground
+
+<Canvas>
+  <Story
+    name="Playground"
+    args={defaultArgs}
+    argTypes={{}}
```
GCHQ-Developer-530, line 650: Can this line be removed if the argTypes are empty?

GCHQ-Developer-530 (approved):

ad9242:
found a slight issue in the select playgrounds: if you set loading to true, then back to false, it doesn't clear & you can no longer see the options

GCHQ-Developer-299:
> found a slight issue in the select playgrounds: if you set loading to true, then back to false, it doesn't clear & you can no longer see the options

I've found the issue I think - the loadingHandler function doesn't consider leaving loading...
```js
  @Watch("loading")
  loadingHandler(newValue: boolean): void {
    newValue && this.triggerLoading();
  }
```

gd2910 (approved):

PR #1997, opened 2024-06-14
docs(react): add Hero playground to React

MI6-255:
## Summary of the changes
add Hero playground to React to test all combinations of props

## Related issue
#1970 

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes

### DIFF
--- a/packages/react/src/stories/ic-hero.stories.mdx
+++ b/packages/react/src/stories/ic-hero.stories.mdx
@@ -55,6 +55,63 @@ export const ControlledReset = () => {
   );
 };
 
+### Playground
+
+export const defaultArgs = {
+  aligned: "left",
+  backgroundImage: image,
+  contentAligned: "left",
+  disableBackgroundParallax: false,
+  heading: "Hero heading",
+  secondaryHeading: "Secondary Heading",
+  secondarySubheading: "This is a secondary description.",
+  size: "default",
+  subheading: "Hero description. This is a Hero component, it should be used as a page heading.",
+};
+
+<Canvas>
+  <Story
+    args={defaultArgs}
+    name="Playground"
+    parameters={{ layout: "fullscreen" }}
+    argTypes={{
+      aligned: {
+        options: ["center", "full-width", "left"],
+        control: { type: "select" },
+      },
+      contentAligned: {
+        options: ["center", "left"],
+        control: { type: "select" },
+      },
+      size: {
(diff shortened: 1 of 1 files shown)


packages/react/src/stories/ic-hero.stories.mdx
```diff
@@ -55,6 +55,81 @@ export const ControlledReset = () => {
...
+      contentAligned: {
+        options: ["center", "left"],
+        control: { type: "select" },
+      },
+      disableBackgroundParallax: {
+        control: { type: "boolean" },
+      },
+      heading: {
+        control: { type: "text" },
+      },
```
GCHQ-Developer-530, line 94: For any text ones I don't think you have to add a control type, it infers it from the default arg being a string. Same for boolean ones too

GCHQ-Developer-530 (approved):

ad9242:
when setting `contentAligned` to `center`, the secondary heading does not center align - is that correct?

image

packages/react/src/stories/ic-hero.stories.mdx
```diff
@@ -55,6 +55,63 @@ export const ControlledReset = () => {
...
+  size: "default",
+  subheading: "Hero description. This is a Hero component, it should be used as a page heading.",
+};
+
+<Canvas>
+  <Story
+    args={defaultArgs}
+    name="Playground"
+    parameters={{ layout: "fullscreen" }}
+    argTypes={{
```
ad9242, line 77: could we make the background image be a select (with an empty value in the list too ) rather than a string? that way you could switch between the image & none

ad9242 (changes requested):

packages/react/src/stories/ic-hero.stories.mdx
MI6-255: I could do, but I thought it'd be more flexible if I left it as a string, as it defaults to an image we have, but then it also can be changed to the URL or any other image. So the string allows the flexibility of the current image, no image, and any other image. Whereas a select wouldn't allow any other image. Let me know if you still think we go down the route of using a switch though

MI6-255:
> when setting `contentAligned` to `center`, the secondary heading does not center align - is that correct?
> 
> image

Good spot, I'll open a ticket for it!

ad9242 (approved):

PR #2014, opened 2024-06-17
add Page Header playground to React so all props can be tested

MI6-255:
## Summary of the changes
add Page Header playground to React so all props can be tested

## Related issue
#1970

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes

### DIFF
--- a/packages/react/src/stories/ic-page-header.stories.mdx
+++ b/packages/react/src/stories/ic-page-header.stories.mdx
@@ -36,6 +36,94 @@ import readme from "../../../web-components/src/components/ic-page-header/readme
 
 <Description markdown={readme} />
 
+### Playground
+
+export const defaultArgs = {
+  aligned: "left",
+  border: true,
+  heading: "Page Header Title",
+  reverseOrder: false,
+  size: "default",
+  sticky: false,
+  stickyDesktopOnly: false,
+  subheading: "Page Header Subtitle",
+};
+
+<Canvas>
+  <Story
+    args={defaultArgs}
+    name="Playground"
+    parameters={{ layout: "fullscreen" }}
+    argTypes={{
+      aligned: {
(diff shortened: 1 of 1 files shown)


packages/react/src/stories/ic-page-header.stories.mdx
```diff
@@ -36,6 +36,52 @@ import readme from "../../../web-components/src/components/ic-page-header/readme
...
+          aligned={args.aligned}
+          border={args.border}
+          heading={args.heading}
+          reverseOrder={args.reverseOrder}
+          size={args.size}
+          sticky={args.sticky}
+          stickyDesktopOnly={args.stickyDesktopOnly}
+          subheading={args.subheading}
+        >
+          <IcStatusTag slot="heading-adornment" label="Beta" />
```
GCHQ-Developer-530, line 79: Would be good to have some buttons in the action slot so that we can see the reverseOrder prop working

MI6-255: Agreed, I'll add

ad9242:
do you think it should have some page content, so that it can be scrolled & see the sticky working?

MI6-255:
> do you think it should have some page content, so that it can be scrolled & see the sticky working?

Agreed, added

packages/react/src/stories/ic-page-header.stories.mdx
MI6-255: Added

MI6-255:
Bug opened for reverseOrder https://github.com/mi6/ic-ui-kit/issues/2047

GCHQ-Developer-299 (approved):

ad9242 (approved):

Issue #1877, opened 2024-05-17
[IcDatePicker] - changing startOfWeek prop from initial value has no effect


ad9242:
## Summary of the bug 
If changing the startOfWeek prop from an initial value, it has no effect on the display of the picker

## 🪜 How to reproduce 
With the following [stackblitz example](https://stackblitz.com/edit/vitejs-vite-gtn2tc?file=src%2FApp.jsx)
1. Open the date picker
2. See that Sunday is the start of the week in the picker
Click on the Button to change the startOfWeek
4. Open the date picker & see that sunday is stil the first day of the week - it's should now be friday

## 🧐 Expected behaviour 
The first day of the week in the picker should reflect the new value set

## 📝 Acceptance Criteria  

*Given* a datepicker has an initial startOfWeek prop value set
*When*  changing the prop to a new value
*Then* the start of week should change in the picker

PR #2032, opened 2024-06-19
1970 canary date component playgrounds

GCHQ-Developer-530:
## Summary of the changes
Add playground stories for date input and date picker

Ticket already created for startOfWeek not updating - https://github.com/mi6/ic-ui-kit/issues/1877

## Related issue
Part of #1970

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes

### DIFF
--- a/packages/canary-react/src/stories/ic-date-input.stories.mdx
+++ b/packages/canary-react/src/stories/ic-date-input.stories.mdx
@@ -6,9 +6,10 @@ import {
   Description,
 } from "@storybook/addon-docs";
 import { IcDateInput } from "../components";
-import { IcButton, IcTypography } from "@ukic/react";
+import { IcButton, IcTypography, IcLink } from "@ukic/react";
 import readme from "../../../canary-web-components/src/components/ic-date-input/readme.md";
 import { useState } from "react";
+import { useArgs } from '@storybook/client-api';
 
 <Meta title="React Components/Date Input" component={IcDateInput} />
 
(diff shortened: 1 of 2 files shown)


GCHQ-Developer-299 (approved):

ad9242 (changes requested):
i think it would be quite nice to make value a string, then we could test with "-" separator too

ad9242:
There seems to be a problem where, if you have a value set, changing the date-format clears the field. I'd expect it to update the value, but that should be a new ticket as it doesn't support it at the moment (would need changes to input & picker i think)

GCHQ-Developer-530:
> There seems to be a problem where, if you have a value set, changing the date-format clears the field. I'd expect it to update the value, but that should be a new ticket as it doesn't support it at the moment (would need changes to input & picker i think)

https://github.com/mi6/ic-ui-kit/issues/2056 created

ad9242 (changes requested):
I think the blanking of the value may be a bigger issue, than just changing the date format. it seems to be that, if you have entered a value in the picker\input and then change a storybook control eg, disabled, then it blanks the value. I think this is because it re-uses the value from the storybook control.

you can keep them in sync with https://storybook.js.org/docs/addons/addons-api#useargs 
see checkbox playground for an example

GCHQ-Developer-530:
> I think the blanking of the value may be a bigger issue, than just changing the date format. it seems to be that, if you have entered a value in the picker\input and then change a storybook control eg, disabled, then it blanks the value. I think this is because it re-uses the value from the storybook control.
> 
> you can keep them in sync with https://storybook.js.org/docs/addons/addons-api#useargs see checkbox playground for an example

Ah brilliant, thank you. I've updated it

packages/canary-react/src/stories/ic-date-input.stories.mdx
```diff
@@ -607,3 +608,86 @@ The date input can be cleared by setting the the value attribute to one of the f
...
+          openAtDate={args.openAtDate}
+          required={args.required}
+          showDaysOutsideMonth={args.showDaysOutsideMonth}
+          showPickerClearButton={args.showPickerClearButton}
+          showPickerTodayButton={args.showPickerTodayButton}
+          size={args.size}
+          startOfWeek={args.startOfWeek}
+          validationStatus={args.validationStatus}
+          validationText={args.validationText}
+          value={value}
```
ad9242, line 689: think you need to add an onIcChange event that calls updateArgs with the new value, to keep the control in sync:

```
onIcCheck={(ev) => updateArgs({ value: ev.detail.value })}
```

ad9242 (changes requested):

packages/canary-react/src/stories/ic-date-picker.stories.mdx
```diff
@@ -506,4 +508,109 @@ const DatePickerWithIcChange = () => {
...
+    }}
+    name="Playground example"
+  >
+    {(args) => {
+      const [{ value }, updateArgs] = useArgs();
+
+      const updateDate = (ev) => {
+        const date = ev.detail.value;
+        const formattedDate = `${date.getDate()}-${date.getMonth() + 1}-${date.getFullYear()}`;
+        updateArgs({ value: formattedDate});
```
ad9242, line 584: sorry, could this also handle setting an empty string when the value is cleared?

ad9242 (changes requested):

ad9242 (approved):

GCHQ-Developer-847 (approved):

PR #2036, opened 2024-06-20
1970 more component playgrounds

GCHQ-Developer-530:
## Summary of the changes
Add playground story for link and skeleton

## Related issue
Part of #1970 

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes

### DIFF
--- a/packages/react/src/stories/ic-link.stories.mdx
+++ b/packages/react/src/stories/ic-link.stories.mdx
@@ -98,3 +98,45 @@ export const DailyTippers = () => (
     </Routes>
   </Story>
 </Canvas>
+
+### Playground
+
+export const defaultArgs = {
+  appearance: "default",
+  href: "/",
+  target: "",
+  download: false,
+};
+
+<Canvas>
+  <Story
+    args={defaultArgs}
+    parameters={{ loki: { skip: true } }}
+    argTypes={{
+      appearance: {
+        options: ["default", "light", "dark"],
+        control: {
+          type: "radio",
+        },
+      },
+      target: {
+        options: ["_blank", "_self", "_parent", "_top", ""],
+        control: {
+          type: "radio",
+        },
+      },
+    }}
+    name="Playground"
+  >
+    {(args) =>(
+        <div style={{ backgroundColor: args.appearance == "light" ? "black" : ""}}>
+          <IcLink
+            href={args.href}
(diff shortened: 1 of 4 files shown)


ad9242 (changes requested):
seem to be missing some props on link: `hreflang`, `rel`, `referrerpolicy`, `download`. would be good to add them, just to see how they are passed down to the native `<a>` tag under the covers

GCHQ-Developer-530:
> seem to be missing some props on link: `hreflang`, `rel`, `referrerpolicy`, `download`. would be good to add them, just to see how they are passed down to the native `<a>` tag under the covers

I had considered these, when I added them they didn't change anything visually and especially download seemed like even the prop just being there meant that it was being applied (but I tested it and it works that way on the native `<a>` tag too) so couldn't test href or target if the link is downloaded even when download="false".

ad9242:
> > seem to be missing some props on link: `hreflang`, `rel`, `referrerpolicy`, `download`. would be good to add them, just to see how they are passed down to the native `<a>` tag under the covers
> 
> I had considered these, when I added them they didn't change anything visually and especially download seemed like even the prop just being there meant that it was being applied (but I tested it and it seemed to work that way on the native `<a>` tag too) so couldn't test href or target if the link is downloaded even when download="false".

I think it would be good to add the download one. Looks like we could add 
`download={args.download === true}`
to only apply it, if true

GCHQ-Developer-530:
> > > seem to be missing some props on link: `hreflang`, `rel`, `referrerpolicy`, `download`. would be good to add them, just to see how they are passed down to the native `<a>` tag under the covers
> > 
> > 
> > I had considered these, when I added them they didn't change anything visually and especially download seemed like even the prop just being there meant that it was being applied (but I tested it and it seemed to work that way on the native `<a>` tag too) so couldn't test href or target if the link is downloaded even when download="false".
> 
> I think it would be good to add the download one. Looks like we could add `download={args.download === true}` to only apply it, if true

I've fixed this for React but it still doesn't work for web components so I've created a ticket to investigate (https://github.com/mi6/ic-ui-kit/issues/2082) as I can't work out why it would work for one and not the other

GCHQ-Developer-847 (approved):

ad9242 (approved):

Issue #2039, opened 2024-06-20
[ic-stepper, ic-step] step-connector not resizing on `aligned` prop change


gd2910:
## Summary of the bug 
When ic-stepper is aligned `left` and then changed to aligned `full-width` the step-connector remains the same size.

## 🪜 How to reproduce 
Tell us the steps to reproduce the problem: 
1. Go to page: IcStepper playground
2. Click on: set to left aligned then full-width
3. See the error 

## 📸 Screenshots or code 
Left aligned
![image](https://github.com/mi6/ic-ui-kit/assets/122624946/82aeab60-3bdf-4c50-a3bf-08d046d8195e)

Full-width aligned
![image](https://github.com/mi6/ic-ui-kit/assets/122624946/a01ddb1e-9df7-47ff-a70a-e7a74b2b29d9)

## 🧐 Expected behaviour 
The step-connect should remain at 100% width after the change.

Issue #1609, opened 2024-02-29
Stepper - Missing Properties 


ASM995:
## Summary 
Add size and appearance prop to stepper component.
## 💬 Description 
Add size property including options for 'Default' 'Small' and 'Large', where 'default' is the default.

Add appearance property including options for 'Default', 'Light' and 'Dark'.

Check component to confirm which prop options are required.

## 💰 User value 
Provides consistent functionality across all ICDS components. 

## 📚 User Stories 
If relevant, describe the high-level functionality as user stories. 

*As an* ICDS user:  
*I need* 
*So that*  

## 📝 Acceptance Criteria  
If relevant, describe in full detail the different interactions and edge cases that the component or patterns needs to fulfil. 

*Given* 
*When*  
*Then* 

## ✏ Designs 
If there's a Figma design file (or other mock-up), include it here. 

## 🧾 Guidance 
If there's written guidance or documentation, include a link to it here.  

## Additional info 
Tell us anything else useful to help us understand your suggestion.

gd2910:
Design will have to take into consideration for `size` prop the two variants, i.e. three sizes for each variant?
There are no designs as yet for light/dark appearance.

gd2910:
![image](https://github.com/mi6/ic-ui-kit/assets/122624946/00dbe028-dc5d-4eda-874c-0987376f59d0)
The appearance is changed through checking if the parent (ic-stepper) has the prop in the DOM, we may want to change this to a @Watch() to ensure the appearance can change post render.

PR #2041, opened 2024-06-20
test(react): icStepper playground

gd2910:
IcStepper playground added for testing

Includes fix for when hideStepInfo is dynamically set and unset the step info would not re-appear

Ticket #2039 created and #1609 commented as a result of playground testing.

## Summary of the changes
React IcStepper playground added and a watch decorator for hideStepInfo

## Related issue
#1970


GCHQ-Developer-530:
https://github.com/mi6/ic-ui-kit/assets/111882034/b7ca21d5-db56-4c16-b91b-d58cc5bcf622

Getting some weird behaviour when toggling between compact and default, it flashes up for a second on desktop then disappears. And for mobile, it doesn't appear at all. Would be worth creating an issue for this unless you know what's causing it

packages/web-components/src/components/ic-stepper/ic-stepper.tsx
```diff
@@ -231,12 +239,26 @@ export class Stepper {
...
 
+  private setHideStepInfo = (): void => {
+    this.steps.forEach((step) => {
+      const stepTitleArea = step.shadowRoot.querySelector(
+        ".step > .step-title-area"
+      );
+
+      if (this.hideStepInfo && stepTitleArea !== null) {
+        stepTitleArea.classList?.add(this.visuallyHidden);
+      } else {
```
ad9242, line 256: this seems to be causing an error in the tests as `hideStepInfo` is `true`, but `stepTitleArea` is `null`

packages/web-components/src/components/ic-stepper/ic-stepper.tsx
ad9242, line 1: Are the calls to `this.getChildren()  required in `overrideVariant`, as there is now a watch for when variant changes?

ad9242 (changes requested):

GCHQ-Developer-299 (approved):

ad9242 (approved):

PR #1992, opened 2024-06-13
1970 text field playground

GCHQ-Developer-741:
## Summary of the changes
- Updated state when maxCharacters is set to zero after the validation has been shown, to allow users to enter text in this instance. Previously the text-field would not allow new characters due to the max being zero.
- Added child components to mutation observer for ic-text-field and ic-input-component-container, which will force an update if a slotted icon is added as a child, meaning that the icon can now be conditionally rendered

## Related issue
#1970 

### General 

- [x] Changes to docs package checked and committed.
- [x] All acceptance criteria reviewed and met. 

### Testing

- [x] Relevant unit tests and visual regression tests added. 
- [x] Visual testing against Figma component specification completed. 

### Accessibility 

- [x] Accessibility Insights FastPass performed.
- [x] A11y unit test added and yields no issues.
- [x] A11y plug-in on Storybook yields no issues. 
- [x] Manual screen reader testing performed using NVDA and VoiceOver. 
- [x] Manual keyboard testing for keyboard controls and logical focus order. 
- [x] Correct roles used and ARIA attributes used correctly where required. 
- [x] Logical heading structure is maintained, and the HTML elements used for headings can be changed to fit within the wider page structure. 

### Resize/zoom behaviour 

- [x] Page can be zoomed to 400% with no loss of content. 
- [x] Screen magnifier used with no issues. 
- [x] Text resized to 200% with no loss of content.
- [x] Text spacing increased as per the [WCAG 1.4.12 success criterion](https://www.w3.org/TR/WCAG21/#text-spacing) with no loss of content.

### System modes

- [x] Browser setting 'prefers reduced motion' tested. No animations or motion visible whilst this setting is on. 
- [x] Windows High Contrast mode tested with no loss of content.
- [x] System light and dark mode tested with no loss of content.
- [x] Browser support tested (Chrome, Safari, Firefox and Edge). 

### Testing content extremes

- [x] Min/max content examples tested with no loss of content or overflow. 
- [x] All prop combinations work without issue. 
- [x] Tested for FOUC (Flash of Unstyled Content) in both SSR (Server-Side Rendering) and SSG (Static Site Generation) settings.
- [x] Controlled and uncontrolled input components tested.

PR #2071, opened 2024-06-27
2045 add the ability to display an icon in a select

GCHQ-Developer-530:
## Summary of the changes
Add icon slot to select, update unit tests and cypress tests and add icon slot to storybook playground for default and searchable select

## Related issue
#2045 

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes


ad9242 (changes requested):
It seems the icon doesn't get shown if you have the slot empty initially & then add an icon. it can be replicated by setting it to `false` initially in the playground

ad9242:
should the icon be visible in readonly mode (if there is a value set) - to match TextField behaviour?

ad9242:
should the icon be a slightly greyer version until a value is entered? To match textfield?

GCHQ-Developer-741:
> It seems the icon doesn't get shown if you have the slot empty initially & then add an icon. it can be replicated by setting it to `false` initially in the playground

I faced a similar issue in https://github.com/mi6/ic-ui-kit/pull/1992 whilst working on text-field. Could help with this

packages/web-components/src/components/ic-select/ic-select.tsx
```diff
@@ -990,6 +993,15 @@ export class Select {
     }
   };
 
+  private hasLeftIconSlot(): boolean {
+    const iconEl = this.el.querySelector(`[slot="icon"]`);
+    return iconEl !== null;
+  }
+
+  private hasValue(): boolean {
+    return this.value !== null && this.value !== "" && this.value !== undefined;
```
GCHQ-Developer-741, line 1002:
```suggestion
    return !!this.value;
```

I feel like this could even be moved from a helper function, all that would be needed would be `!this.value` and `!!this.value`

ad9242 (approved):

GCHQ-Developer-847 (approved):

Issue #1064, opened 2023-09-05
Rename max-length prop on text fields to something more understandable


DBD324:
## Summary 
The existing prop max-length is used to display a character count. The terminology used here is a bit confusing, as max-length may cause misunderstanding of its function when compared with maxLength, as it doesn't actually limit the number of characters that can be entered into the field.

## 💬 Description 

**Existing max-length prop**
This property currently is used to show the character count. The terminology here is incorrect, as it it doesn't actually affect the maximum allowable input characters, but instead sets the value of the character count. The user can still enter more characters than the character count into the field.

Change naming to more accurately describe what this prop does. This is a breaking change as users may already be using this prop name in their code. 

## 💰 Use value 
Makes it easier to understand the different properties available on text field. 

## 📝 Acceptance Criteria  
If relevant, describe in full detail the different interactions and edge cases that the component or patterns needs to fulfil.  

*Given* 
*When*  
*Then* 

## ✏ Designs 
If there's a Figma design file (or other mock-up), include it here. 

## 🧾 Guidance 
If there's written guidance or documentation, include a link to it here.  

## Additional info 
Tell us anything else useful to help us understand your suggestion.

DBD324:
Relates to #604

MI6-255:
Scrap max-length and have a prop to show max characters. Design to clarify if this is okay

GCHQ-Developer-094:
According to site docs we have both properties max-characters and max-length, with the outcome of previous ticket 604 simply stating to just make the description clearer for max-length in the documentation. 

I'd suggest scrap max-length, use max-characters and just add another boolean property something along the lines of hideCharCount.

PR #2113, opened 2024-07-05
Feature/1064 refactor max length prop

GCHQ-Developer-847:
## Summary of the changes
Removed `maxLength` prop from text field and made character count display when the `maxCharacters` prop is used instead. Added `hideCharCount` prop to provide the option to visually hide the character count. Updated tests, including removing `maxLength` usage / replacing with `maxCharacters`.

The character count information for screen readers still remains when the visual character count is hidden. I felt that this would be a good idea to make it more obvious to screen reader users that their input is being limited. 

I have also updated the character count screen reader information to read out how many characters are remaining as originally the aria-live region was just reading out "10 slash 25", for example; now it would say "15 characters remaining", which I felt was more user-friendly and also matches the [GOV.UK](https://design-system.service.gov.uk/components/character-count/) behaviour.

## Related issue
#1064 

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes


packages/react/src/stories/ic-text-field.stories.mdx
```diff
@@ -471,12 +447,18 @@ export const Uncontrolled = () => {
     parameters={{ loki: { skip: true } }}
   >
     <IcTextField
-      max-characters={2}
+      maxCharacters={2}
```
GCHQ-Developer-299, line 450: I'm not sure if the warning behaviour is correct - it's displaying when i'm at the limit of characters, i assumed it'd only happen when i'm *over* the limit
Screenshot 2024-07-05 at 15 24 50

GCHQ-Developer-299 (dismissed):

GCHQ-Developer-299 (approved):

ad9242:
Should the validation\max length check be triggered on first load. For example if you have a value set that exceeds the maxCharacters? 

I have a long `value` set which exceeds the limit.

Initially this displays as follows:
image

After pressing a key in the input field, the warning is applied:
image


The old behaviour appears to be to trigger the error on load:
image

ad9242:
there no longer seems to be a visual test (with baseline image) where the maxCharacters is exceeded. I think it would be good to have one, just to ensure styling doesn't change

GCHQ-Developer-847:
> Should the validation\max length check be triggered on first load. For example if you have a value set that exceeds the maxCharacters?
> 
> I have a long `value` set which exceeds the limit.
> 
> Initially this displays as follows: image
> 
> After pressing a key in the input field, the warning is applied: image
> 
> The old behaviour appears to be to trigger the error on load: image

With the changes in this PR, it is not possible to get the "maximum length exceeded" error when typing in characters now because it prevents any more input once the maxCharacters amount has been reached.

I have pushed up some changes now which just cut off the value based on the maxCharacters. So now, if the `value` prop is set on page load, or if the prop is updated, the input in the text field cannot be over the maxCharacters amount. I felt that this would be the best solution as it essentially matches the behaviour you were to get if you actually typing in the text field (including the fact that if you try to type more characters, the warning will then appear). Hopefully this is ok :)

GCHQ-Developer-847:
> there no longer seems to be a visual test (with baseline image) where the maxCharacters is exceeded. I think it would be good to have one, just to ensure styling doesn't change

It is no longer possible to go over the maximum allowed number of characters so this error cannot occur any more which is why the image / test has been removed completely

ad9242:
> > Should the validation\max length check be triggered on first load. For example if you have a value set that exceeds the maxCharacters?
> > I have a long `value` set which exceeds the limit.
> > Initially this displays as follows: image
> > After pressing a key in the input field, the warning is applied: image
> > The old behaviour appears to be to trigger the error on load: image
> 
> After the changes in this PR, it is not possible to get the "maximum length exceeded" error when typing in characters now because it prevents any more input once the maxCharacters amount has been reached.
> 
> I have pushed up some changes now which just cut off the value based on the maxCharacters. So now, even if the `value` prop is set on page load, or if the prop is updated, the value displayed in the text field cannot be over the maxCharacters amount. I felt that this would be the best solution as it essentially matches the behaviour you were to get if you actually typing in the text field (including the fact that if you try to type more characters after the prop has been set, the warning will then appear). Hopefully this is ok :)

The only slight issue I can see is that, if the initial value gets truncated, it will trigger an icChange event with the new shortened value. Also, if the data was loaded into a form and then gets submitted, the new shorter value would get submitted and, without some validation being done, potentially overwrite existing data on the server side. I think it would be good to just make it clearer (in the `maxCharacters` prop description?) that the value will get truncated if it exceeds the limit. Just so users are aware and can ensure they validate values

GCHQ-Developer-847:
> > > Should the validation\max length check be triggered on first load. For example if you have a value set that exceeds the maxCharacters?
> > > I have a long `value` set which exceeds the limit.
> > > Initially this displays as follows: image
> > > After pressing a key in the input field, the warning is applied: image
> > > The old behaviour appears to be to trigger the error on load: image
> > 
> > 
> > After the changes in this PR, it is not possible to get the "maximum length exceeded" error when typing in characters now because it prevents any more input once the maxCharacters amount has been reached.
> > I have pushed up some changes now which just cut off the value based on the maxCharacters. So now, even if the `value` prop is set on page load, or if the prop is updated, the value displayed in the text field cannot be over the maxCharacters amount. I felt that this would be the best solution as it essentially matches the behaviour you were to get if you actually typing in the text field (including the fact that if you try to type more characters after the prop has been set, the warning will then appear). Hopefully this is ok :)
> 
> The only slight issue I can see is that, if the initial value gets truncated, it will trigger an icChange event with the new shortened value. Also, if the data was loaded into a form and then gets submitted, the new shorter value would get submitted and, without some validation being done, potentially overwrite existing data on the server side. I think it would be good to just make it clearer (in the `maxCharacters` prop description?) that the value will get truncated if it exceeds the limit. Just so users are aware and can ensure they validate values

Added a note to the prop description 👍

packages/web-components/src/components/ic-text-field/ic-text-field.tsx
```diff
@@ -331,16 +339,19 @@ export class TextField {
...
     if (this.value !== this.initialValue) {
       this.watchValueHandler(this.value);
+    } else {
+      if (this.maxCharacters > 0) {
+        this.value = this.value.substring(0, this.maxCharacters);
+      }
     }
 
-    this.getMaxLengthExceeded(this.value);
+    this.getMaxValueExceeded(this.value);
```
ad9242, line 348: just realized this is slightly inefficient as both `this.getMaxValueExceeded` and `this.getMaxCharactersReached` may have already been called due to the `watchValueHandler` being called directly (line 341), or indirectly when `this.value` is changed on line 344

ad9242 (changes requested):

packages/web-components/src/components/ic-text-field/ic-text-field.tsx
GCHQ-Developer-847: I can't seem to get it to work...

I tried moving those two methods into an `else` on line 344, but then [tests were failing](https://github.com/mi6/ic-ui-kit/actions/runs/9940487837/job/27457312100). It was like line 343 was not causing the `watchValueHandler` to be called at all to set the character count correctly.

So I'm a bit confused 😅 don't know if it's just best to keep it as is even though it is a bit inefficient

GCHQ-Developer-299 (approved):

PR #2116, opened 2024-07-06
fix(web-components): remove aria-hidden in see more/less buttons

### LINKED ISSUE #1854: [ic-typography] accessibility error on see more/see less button
## Summary of the bug 
When running accessibility tests using accessibility insights and the storybook a11y add on, the same error occurs. 

aria-hidden-focus - ensure aria-hidden components aren’t focusable. 

This is because the see more/see less typography button is focusable and aria-hidden

## 🪜 How to reproduce 
1. Go to truncation example on storybook
2. Click on the accessibility tab in storybook or run an accessibility insights fast pass
3. See the error 

## 🧐 Expected behaviour 
It shouldn't produce this accessibility error.

evenstensberg:
## Summary of the changes
Removes aria hidden attribute and enables skipped a11y tests in ic -typography.

## Related issue
Fixes #1854

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes

### DIFF
--- a/packages/web-components/src/components/ic-typography/ic-typography.tsx
+++ b/packages/web-components/src/components/ic-typography/ic-typography.tsx
@@ -203,7 +203,6 @@ export class Typography {
             onBlur={this.truncButtonBlur}
             onMouseDown={this.truncButtonFocusFromMouse}
             onClick={this.toggleExpanded}
-            aria-hidden="true"
           >
             {expanded ? "See less" : "See more"}
           </button>
--- a/packages/react/src/component-tests/IcTypography/IcTypography.cy.tsx
+++ b/packages/react/src/component-tests/IcTypography/IcTypography.cy.tsx
@@ -132,7 +132,7 @@ describe("IcTypography visual regression and a11y tests", () => {
 
     cy.checkHydrated(TYPOGRAPHY_SELECTOR);
 
-    //cy.checkA11yWithWait();
+    cy.checkA11yWithWait();
     cy.compareSnapshot({
       name: "truncation",
       testThreshold: setThresholdBasedOnEnv(DEFAULT_TEST_THRESHOLD + 0.071),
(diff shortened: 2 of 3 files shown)


GCHQ-Developer-530:
Hi @evenstensberg, part of #1854 is to uncomment out the a11y checks in the cypress tests for typography? Please could you add a new commit for that work? Thank you

packages/web-components/src/components/ic-typography/ic-typography.tsx
```diff
@@ -203,7 +203,7 @@ export class Typography {
             onBlur={this.truncButtonBlur}
             onMouseDown={this.truncButtonFocusFromMouse}
             onClick={this.toggleExpanded}
-            aria-hidden="true"
+            tabindex="0"
```
ad9242, line 206: I don't think that explicitly setting `tabindex` to `0` is required, as a `<button>` will be focusable by default. So hopefully just removing aria-hidden is all that is required. The unit tests will fail as a result, so the snapshot files should be updated to reflect the change

evenstensberg: @ad9242 which npm command updates snapshots, and snapshots only?

evenstensberg:
@GCHQ-Developer-530 Rebased the change, but need to update the snapshots too.

packages/web-components/src/components/ic-typography/ic-typography.tsx
ad9242: I don't think there is an npm command for it, but you can delete the existing snapshot file & then run the tests locally (npm run test), to have it regenerate the file and then validate manually that the changes are what you would expect

evenstensberg:
@ad9242 @GCHQ-Developer-530 snapshot is updated, PTAL!

GCHQ-Developer-530:
Hi @evenstensberg, this work still needs the cypress update I mentioned in my previous comment. It should just be a case of uncommenting out any checkA11yWithWait() commands from the typography tests e.g. [this line in the file](https://github.com/mi6/ic-ui-kit/blob/develop/packages/react/src/component-tests/IcTypography/IcTypography.cy.tsx#L135) should be uncommented. Thanks

GCHQ-Developer-530:
Also, please can you squash your 2 web-components scoped commits into one commit. Thank you

evenstensberg:
@GCHQ-Developer-530 Done. As the a11y test was in react package it's in its own commit.

GCHQ-Developer-530 (approved):

ad9242 (approved):

Issue #1908, opened 2024-05-23
[ic-page-header] storybook accessibility plug in complaining of two elements with role="banner"


GCHQ-Developer-530:
## Summary of the bug 
On storybook examples of page header with top nav, the built in storybook accessibility plug in is failing due to two elements having a banner role and all roles should be unique. 

## 🪜 How to reproduce 
Tell us the steps to reproduce the problem: 
1. Go to page-header sticky-with-content-and-footer example
2. Click on accessibility plug in in the drawer
4. See the error 

alternatively the error can be replicated on our https://design.sis.gov.uk/top-navigation-page-header-pattern/
If you do a full page scan using the 'axe Dev Tools' browser plugin, it appears as a moderate issue.
If you do an Accessibility Assessment using the IBM Equal Access Accessibility Checker it also appears.

## 📸 Screenshots or code 
<img width="500" alt="Screenshot 2024-05-23 at 14 38 12" src="https://github.com/mi6/ic-ui-kit/assets/111882034/cb9f1599-45c3-4aed-bae4-9335d8d9b0aa">

## 🧐 Expected behaviour 
It should only have one banner.

## Additional info 
Weirdly page header doesn't have a banner role. This doesn't fail accessibility insights fast pass either so it's hard to tell if it's a weird quirk of storybook or not.

gd2910:
Investigate ways around this, consider https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/presentation_role

Or conditional role on IcPageHeader when in conjunction with IcTopNavigation

gd2910:
Test IcTopnavigation and IcPageHeader layout patterns to recreate issue.

GCHQ-Developer-299:
Error can be replicated on our https://design.sis.gov.uk/top-navigation-page-header-pattern/ 
If you do a full page scan using the 'axe Dev Tools' browser plugin, it appears as a moderate issue.
If you do an Accessibility Assessment using the IBM Equal Access Accessibility Checker it also appears.

MI6-255:
Take banner off Page Header and find a role to replace it

PR #2131, opened 2024-07-11
1908 page header complaining of two elements with role banner

GCHQ-Developer-530:
## Summary of the changes
Update role on page header to presentation and add aria label to copy old behaviour more

## Related issue
#1908 

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes

### DIFF
--- a/packages/web-components/src/components/ic-page-header/ic-page-header.tsx
+++ b/packages/web-components/src/components/ic-page-header/ic-page-header.tsx
@@ -180,12 +180,14 @@ export class PageHeader {
           ["sticky"]: sticky,
           ["sticky-desktop"]: !sticky && stickyDesktopOnly,
         }}
+        aria-label={this.el.ariaLabel || "page header"}
       >
         <header
           class={{
             ["border-bottom"]: border,
             ["tabs"]: isSlotUsed(this.el, "tabs"),
           }}
+          role="presentation"
         >
           <ic-section-container
             aligned={aligned}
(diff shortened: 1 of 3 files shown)


packages/web-components/src/components/ic-page-header/ic-page-header.tsx
```diff
@@ -180,12 +180,14 @@ export class PageHeader {
           ["sticky"]: sticky,
           ["sticky-desktop"]: !sticky && stickyDesktopOnly,
         }}
+        aria-label="page header"
```
ad9242, line 183: Could hard-coding this cause some problems, as will override if people are currently setting an aria-label. Maybe this should be the default if nothing already set?

ad9242 (changes requested):

packages/web-components/src/components/ic-page-header/ic-page-header.tsx
```diff
@@ -180,12 +180,14 @@ export class PageHeader {
...
       >
         <header
```
GCHQ-Developer-299, line 185: If we're stripping <header>'s semantic value with the role=presentation, is there any reason to keep it as a <header> element?  Could we replace it with a <div> (and therefore not need the role=presentation)

GCHQ-Developer-530: I think changing the element could make it a breaking change if users are looking for `<header>` in tests, but if we think this is the best way around it then I could move this into the v3 work

GCHQ-Developer-299: darn, that is tricky. I don't want to prevent this fix going into V2. Would two PRs (one here for the base fix, then once this is merged in another for that refactor) be too much?

GCHQ-Developer-530: Just been having a look around and I think it should still stay as a `<header>` element, it looks like under the hood it does more than setting the role to banner

GCHQ-Developer-299: Righto, no worries then :)

GCHQ-Developer-299 (approved):

ad9242 (approved):

Issue #1820, opened 2024-05-01
[ic-radio-group] overriding orientation="horizontal" doesn't work


GCHQ-Developer-530:
## Summary of the bug 
When you set orientation to horizontal on radio group but one of the radio options has a text field, or there's more than 2 radio options, it should set the orientation to vertical. This isn't currently working.

## 🪜 How to reproduce 
Tell us the steps to reproduce the problem: 
1. Go to storybook locally
2. Update radio conditional dynamic example to have orientation="horizontal"
4. See that it renders horizontal and not vertical

## 🧐 Expected behaviour 
It should update to vertical for those use cases. It should also update if it runs out of space on the screen so might be worth checking resize observer code as well.

## Additional info 
This test was removed due to failures, it should be added back in with this work:
```
  it("should change the orientation of the radio group to vertical if the user has additional fields on any of the radio buttons in the group", async () => {
    const page = await newSpecPage({
      components: [RadioGroup, RadioOption, TextField],
      html: `<ic-radio-group label="test label" name="test" required orientation="horizontal">
        <ic-radio-option value="test1" selected></ic-radio-option>
        <ic-radio-option value="test" disabled label="test label" group-label="test group">
         <ic-text-field slot="additional-field" placeholder="Placeholder" label="Test label"></ic-text-field>
        </ic-radio-option>
      </ic-radio-group>`,
    });

    expect(page.rootInstance.currentOrientation).toMatch("vertical");
  });
```

PR #2171, opened 2024-07-17
1820 ic radio group overriding orientation

GCHQ-Developer-773:
## Summary of the changes
Now when the radio group has an additional field it will change from horizontal to vertical, along with following this rule if there are more then 2 radio options in the group.

Also, radio groups will switch from horizontal to vertical if the totalwidth becomes greater then the client width it finds itself in.

## Related issue
#1820 

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes

### DIFF
--- a/packages/docs/docs.json
+++ b/packages/docs/docs.json
@@ -1,5 +1,5 @@
 {
-  "timestamp": "2024-07-19T12:14:10",
+  "timestamp": "2024-08-12T14:32:40",
   "compiler": {
     "name": "@stencil/core",
     "version": "4.10.0",
@@ -10,7 +10,7 @@
       "filePath": "src/components/ic-accordion/ic-accordion.tsx",
       "encapsulation": "shadow",
       "tag": "ic-accordion",
-      "readme": "# ic-accordion\n\n\n",
+      "readme": "# ic-accordion\r\n\r\n\r\n\r",
       "docs": "",
       "docsTags": [
         {
@@ -202,7 +202,7 @@
       "filePath": "src/components/ic-accordion-group/ic-accordion-group.tsx",
       "encapsulation": "shadow",
       "tag": "ic-accordion-group",
-      "readme": "# ic-accordion-title\n\n\n",
+      "readme": "# ic-accordion-title\r\n\r\n\r\n\r",
       "docs": "",
       "docsTags": [
         {
@@ -441,7 +441,7 @@
       "filePath": "src/components/ic-alert/ic-alert.tsx",
       "encapsulation": "shadow",
       "tag": "ic-alert",
-      "readme": "# ic-alert\n\n\n",
+      "readme": "# ic-alert\r\n\r\n\r\n\r",
       "docs": "",
       "docsTags": [
         {
@@ -697,7 +697,7 @@
       "filePath": "src/components/ic-back-to-top/ic-back-to-top.tsx",
       "encapsulation": "shadow",
       "tag": "ic-back-to-top",
-      "readme": "# ic-back-to-top\n\n\n",
+      "readme": "# ic-back-to-top\r\n\r\n\r\n\r",
       "docs": "",
       "docsTags": [],
       "usage": {},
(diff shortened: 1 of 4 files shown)


packages/docs/docs.json
GCHQ-Developer-299, line 1: None of these docs changes regard ic-radio - they all seem to be formatting changes due to your OS - and since you've not changed the prop/slot comments in radio button there wouldn't be any docs changes.   Please remove this commit from your PR :)

GCHQ-Developer-299 (changes requested):
In the ticket (#1820) it's requested that this work add back in a spec test. Could you please add that back in?

packages/docs/docs.json
GCHQ-Developer-773: Hey, Cheers for the comment, I've removed the docs based commit now.
Also, the test from 1820 is already in the ic-radio-group.spec.ts at line 295. Not sure if this needs to be elsewhere.

GCHQ-Developer-299 (changes requested):
Have noticed unusual behaviour on refresh - the orientation is horizontal for a moment before becoming vertical. This does not occur on develop branch

packages/web-components/src/components/ic-radio-group/ic-radio-group.stories.mdx
```diff
@@ -165,21 +165,10 @@ import readme from "./readme.md";
               placeholder="Placeholder"
               label="What's your favourite type of coffee?"
             >
-            </ic-text-field>
-          </ic-radio-option>
-          <ic-radio-option
```
GCHQ-Developer-299, line 170: Any reason for removing this third checkbox?

GCHQ-Developer-773: I was testing to make sure it did with two but with the additional field to make it go vertical, I can add this back in, must've forgot I removed it.

packages/web-components/src/components/ic-radio-group/ic-radio-group.stories.mdx
```diff
@@ -140,7 +156,11 @@ import readme from "./readme.md";
             .querySelector("ic-radio-group")
             .addEventListener("icChange", handleIcChange);
         </script>
-        <ic-radio-group label="Conditional dynamic" name="1">
+        <ic-radio-group
+          label="Conditional dynamic"
+          orientation="horizontal"
```
GCHQ-Developer-299, line 161: Should the horizontal orientation be removed from this story? My thinking is that the horizontal overide isn't being tested in this one, it's all about the Conditional rendering

GCHQ-Developer-299 (approved):

packages/web-components/src/components/ic-radio-group/ic-radio-group.tsx
```diff
@@ -181,21 +184,32 @@ export class RadioGroup {
...
-        totalWidth += clientWidth;
-        if (i < arr.length - 1) totalWidth += 40;
-      });
+      if (Array.isArray(this.radioOptions)) {
+        this.radioOptions.forEach(({ clientWidth }, i, arr) => {
+          totalWidth += clientWidth;
+          if (i < arr.length - 1) totalWidth += 40;
+        });
+      } else {
+        console.warn("radioOptions is undefined or not an array");
```
jd239, line 193: I'm not sure if we need this. `radioOptions` is an internal variable so it's not something we would expose to the consumer. If it fails gracefully, that should be good enough.

packages/web-components/src/components/ic-radio-group/ic-radio-group.tsx
```diff
@@ -181,21 +184,32 @@ export class RadioGroup {
...
-        this.currentOrientation === "horizontal" &&
-        totalWidth > this.radioContainer.clientWidth
-      ) {
-        this.currentOrientation = "vertical";
-      } else if (
-        this.currentOrientation === "vertical" &&
-        totalWidth < this.radioContainer.clientWidth
-      ) {
-        this.currentOrientation = "horizontal";
+      if (this.initialOrientation == "horizontal") {
```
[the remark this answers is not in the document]
jd239: Could we do another one for `vertical` as well please?

jd239 (changes requested):
Great job. Just a few comments.

jd239 (approved):

GCHQ-Developer-299 (approved):

PR #2266, opened 2024-08-07
test(react): add fix for flaky Cypress test

MI6-286:
Add fix for flaky Cypress test


packages/react/src/component-tests/IcButton/IcButton.cy.tsx
```diff
@@ -686,7 +686,7 @@ describe("IcButton visual regression tests in high contrast mode", () => {
   it("should render icon variants in high contrast mode", () => {
     mount(<IconButtonGroup />);
```
jd239, line 688: Do we need to check that the component is hydrated before running the screenshot?

jd239 (changes requested):
Just a single comment

packages/react/src/component-tests/IcButton/IcButton.cy.tsx
MI6-286: Of course, my bad. Added those in now.

jd239 (dismissed):

jd239 (approved):

GCHQ-Developer-299 (approved):

PR #2315, opened 2024-08-19
Dark mode prep

ad9242:
## Summary of the changes
This is an initial PR to help us get going on dark mode (and actually will enable theming too)

This has been implemented in such a way that there is a hierarchy of settings

If someone is not using the ic-theme component, then the os\device settings for dark\light mode will be used
If using ic-theme, then there is “mode” prop, which can be set to
“auto”, to honour device settings
“dark”, to always use our dark mode
“light” to always use our light mode
So long as  we set all our tokens appropriately, we should not need to implement switching of the “appearance” prop on all our components/ have each component listening for a change of a top level switch.

Ultimately, it will also be possible to override an individual instance of a component too (using the “appearance” prop, but that’s not really shown here (slight caveat to that below using a different mechanism)

There has been some change s to the structure of our css tokens. They are now split out into files such as “borders.css”, “fonts.css”, “colors.css” rather than a single file. There are also separate files for light & dark mode, so the tokens can be specified there. At the moment, the tokens are set to use the light values, even in dark mode, but you can pull the code down locally & uncomment the dark tokens to see them in action. Obviously more tokens will be added as we go forwards

One nice side effect of the ic-theme work is that you can actually use it as a mechanism to wrap some components in a different mode\theme. 

For example, you could set a light mode to wrap your whole app, but then maybe have a single card that is wrapped in a ic-theme with `mode=dark` set

```
<ic-theme mode=”light”>
   …..loads of light comps
   <ic-theme mode=”dark”>
     <ic-card>
        Dark mode card stuff	
     </ic-card>
    </ic-theme>
 …. More light mode stuff
</ic-theme>
```

I’m not proposing this as a solution/alternative to having an “appearance” prop on all the components, but it’s a little extra bonus

You may notice that currently very little changes depending on the mode prop. This is because the ic-typography component needs to be modified to honour a text colour. However, doing so is a huge breaking change (all the visual tests will break). So for the time being, if you want to see this in action, please uncomment the code in the ic-typography.css file


packages/react/src/stories/ic-theme.stories.mdx
```diff
@@ -185,3 +189,197 @@ export const Controlled = () => {
...
+            <p>
+              This is some text and
+              <ic-link href="/"> this is an inline link ic-link</ic-link>
+              within the text
+            </p>
+          </IcTypography>
+        </IcAlert>
+        <div style={{ border: "1px solid #ffffff", padding: "4px", margin: "4px"}}>
+          <span style={{color:"#ffffff"}}>Content in here should always be dark mode</span>
+           <IcTheme mode="dark">
```
gd2910, line 322: Would it be helpful to use the components to indicate that they are wrapped with dark-mode IcTheme? 
i.e. 'This IcCard is wrapped by IcTheme with dark-mode enabled'?

packages/react/src/stories/ic-theme.stories.mdx
```diff
@@ -185,3 +189,197 @@ export const Controlled = () => {
     </div>
   </Story>
 </Canvas>
+
+
+### Playground
+
+export const defaultArgs = {
+  color: "rgba(27, 60, 121, 1)",
```
jd239, line 197: Do we need to set a color?

packages/web-components/src/global/borders.css
```diff
@@ -0,0 +1,35 @@
+/* borders */
+:root {
+  --ic-border-radius: 2px;
```
gd2910, line 3:
```suggestion
  --ic-border-radius: 0.125rem;
```

packages/web-components/src/global/borders.css
```diff
@@ -0,0 +1,35 @@
...
+  --ic-border-radius-inset: 11px;
```
gd2910, line 4:
```suggestion
  --ic-border-radius-inset: 0.688rem;
```

packages/web-components/src/global/borders.css
```diff
@@ -0,0 +1,35 @@
...
+  --ic-border-width: 0.0625rem;
+  --ic-border-focus: 0 0 0 2px var(--ic-architectural-white),
```
gd2910, line 6:
```suggestion
  --ic-border-focus: 0 0 0 0.125rem var(--ic-architectural-white),
```

packages/web-components/src/global/borders.css
```diff
@@ -0,0 +1,35 @@
...
+    0 0 0 4px var(--ic-focus-blue), 0 0 0 8px var(--ic-focus-glow);
```
gd2910, line 7:
```suggestion
    0 0 0 0.25rem var(--ic-focus-blue), 0 0 0 0.5rem var(--ic-focus-glow);
```

packages/web-components/src/global/borders.css
```diff
@@ -0,0 +1,35 @@
...
+  --ic-border-focus-inset: inset 0 0 0 4px var(--ic-focus-glow),
```
gd2910, line 8:
```suggestion
  --ic-border-focus-inset: inset 0 0 0 0.25rem var(--ic-focus-glow),
```

packages/web-components/src/global/borders.css
```diff
@@ -0,0 +1,35 @@
...
+    inset 0 0 0 6px var(--ic-focus-blue),
```
gd2910, line 9:
```suggestion
    inset 0 0 0 0.375rem var(--ic-focus-blue),
```

packages/web-components/src/global/borders.css
```diff
@@ -0,0 +1,35 @@
...
+    inset 0 0 0 8px var(--ic-architectural-white);
```
gd2910, line 10:
```suggestion
    inset 0 0 0 0.5rem var(--ic-architectural-white);
```

packages/web-components/src/global/borders.css
```diff
@@ -0,0 +1,35 @@
...
+  --ic-border-focus-dark: 0 0 0 2px white, 0 0 0 5px rgba(255 255 255 / 52%);
```
gd2910, line 11:
```suggestion
  --ic-border-focus-dark: 0 0 0 0.125rem white, 0 0 0 0.313rem rgba(255 255 255 / 52%);
```

packages/web-components/src/global/colors.css
```diff
@@ -0,0 +1,265 @@
+@import "system-light-mode.css";
+@import "system-dark-mode.css";
+
+/* colours */
+:root {
+  --ic-keyline-darken-rgb: rgb(0 0 0 / 20%);
+  --ic-keyline-darken: 1px solid var(--ic-keyline-darken-rgb);
```
gd2910, line 7:
```suggestion
  --ic-keyline-darken: 0.063rem solid var(--ic-keyline-darken-rgb);
```

packages/web-components/src/global/colors.css
```diff
@@ -0,0 +1,265 @@
...
+  --ic-keyline-lighten-rgb: rgb(255 255 255 / 20%);
+  --ic-keyline-lighten: 1px solid var(--ic-keyline-lighten-rgb);
```
gd2910, line 9:
```suggestion
  --ic-keyline-lighten: 0.063rem solid var(--ic-keyline-lighten-rgb);
```

packages/web-components/src/global/colors.css
```diff
@@ -0,0 +1,265 @@
...
+  --ic-hyperlink-visited: #330072;
+  --ic-hyperlink-hover: #7c2855;
+  --ic-hyperlink-contrast: #5c98f2;
+  --ic-hyperlink-contrast-visited: #bc75ff;
+  --ic-hyperlink-contrast-hover: #db76ac;
+
+  /* focus */
+  --ic-focus-blue: #0044d7;
+  --ic-focus-glow: #80a1e8;
+  --ic-hc-focus-outline: 3px solid transparent;
```
gd2910, line 261:
```suggestion
  --ic-hc-focus-outline: 0.188rem solid transparent;
```

packages/web-components/src/global/borders.css
ad9242: leaving for now

packages/web-components/src/global/borders.css
ad9242: leaving for now

packages/web-components/src/global/borders.css
ad9242: leaving for now

packages/web-components/src/global/borders.css
ad9242: leaving for now

packages/web-components/src/global/borders.css
ad9242: leaving for now

packages/react/src/stories/ic-theme.stories.mdx
ad9242: it's needed to set the colour of the top nav\hero etc (the prop will be renamed "brand" in a future ticket)

jd239 (approved):

PR #2316, opened 2024-08-19
fix(web-components): restore slotted-svg styles in footer link

GCHQ-Developer-299:
Restored some CSS in footer-link to style slotted SVG icons correctly in theme switcher

## Summary of the changes
See the slottedSVG displaying incorrectly in black on v3.0.0/develop branch:
Screenshot 2024-08-19 at 16 11 08

### DIFF
--- a/packages/web-components/src/components/ic-footer-link/ic-footer-link.css
+++ b/packages/web-components/src/components/ic-footer-link/ic-footer-link.css
@@ -39,14 +39,6 @@
   fill: var(--ic-hyperlink);
 }
 
-.ic-link.dark > .ic-link-open-in-new-icon {
-  fill: var(--ic-color-primary-text);
-}
-
-.ic-link.light > .ic-link-open-in-new-icon {
-  fill: var(--ic-color-white-text);
-}
-
 .ic-link-open-in-new-icon > svg {
   width: var(--ic-space-md);
   height: var(--ic-space-md);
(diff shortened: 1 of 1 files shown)


packages/web-components/src/components/ic-footer-link/ic-footer-link.css
```diff
@@ -90,3 +82,18 @@
 :host(.breadcrumb-link.current-page) .ic-link:visited {
   color: var(--ic-color-primary-text);
 }
+
+:host(.footer-link-light) ::slotted(svg) {
+  fill: var(--ic-architectural-white);
+}
+
+:host(.footer-link-dark) ::slotted(svg) {
+  fill: var(--ic-architectural-black);
```
ad9242, line 91: Rather than using those architectural white\black tokens, does it work if you just use --ic-theme-text? As that should get calculated to be the appropriate color

ad9242 (changes requested):

packages/web-components/src/components/ic-footer-link/ic-footer-link.css
GCHQ-Developer-299: Good suggestion! Tried it out locally and it seems to work :)

ad9242 (approved):

GCHQ-Developer-530 (approved):

PR #2325, opened 2024-08-20
Initial dark mode prep

ad9242:
## Summary of the changes
This is an initial PR to help us get going on dark mode (and actually will enable theming too)

This has been implemented in such a way that there is a hierarchy of settings

If someone is not using the ic-theme component, then the os\device settings for dark\light mode will be used
If using ic-theme, then there is “theme” prop, which can be set to
“system”, to honour device settings
“dark”, to always use our dark mode
“light” to always use our light mode

Ultimately, it will also be possible to override an individual instance of a component too (using the a prop, but that’s not really shown here (slight caveat to that below using a different mechanism)

There has been some change s to the structure of our css tokens. They are now split out into files such as “borders.css”, “fonts.css”, “colors.css” rather than a single file. There are also separate files for light & dark mode, so the tokens can be specified there. At the moment, the tokens are set to use the light values, even in dark mode, but you can pull the code down locally & uncomment the dark tokens to see them in action. Obviously more tokens will be added as we go forwards

One nice side effect of the ic-theme work is that you can actually use it as a mechanism to wrap some components in a different mode\theme. 

For example, you could set a light mode to wrap your whole app, but then maybe have a single card that is wrapped in a ic-theme with `theme=dark` set

```
<ic-theme theme=”light”>
   …..loads of light comps
   <ic-theme theme=”dark”>
     <ic-card>
        Dark mode card stuff	
     </ic-card>
    </ic-theme>
 …. More light mode stuff
</ic-theme>
```

I’m not proposing this as a solution/alternative to having an “appearance” prop on all the components, but it’s a little extra bonus

You may notice that currently very little changes depending on the mode prop. This is because the ic-typography component needs to be modified to honour a text colour. However, doing so is a huge breaking change (all the visual tests will break). So for the time being, if you want to see this in action, please uncomment the code in the ic-typography.css file and also enable the dark mode variables in global/color-mode.css


packages/web-components/src/global/colors.css
```diff
@@ -0,0 +1,265 @@
...
+    - The ICDS does not use pure black RGB(0, 0, 0) as it's font, it uses #0b0c0c/RGB(11, 12, 12)
+    - The max() CSS function therefore ensures any negative numbers can be no less than 0
+    - As any negative numbers are now 0, finally the numbers 11, 12 and 12 are added to the R, G, B values to make the ICDS shade of black
+    - A similar adjustment is made to add 60, 70 and 77 to give hover & active colors
+    - The min() CSS function is then used to ensure that none of the values exceeds 255
+    - This is due to an issue found in Chromium v117 (sept 2023), where very large values would result the alpha value being ignored and everything appearing white
+    - The result then returns the RGB values for the ICDS white or black font color, dependent on which is the most color contrast compliant
+    This is a similar calculation to it's JS counterpart: "getThemeFontColor"
+  */
+  --ic-theme-dark-mode-threshold: 133.3505;
```
MI6-255, line 220: Would we actually want some of the old theme (soon to be called brand) variables duplicated? As regardless of the theme (light or dark), the brand colours would remain the same, as would the calculations throughout. Or is it a case we need them for it to work?

ad9242: as discussed, these are not duplicated

GCHQ-Developer-530 (dismissed):

gd2910 (approved):

GCHQ-Developer-530 (approved):

GCHQ-Developer-299 (approved):

MI6-255:
Looks good, my only comment is one we already discussed. Less about this PR, but more about one of your next maybe. I want to make sure that customers apps will not be broken from them adopting v3, where they haven't or have no plans to support dark mode, if someone comes with dark mode enabled and no theme, it simply defaults to dark.

I couldn't see any code in here I could reference, but my ideas to achieve that, in order of preference:

1. Where either the ic-light or ic-dark css class is applied when set to "inherit" at a component level, it either inherits what is set at a theme level, or if no theme, it defaults to light, to match how it works currently. Surely there's some logic we can put in place to just default to light, if there is no theme?
2. We simply force people to adopt ic-theme in our getting started, and say without it, system settings will break apps. Currently it is optional, so we'd change it to not be.
3. There is an option of adding some css manually as you've mentioned, but I feel at that point, just tell them to use theme as that's what it does.

My top preference being option 1.

PR #2327, opened 2024-08-20
[ic-data-table]: Truncation bug fixes

jd239:
## Summary of the changes
Bug fixes include re-rendering truncation with large datasets and sorting. There was an issue with using setting 'items per page' to 'All' and then sorting a column. Some of the data was not truncated. This ticket fixes that. There is also an update to the demos within storybook as the some of the data was appearing underneath the wrong column and dataType. 

## Related issue
N/A

## Checklist

Check list checks carried out in https://github.com/mi6/ic-ui-kit/pull/2138

### Testing


ad9242:
A couple of little issues with playing around with the "Dev Area" story, 

1) after first loading, the "toggle truncation" button seems to have no effect
2)I seem to be able to get it in a state where there is a tooltip applied to cells, but no ellipsis. I 
 - loaded the story
 - click to set row height to 40
 - clicked the "toggle truncation button 2 times.

and this happened (note obscured text & tooltip is way below the actual cell it relates to):

image

jd239:
> 1. after first loading, the "toggle truncation" button seems to have no effect

By default, the set row height is auto so it will be the max height to based on the content. 
For the data to truncate, the row height has to be set small enough (i.e. 40) for the data to be truncated.
I have added more detail about this in the Storybook and will be copied across to the guidance site.

>   2)I seem to be able to get it in a state where there is a tooltip applied to cells, but no ellipsis. I
> 
> * loaded the story
> * click to set row height to 40
> * clicked the "toggle truncation button 2 times.
> 
> and this happened (note obscured text & tooltip is way below the actual cell it relates to):
> 

The order in which the props are set play a role. If the global row height is set first, the row height will not change because the max height of the row will be dictated by the table cell with the most data (lines of data within the a cell). The data will not appear cut off. If the type of truncation is set and then the row height is set after to a smaller number like 40, the data will truncate correctly as the cell space is too small to display the data fully.

Toggling the truncation pattern again to show hide requires an additional line within the table cell but the row height has been set explicitly to 40 initially. 

I guess we need to decide whether truncation would be changed on the fly and if so, should changing the truncation pattern also change the row height which was only explicitly set in the first instance?

GCHQ-Developer-299 (dismissed):

GCHQ-Developer-299 (dismissed):

GCHQ-Developer-299 (approved):

ad9242 (approved):

PR #2420, opened 2024-09-12
fix(web-components): breadcrumb group collapsed prop issue

### LINKED ISSUE #1998: [ic-breadcrumb-group] collapsed prop won't update once set
## Summary of the bug 
Once a boolean value has been set for the collapsed prop, toggling it does not update the collapsed state. 

I tried moving the collapsed code from the `componentWillLoad` to its own private function so that it could be called from the `componentWillLoad` and from a `@Watch` but then when toggling it just adds more `...` links rather than toggling between `Breadcrumb 2` and `...`

The collapsed prop also doesn't take immediate effect.
The styling for the collapsed section also doesn't take into account the current appearance. It always takes the default styling

https://github.com/mi6/ic-ui-kit/assets/113986285/0d7db269-aab0-4804-8bc4-8f37ad76190e

## 🪜 How to reproduce 
1. Go to breadcrumb playground (should be completed by the end of #1970)
2. Toggle collapsed boolean
3. See the error 

## 🧐 Expected behaviour 
Should be able to toggle between collapsed and uncollapsed not just by clicking the ellipses.
Ellipsis should inherit current Appearance

### LINKED ISSUE #2028: [ic-breadcrumb-group] backBreadcrumbOnly bug when toggled between true and false
There's a bug with breadcrumb group:
When you set backBreadcrumbOnly as true and then set it as false, the back arrow stays. It replaces the chevron that should be there

https://github.com/mi6/ic-ui-kit/assets/113986285/1a4a3df5-a850-496e-9b5d-c33f629d010e

_Originally posted by @GCHQ-Developer-112 in https://github.com/mi6/ic-ui-kit/issues/2001#issuecomment-2178087716_

lz405:
Fixed the issue with the breadcrumb group collapsed prop not toggling properly and fixed the back arrow not turning back into a chevron

fix #1998 
fix #2028

## Summary of the changes
Fixed the issue where the breadcrumb group prop would, when toggled, not remove the collapsed ellipses element so they would just stack up
Also fixed the issue where, when toggled back and forth, the back arrow for the 'back breadcrumb only' prop would not switch back to a chevron icon

## Related issue
#1998 
#2028

### DIFF
--- a/packages/react/src/component-tests/IcBreadcrumb/IcBreadcrumb.cy.tsx
+++ b/packages/react/src/component-tests/IcBreadcrumb/IcBreadcrumb.cy.tsx
@@ -9,6 +9,8 @@ import {
   WithCurrentPage,
   Collapsed,
   Appearance,
+  ToggleCollapsed,
+  ToggleBackBreadcrumb,
 } from "./IcBreadcrumbTestData";
 import {
   BE_VISIBLE,
@@ -56,6 +58,23 @@ describe("IcBreadcrumb end-to-end tests", () => {
     cy.get('[page-title="Coffee"]').should(NOT_BE_VISIBLE);
   });
 
+  it("should switch from default view to back arrow with parent breadcrumb when backBreadcrumbOnly prop is set on IcBreadcrumbGroup and revert when prop is removed", () => {
+    mount(<ToggleBackBreadcrumb />);
+
+    cy.checkHydrated(IC_BREADCRUMB_LABEL);
+    cy.get("ic-button").click();
+    cy.get('ic-breadcrumb[page-title="Beverages"]').should(
+      HAVE_ATTR,
+      "show-back-icon"
+    );
+    cy.get("ic-button").click();
+    cy.get('ic-breadcrumb[page-title="Beverages"]').should(
+      HAVE_ATTR,
+      "show-back-icon",
+      "false"
+    );
+  });
+
   it("should render breadcrumb with icons", () => {
     mount(<WithIcons />);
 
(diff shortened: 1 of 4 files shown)


GCHQ-Developer-530:
I'm getting some odd behaviour on the collapsed state still, but I think it's a weird edge case. If I toggle in the side panel, it works fine, but when the toggle is set to collapsed="true" and then I click the ellipsis, and set collapsed to "false" in the side panel again, it completely removes breadcrumb 2. It could just be a storybook nuance, but it'd be good if there was a way the playground collapsed value could reflect the actual collapsed value if the ellipsis is clicked e.g. updating the switch to false if the ellipsis is clicked.

https://github.com/user-attachments/assets/eaf8110a-8b6e-447f-9ed4-94da47a84253

packages/react/src/component-tests/IcBreadcrumb/IcBreadcrumb.cy.tsx
```diff
@@ -56,6 +58,28 @@ describe("IcBreadcrumb end-to-end tests", () => {
     cy.get('[page-title="Coffee"]').should(NOT_BE_VISIBLE);
   });
 
+  it("should switch to the back arrow then back to the chevron", () => {
```
gd2910, line 61: Please could you alter the test title to be more descriptive, `should switch from default view to back arrow with parent breadcrumb when backBreadcrumbOnly prop is set on IcBreadcrumbGroup and revert when prop is removed`

packages/react/src/component-tests/IcBreadcrumb/IcBreadcrumb.cy.tsx
```diff
@@ -56,6 +58,28 @@ describe("IcBreadcrumb end-to-end tests", () => {
...
+    cy.get("ic-button").click();
+    cy.get('ic-breadcrumb[page-title="Beverages"]').should(
+      HAVE_ATTR,
+      "show-back-icon"
+    );
+    cy.get('ic-breadcrumb[page-title="Beverages"]').should(
+      NOT_HAVE_ATTR,
+      "show-back-icon",
+      "false"
+    );
```
gd2910, line 74: Can be removed, not testing anything new

gd2910 (changes requested):
Also getting some strange behaviour in the react storybook where the backBreadcrumbOnly styling is being applied to the current breadcrumb rather than the parent, can't see why this is, may be a SB quirk?

GCHQ-Developer-847 (approved):

GCHQ-Developer-530 (approved):

Issue #1799, opened 2024-04-23
[tabs]: Dynamically generating tabs produces unusual behaviour


jd239:
## Summary of the bug 
Dynamically generating tabs produces the following issues:
- Tab panel appears above the tabs
- The first tab is not selected
- If the second tab is selected and another tab is dynamically removed, the second tab loses the selected state and the first tab is selected
- All tabPanels are displayed when tabs are dynamically loaded and a tab is selected

## 🪜 How to reproduce 
See below

## 📸 Screenshots or code 
Code example which can be added to Stackblitz or application to reproduce:

Link to stackblitz: https://stackblitz.com/edit/vitejs-vite-ywsvuc?file=src%2FApp.tsx

## 🖥 📱 Device 
- Type: Desktop
- Device: Windows 
- OS version: N/A
- Browser version: Chrome 122

## 🧐 Expected behaviour 
- The tab panel should appear below the tabs
- The first tab should be selected when dynamically adding all tabs
- Only selected tab panel should appear

## 📝 Acceptance Criteria  
If relevant, describe in full detail the different interactions and edge cases that the component or patterns needs to fulfil. 

*Given the tabs has no data* 
*When I add tabs dynamically*  
*Then the first tab should appear selected with the corresponding tab panel appearing underneath* 

## Additional info 
Tell us anything else useful to help us fix or understand the problem.

GCHQ-Developer-741:
Work has been done on the linked branch to address the issues. Currently fixed the first tab not selecting, and the issue around the second tab losing selection during removal could not be replicated. The other two issues are still outstanding.

GCHQ-Developer-299:
Regarding the first issue "Tab panel appears above the tabs" I think this is because IcTabContext doesn't mandate the order in which its child components are rendered (i don't know much about slots) 

I stuck a couple of <p> tags between the Group and Panels and it ended up sandwiching the panels:
![Image](https://github.com/user-attachments/assets/ae9eaa8c-da00-44d4-ba30-4856face924a)

So i think the solution to this might be specifying TabContext's child slots, but that would be a breaking change...  will see if I can solve this by just modifying the Dynamic Rendering story

GCHQ-Developer-299:
Regards "All tabPanels are displayed when tabs are dynamically loaded and a tab is selected"
[...]

PR #2510, opened 2024-10-01
Bugfixes for dynamic tab panels - feedback requested

GCHQ-Developer-299:
## Summary of the changes
Significant refactor of tab components - internal props removed from IcTabPanel, because they were managed by IcTabContext already.

The ticket identifies a number of issues:

### Tab panel appears above the tabs
Fixed - but with the addition of an empty `<span>` inside the 'Dynamic Rendering' story's IcTabContext.
I think a proper fix to this might require enforcing slots in IcTabContext, so it can enforce the order that TabGroup and TabPanels render in? 

### The first tab is not selected
Fixed! 

### If the second tab is selected and another tab is dynamically removed, the second tab loses the selected state and the first tab is selected
Unable to replicate!

### All tabPanels are displayed when tabs are dynamically loaded and a tab is selected
Unable to resolve - I spent a long time looking into this. The issue seems to be that newly created tabs aren't given the `hidden='true'` attribute and that's why they all appear when a new tab is selected. 

I think it's down to the IcTabContext not setting them up correctly - but i spent a significant amount of time trying to adjust configureTabs() and linkTabs()  - and even hard-setting hidden to true on new tab panels didn't stick, they would still render when a new tab was selected 

## Related issue
#1799 

### Testing

### Accessibility 

### Testing content extremes


ad9242:
**Tab panel appears above the tabs**
yep, appears fixed. seems reasonable as otherwise we need a breaking change i guess to resolve as you suggest

**The first tab is not selected**
Fixed

**If the second tab is selected and another tab is dynamically removed, the second tab loses the selected state and the first tab is selected**
Also unable to replicate!

**All tabPanels are displayed when tabs are dynamically loaded and a tab is selected**
Having looked as this too (a while ago). It seems to be a timing issue with iterating over the same array of elements to render tab and the tap-panel. I'm not really sure what can be done to resolve this

GCHQ-Developer-847:
A few comments - generally agree with @ad9242:

**Tab panel appears above the tabs** 
I agree that adding slots for the TabGroup and TabPanels would be a way to fix this properly in future, and that the current fix is reasonable.

**The first tab is not selected** 
Yep, agree that this is fixed. 

**If the second tab is selected and another tab is dynamically removed, the second tab loses the selected state and the first tab is selected** 
I am able to replicate this in the StackBlitz provided in the ticket (although only after refreshing the whole page, and only sometimes!). But I think the changes in this PR have fixed it - I am not able to replicate it in the story that has been added. 

**All tabPanels are displayed when tabs are dynamically loaded and a tab is selected** 
Not sure what can be done about this either - I would need to spend a lot more time looking into it in detail before I can suggest anything really

GCHQ-Developer-847 (approved):
I have approved this because most issues are fixed, and everything looks good to me. 

I thought to create a couple of tickets for the remaining issues so that we can get this PR  merged as it has been open for a while. And then the remaining issues can be done at a later date. But if anyone isn't happy with this let me know :)

GCHQ-Developer-741 (approved):

Issue #1214, opened 2023-10-24
Dark mode: look into different appearances for components


GCHQ-Developer-530:
## Summary 
Currently some components have default, light and dark appearances, and a couple are missing either light or dark. We should investigate and design the missing appearances where necessary.

If you haven't done a component before, reach out to @GCHQ-Developer-530 , @ad9242 , @gd2910 or @GCHQ-Developer-741 who can walk you through the process to help you get started.

## 💬 Description 
Components that need looking into are:
- [x] - Accordion - @GCHQ-Developer-847
- [x] - Alert - @GCHQ-Developer-530
- [x] - Back to top -  @ad9242 
- [x] - Badge - @GCHQ-Developer-530
- [x] - Breadcrumb - @GCHQ-Developer-530
- [x] - Button - @gd2910 
- [x] - Card - @GCHQ-Developer-741 
- [x] - Card - Horizontal - @GCHQ-Developer-741 
- [x] - Checkbox - @GCHQ-Developer-530
- [x] - Chip @gd2910 @ad9242
- [x] - Classification banner - @GCHQ-Developer-530
- [x] - Data entity - @GCHQ-Developer-530
- [x] - Data table - @GCHQ-Developer-741 
- [x] - Date input - @ad9242
- [x] - Date picker - @ad9242
- [x] - Dialog - @GCHQ-Developer-530
- [x] #2527 @gd2910  - theme tokens added to old ic-divider, ready to be cherry-picked into new [ic-divider](https://github.com/mi6/ic-ui-kit/issues/1331) 
- [x] - Empty state - @GCHQ-Developer-530
- [x] - Footer - @GCHQ-Developer-530
- [x] - Hero - @GCHQ-Developer-530
- [x] - Horizontal scroll (This is reliant on button, and impacts the work in page header, tabs and top nav, for now do a work around for appearance on horizontal scroll) @gd2910 done as part of button work
- [x] - Link - @GCHQ-Developer-530
- [x] - Loading indicator - @GCHQ-Developer-741
- [x] - Multi-select - @GCHQ-Developer-741 
- [x] - Navigation Bar - ready for dev (seems to be part of top nav so will tick off when that's done
- [x] - Page header - @GCHQ-Developer-112
- [x] - Pagination - @GCHQ-Developer-530
- [x] - Popover menu - @gd2910 
- [x] - Radio button - @GCHQ-Developer-530
- [x] - Search bar - @GCHQ-Developer-530
- [x] - Section container
[...]

PR #2553, opened 2024-10-09
1214 loading indicator dark mode

GCHQ-Developer-741:
## Summary of the changes
- Removed appearance prop from ic-loading-indicator, adding theme to support dark mode
- Updated storybook, tests and docs with prop changes
- Added styles to components where ic-loading-indicator is used internally (such as ic-toast) as their shadow-doms would prevent the dark mode tokens from reaching the component.

## Related issue
#1214 

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes


ad9242 (approved):

ad9242 (approved):

GCHQ-Developer-847 (changes requested):
A couple of comments:
1. Looks from the designs like the loading indicator needs a `monochrome` prop

2. Some of the variables used are not the same as in the designs (even though they are the same colour). I think it would be good to use the same ones so everything matches the designs, and I think this would reduce the risk of there being unexpected changes in future if we decide to use a different colour for something (e.g. we may want to change the colour of text, and at the moment it would change the loading indicator because `--ic-color-text-primary` is used)

    ... Basically, I think only use the variables listed in the token table on Figma, e.g. instead of `--ic-color-text-primary` for the "inner color" in dark mode, it would be `--ic-action-light`
   
 Hopefully this makes sense - took a while to get my head around things 😅

gd2910 (dismissed):

gd2910 (approved):

packages/react/src/stories/ic-loading-indicator.stories.mdx
```diff
@@ -246,8 +247,8 @@ export const defaultArgTypes = {
 <Canvas>
   <Story name="Linear dark" parameters={{ backgrounds: { default: "dark" } }}>
     <div style={{ display: "grid", gap: "50px" }}>
-      <IcLoadingIndicator label="Default" type="linear" dark />
-      <IcLoadingIndicator label="Large" size="large" type="linear" dark />
+      <IcLoadingIndicator label="Default" type="linear" theme="dark" />
+      <IcLoadingIndicator label="Large" size="large" type="linear" theme="dark" />
       <IcLoadingIndicator
```
GCHQ-Developer-847, line 252: I think this third example needs changing to use `theme="dark"` as well

packages/web-components/src/components/ic-loading-indicator/ic-loading-indicator.stories.mdx
```diff
@@ -258,83 +259,87 @@ export const defaultArgTypes = {
...
           label="Default"
-          appearance="light"
+          theme="dark"
         ></ic-loading-indicator>
         <ic-loading-indicator
           label="Large"
           size="large"
-          appearance="light"
+          theme="dark"
+        ></ic-loading-indicator>
```
GCHQ-Developer-847, line 284: There is already a default-sized loading indicator on line 276 so this one isn't needed

GCHQ-Developer-847 (changes requested):
Just noticed a couple of small things, but looking great overall :)

GCHQ-Developer-847 (approved):

PR #2626, opened 2024-10-22
Ic-radio-option: fix click event not working on components in additional field slot

### LINKED ISSUE #2514: IcRadio additional field with clickable element
When nesting a component inside a radio options additional field any clickable elements can no longer be clicked. This can be seen using this example.

```
import { FC } from "react";
import { IcRadioGroup, IcRadioOption, IcButton } from "@ukic/react";

import "@ukic/fonts/dist/fonts.css";
import "@ukic/react/dist/core/core.css";
import "@ukic/react/dist/core/normalize.css";

const RadioButton: FC = () => {
  return (
    <IcRadioGroup
      name="radio-group-3"
      label="Do you have any special requests?"
    >
      <IcRadioOption
        value="request"
        label="Yes, I want to modify my order"
        additionalFieldDisplay="dynamic"
        selected
      >
        <IcButton onClick={() => alert("hello")} slot="additional-field" />
      </IcRadioOption>
      <IcRadioOption value="none" label="No, standard please" />
    </IcRadioGroup>
  );
};
export default RadioButton;
```

This ticket is to allow for slotting additional elements in the additionalfield & allow then to behave as expected with regard to mouse and keyboard interaction. One thing to consider is what happens regarding the `value` assigned to the radio group, as currently the component assumes an `ic-text-field` will be there to take the value from and use on form submission. It may be the case that it needs documenting that if you slot a different element in there, you have to take responsibility for setting a value for the radio group when that item is selected

Also in scope for this ticket is an issue to do with using the arrow keys. When a textfield or date-input is present as an additional field & has focus, pressing arrowup or arrowdown will cause the selected radio option to change. This is particularly a problem for dateinput as it should allow you to change the values using the arrow keys. To resolve this, focus should only move in this way when it is the radio option itself that has focus

lz405:
## Summary of the changes

fix click event not being emitted properly on components that are in the additional field slot

## Related issue

fix #2514 

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes

### DIFF
--- a/packages/react/src/stories/ic-radio-group.stories.mdx
+++ b/packages/react/src/stories/ic-radio-group.stories.mdx
@@ -136,13 +136,14 @@ import { useState, useEffect, useRef } from "react";
       <IcRadioOption
         additionalFieldDisplay="dynamic"
         value="valueName3"
-        label="option3"
+        label="option3 - clickable component"
       >
-        <IcTextField
+        <IcButton
           slot="additional-field"
-          placeholder="Placeholder"
-          label="What's your favourite type of coffee? "
-        />
+          onClick={()=>alert("Button clicked!")}
+        >
+          Click me for an alert!
+        </IcButton>
       </IcRadioOption>
     </IcRadioGroup>
   </Story>
--- a/packages/react/src/component-tests/IcRadio/IcRadio.cy.tsx
+++ b/packages/react/src/component-tests/IcRadio/IcRadio.cy.tsx
@@ -206,6 +206,17 @@ describe("IcRadio end-to-end tests", () => {
     cy.get(TEXT_FIELD_SELECTOR).eq(2).should("be.visible");
   });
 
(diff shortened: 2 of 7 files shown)


packages/react/src/stories/ic-radio-group.stories.mdx
```diff
@@ -136,13 +136,14 @@ import { useState, useEffect, useRef } from "react";
...
-        <IcTextField
+        <IcButton
           slot="additional-field"
-          placeholder="Placeholder"
-          label="What's your favourite type of coffee? "
-        />
+          onClick={()=>alert("Button clicked!")}
+        >
+          Click me for an alert!
+        </IcButton>
```
GCHQ-Developer-530, line 146: Can we make these same changes in the web components storybook too?

GCHQ-Developer-530 (approved):

ad9242 (approved):

Issue #1897, opened 2024-05-22
[ic-popover-menu] Accessibility failures


GCHQ-Developer-741:
## Requirements

- [ ] Clear the two errors in Accessibility Insights FastPass (check that voiceover/NVDA can still navigate the menu)
- [ ] Add accessibility checks to cypress tests at a point where the menu is open

## Summary of the bug 
When running the ic-popover-menu through Accessibility Insight's FastPass, it highlights two errors when the popover menu is open. These errors are:

====
**Title:** WCAG 1.3.1: Ensures elements with an ARIA role that require child roles contain them (.open,ul[aria-label="popover"][role="menu"])

**Issue:** Ensures elements with an ARIA role that require child roles contain them (aria-required-children - https://accessibilityinsights.io/info-examples/web/aria-required-children)

**Element path:** .open,ul[aria-label="popover"][role="menu"]

**Snippet:** `<ul class="button" aria-label="popover" role="menu"><slot></slot></ul>`

**How to fix:** 
Fix any of the following:
  Element has children which are not allowed: ic-menu-group[aria-label]
  Element uses aria-busy="true" while showing a loader

====

**Title:** WCAG 4.1.2: Ensures all ARIA attributes have valid values (ic-menu-item[label="Actions"],ic-button[aria-owns="ic-popover-submenu-abc"])

**Issue:** Ensures all ARIA attributes have valid values (aria-valid-attr-value - https://accessibilityinsights.io/info-examples/web/aria-valid-attr-value)

**Element path:** ic-menu-item[label="Actions"],ic-button[aria-owns="ic-popover-submenu-abc"]

**Snippet:** `<ic-button exportparts="button" class="button-variant-tertiary button-size-default full-width hydrated" aria-owns="ic-popover-submenu-abc" aria-controls="ic-popover-submenu-abc">`

**How to fix:** 
Fix all of the following:
  Invalid ARIA attribute values: aria-owns="ic-popover-submenu-abc", aria-controls="ic-popover-submenu-abc"

====

## 🪜 How to reproduce 
1. Go to the ic-popover-menu storybook
2. Open up any of the popover menus
3. Open the popover menu so it is visible 
4. Run the FastPass 

## 🧐 Expected behaviour 
These errors should not be present. They need to be solved.

## Additional info 
[...]

PR #2718, opened 2024-11-06
1897 ic popover menu accessibility failures

GCHQ-Developer-299:
## Summary of the changes
Modified some semantics/aria inside popover menu to clear errors in Accessibility Insights. 

I've personally tested the popover in VoiceOver and NVDA (using the TextViewer in a virtual machine) and it reads the elements and their relationships okay in both.

## Related issue
#1897

### Accessibility 

- [x] Accessibility Insights FastPass performed. (this will still fail on popovers with toggle variant entries - have raised #2722)
- [x] A11y unit test added and yields no issues. (Cypress A11y tests are still disabled until #2722 is resolved)

### DIFF
--- a/packages/web-components/src/components/ic-menu-group/ic-menu-group.tsx
+++ b/packages/web-components/src/components/ic-menu-group/ic-menu-group.tsx
@@ -17,13 +17,13 @@ export class MenuGroup {
     const parentMenu = this.el.closest("ic-popover-menu");
 
     return (
-      <Host aria-label={this.label !== null ? this.label : ""}>
+      <Host role="group" aria-label={this.label !== null ? this.label : ""}>
         {isPropDefined(this.label) && (
           <ic-typography variant="subtitle-small">{this.label}</ic-typography>
         )}
-        <ul role="group">
+        <span class="menu-items-wrapper">
           <slot></slot>
-        </ul>
+        </span>
         {/* The line under the menu group is added on all menu groups except in the case that the menu group is the last item in the popover menu */}
         {this.el !== parentMenu?.querySelector("ic-menu-group:last-child") && (
           <hr />
--- a/packages/web-components/src/components/ic-menu-group/test/basic/__snapshots__/ic-menu-group.spec.ts.snap
+++ b/packages/web-components/src/components/ic-menu-group/test/basic/__snapshots__/ic-menu-group.spec.ts.snap
@@ -1,14 +1,14 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`menu group should render a menu group with a label 1`] = `
-<ic-menu-group aria-label="Menu item label" label="Menu item label">
+<ic-menu-group aria-label="Menu item label" label="Menu item label" role="group">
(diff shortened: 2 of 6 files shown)


packages/react/src/component-tests/IcPopoverMenu/IcPopoverMenuData.tsx
```diff
@@ -183,7 +183,6 @@ export const PopoverMenuWithVariants = () => {
             label="Edit"
             description="This is used to edit the original."
           />
-          <IcMenuItem label="Show/Hide" variant="toggle" />
```
GCHQ-Developer-112, line 186: What was the reason for removing this menu item?

packages/web-components/src/components/ic-menu-group/ic-menu-group.tsx
```diff
@@ -17,13 +17,13 @@ export class MenuGroup {
     const parentMenu = this.el.closest("ic-popover-menu");
 
     return (
-      <Host aria-label={this.label !== null ? this.label : ""}>
+      <Host role="group" aria-label={this.label !== null ? this.label : ""}>
         {isPropDefined(this.label) && (
           <ic-typography variant="subtitle-small">{this.label}</ic-typography>
         )}
-        <ul role="group">
+        <span>
```
GCHQ-Developer-112, line 24: Does moving the group role have a negative impact on accessibility? Menu items use the <li> tag and they need to be placed in either a group or a list

packages/react/src/component-tests/IcPopoverMenu/IcPopoverMenuData.tsx
GCHQ-Developer-299: https://github.com/mi6/ic-ui-kit/issues/2722 - nested interactive elements are bad practice (and a failure in Accessibility Insights)

GCHQ-Developer-299: I think i'm going to change this PR so that it doesn't remove the examples and leaves the work to that new ticket. As you've said on internal systems having role='menucheckboxitem' is allowed, we must just need to tweak our components

packages/web-components/src/components/ic-menu-group/ic-menu-group.tsx
GCHQ-Developer-299: Because of your comment I took a second look at the file and have refactored it a little bit more - previously the group role being nested on the ul was causing an error, having it on the host element of the ic-menu-group removes the error and functions fine in voiceover :) 
Screenshot 2024-11-08 at 11 10 41

packages/react/cypress-image-diff-screenshots/baseline/IcPopoverMenu.cy.tsx-menu-group-high-contrast.png
GCHQ-Developer-299, line 1: I've introduced a bug that currently prevents menu-groups with labels from having their first item focussed when the menu is opened

PR #2730, opened 2024-11-07
Feature/1214 toast dark mode

GCHQ-Developer-847:
## Summary of the changes
Added colour tokens for toast component

## Related issue
Part of #1214 

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes


GCHQ-Developer-847:
> Please correct me if I'm wrong, but I don't see any difference in toast in light and dark mode. If there's no difference, there shouldn't be a theme prop applied to toast as it has no effect. If all the tokens for toast are the same no matter the mode, then they should be moved to the colors.css file similar to classification banner

Ah ok makes sense - yep it's only the values of the tokens for the toast heading and message which change, but all others stay the same. I will remove the prop and move the other tokens into colors.css 👍

GCHQ-Developer-847:
> > Please correct me if I'm wrong, but I don't see any difference in toast in light and dark mode. If there's no difference, there shouldn't be a theme prop applied to toast as it has no effect. If all the tokens for toast are the same no matter the mode, then they should be moved to the colors.css file similar to classification banner
> 
> Ah ok makes sense - yep it's only the values of the tokens for the toast heading and message which change, but all others stay the same. I will remove the prop and move the other tokens into colors.css 👍

Actually ignore me about keeping those tokens in - I understand I can just change the token values for the heading and message to always be white. Will sort it

GCHQ-Developer-299 (dismissed):

GCHQ-Developer-112 (approved):

GCHQ-Developer-112 (approved):

PR #2731, opened 2024-11-07
chore(root): update token in cherry-pick workflow

MI6-255:
## Summary of the changes
update token in cherry-pick workflow to auth as a user, using a PAT rather than a GitHub Token, to bypass branch permissions using my user

## Related issue
NA


evenstensberg:
Good stuff! 💯

GCHQ-Developer-112 (approved):

PR #2760, opened 2024-11-12
fix(canary-web-components): skip adding slotted column placeholders when the slotted column already has data

### LINKED ISSUE #2743: [ic-data-table] Wrong columns appear when slotting in multiple columns
## Summary of the bug 

Example data structure:
```ts
    type DataRow = {
        created: number
        first: string
        middle: string
        last: string
        status: string
    }
```

When slotting in a custom element for created and status, the last name column disappears and has the contents of the status column:

image

If I remove the slotting it displays correctly, see:
image

## 🪜 How to reproduce 
1. View page
2. See last name data missing

## 📸 Screenshots or code 
If you can, add screenshots to show the problem. If you need to show a code example, you can include a snippet or link to a repository with an example. 

## 🖥 📱 Device 
- Type: [e.g. Desktop, mobile] 
- Device: [e.g. iPhone, MacBook, ThinkPad] 
- OS version: [e.g. macOS13, iOS16, Android 13] 
- Browser version: [e.g. Chrome 115, Safari 16] 

## 🧐 Expected behaviour 
Tell us, in enough detail, what you expected to happen instead. 

## 📝 Acceptance Criteria  

*Given* 
*When*  
*Then* 

## Additional info 
Tell us anything else useful to help us fix or understand the problem.

timja:
## Summary of the changes

There were a number of issues with how columns were laid out before with slotting this corrects them

~To take out of draft I want to take a look at:~
- [x] I think this may still be broken if data is not provided in the cells - fixed

I have tried to create a unit test for the helper but jest is only configured for a11y on this project and looks to be a pain to setup, I can try if deemed needed

## Related issue
fixes #2743

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes

### DIFF
--- a/packages/canary-web-components/src/components/ic-data-table/ic-data-table.tsx
+++ b/packages/canary-web-components/src/components/ic-data-table/ic-data-table.tsx
@@ -1505,7 +1505,12 @@ export class DataTable {
           ({ key }, index) =>
             isSlotUsed(this.el, `${key}-${rowIndex}`) && { key, index }
         )
-        .filter((col) => !!col);
+        .filter(
+          (col) =>
+            !!col &&
+            // skip the column if its already in the row
+            !Object.prototype.hasOwnProperty.call(row, col.key)
+        );
       return slottedColumns.length > 0
         ? addDataToPosition(row, slottedColumns, "")
         : row;
--- a/packages/canary-web-components/src/utils/helpers.ts
+++ b/packages/canary-web-components/src/utils/helpers.ts
@@ -664,15 +664,18 @@ export const addDataToPosition = (
 ): IcDataTableDataType => {
   const newData: IcDataTableDataType = {};
   const newIndexes = newKeys.map((key) => key.index);
-  let controlledIndex = 0; // When a new key is added to the data, need to increment the index to account for this new object value
 
-  Object.keys(dataObject).forEach((dataKey) => {
-    if (newIndexes.includes(controlledIndex)) {
-      newData[newKeys[newIndexes.indexOf(controlledIndex)].key] = newValue;
-      controlledIndex++;
+  const keys = Object.keys(dataObject);
+  const values = Object.values(dataObject);
+  const numberOfKeys = keys.length + newIndexes.length;
+
+  for (let i = 0, j = 0; i < numberOfKeys; i++) {
+    if (newIndexes.includes(i)) {
+      newData[newKeys[newIndexes.indexOf(i)].key] = newValue;
+      continue;
     }
-    newData[dataKey] = dataObject[dataKey];
-    controlledIndex++;
-  });
+    newData[keys[j]] = values[j];
+    j++;
+  }
   return newData;
 };


timja:
@GCHQ-Developer-741 would you be able to take a look as the author of the previous change please? https://github.com/mi6/ic-ui-kit/pull/2401

packages/canary-web-components/src/utils/helpers.ts
```diff
@@ -664,15 +664,18 @@ export const addDataToPosition = (
 ): IcDataTableDataType => {
   const newData: IcDataTableDataType = {};
   const newIndexes = newKeys.map((key) => key.index);
-  let controlledIndex = 0; // When a new key is added to the data, need to increment the index to account for this new object value
 
-  Object.keys(dataObject).forEach((dataKey) => {
-    if (newIndexes.includes(controlledIndex)) {
```
timja, line 670: this didn't work if e.g. new indexes were 0,1 as 1 would never be called as controlledIndex would skip over it.

Also didn't work for end columns if there were two slotted columns as the object only iterated over the existing keys and it finished iterating before it got to the end column

packages/canary-web-components/src/utils/helpers.ts
```diff
@@ -664,15 +664,18 @@ export const addDataToPosition = (
...
-      newData[newKeys[newIndexes.indexOf(controlledIndex)].key] = newValue;
-      controlledIndex++;
+  const keys = Object.keys(dataObject);
+  const values = Object.values(dataObject);
+  const numberOfKeys = keys.length + newIndexes.length;
```
timja, line 670: we now iterate over the expected amount of keys total so we don't skip keys like previously

packages/canary-web-components/src/utils/helpers.ts
```diff
@@ -664,15 +664,18 @@ export const addDataToPosition = (
...
+
+  for (let i = 0, j = 0; i < numberOfKeys; i++) {
+    if (newIndexes.includes(i)) {
+      newData[newKeys[newIndexes.indexOf(i)].key] = newValue;
+    } else {
+      newData[keys[j]] = values[j];
+      j++;
```
timja, line 677: the position in the original object that we're fetching from, we can't use i as i includes the new positions

packages/canary-web-components/src/utils/helpers.ts
```diff
@@ -664,15 +664,18 @@ export const addDataToPosition = (
 ): IcDataTableDataType => {
   const newData: IcDataTableDataType = {};
   const newIndexes = newKeys.map((key) => key.index);
-  let controlledIndex = 0; // When a new key is added to the data, need to increment the index to account for this new object value
 
-  Object.keys(dataObject).forEach((dataKey) => {
-    if (newIndexes.includes(controlledIndex)) {
-      newData[newKeys[newIndexes.indexOf(controlledIndex)].key] = newValue;
-      controlledIndex++;
+  const keys = Object.keys(dataObject);
```
evenstensberg, line 668: You should use Object.entries to get the keys and values from one expression. Use destructuring to assign them to the key, value variables. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries

packages/canary-web-components/src/components/ic-data-table/ic-data-table.tsx
```diff
@@ -1505,7 +1505,12 @@ export class DataTable {
           ({ key }, index) =>
             isSlotUsed(this.el, `${key}-${rowIndex}`) && { key, index }
         )
-        .filter((col) => !!col);
+        .filter(
+          (col) =>
+            !!col &&
+            // skip the column if its already in the row
+            !Object.prototype.hasOwnProperty.call(row, col.key)
```
evenstensberg, line 1512: I'm not fan of prototype calls vs methods that already exists on the Object, but it's just a nitpick

packages/canary-web-components/src/utils/helpers.ts
```diff
@@ -664,15 +664,18 @@ export const addDataToPosition = (
...
+  const values = Object.values(dataObject);
+  const numberOfKeys = keys.length + newIndexes.length;
+
+  for (let i = 0, j = 0; i < numberOfKeys; i++) {
+    if (newIndexes.includes(i)) {
+      newData[newKeys[newIndexes.indexOf(i)].key] = newValue;
```
evenstensberg, line 674: add continue inside the if statement, and remove the else block. Will make the loop neat.

evenstensberg:
@timja commit should look like this: `fix(canary-web-components): skip adding slotted column placeholders when the slotted column already has data`

packages/canary-web-components/src/components/ic-data-table/ic-data-table.tsx
timja: eslint disagrees: https://eslint.org/docs/latest/rules/no-prototype-builtins#rule-details

evenstensberg: Okay 👍🏾

packages/canary-web-components/src/utils/helpers.ts
timja: I tried that and wasn't able to get it to work, can you provide a GitHub suggestion?

GCHQ-Developer-741 (dismissed):
Nice work. Looks good, thanks for working on this

packages/canary-web-components/src/utils/helpers.ts
evenstensberg: You'll need to use map to normalize the array. But that's more work than it should be so this is ok

timja:
@GCHQ-Developer-299 would you be able to take a look at this please?

timja:
/ @GCHQ-Developer-741 could you re-approve please?

GCHQ-Developer-741 (approved):

GCHQ-Developer-112 (approved):

PR #2786, opened 2024-11-18
feat(web-components): side-nav tokens

gd2910:
added color-mode tokens for side-navigation

BREAKING CHANGE: dark mode


packages/web-components/src/global/color-mode.css
```diff
@@ -605,6 +605,28 @@ ic-theme,
   --ic-skeleton-background: var(--ic-architectural-100);
   --ic-skeleton-background-secondary: var(--ic-architectural-200);
 
+  /* Side navigation  */
```
GCHQ-Developer-530, line 608: Since it's the same tokens for light and dark, these tokens can be moved into colors.css like we've done for classification banner, hero, toast etc.

GCHQ-Developer-530 (changes requested):

packages/react/cypress-image-diff-screenshots/baseline/IcSideNavigation.cy.tsx-basic-open.png
GCHQ-Developer-530, line 1: It looks like the little divider to the left of the app icon and title has the colour swapped with the lighter blue variant of side nav

packages/web-components/src/global/color-mode.css
```diff
@@ -668,6 +668,28 @@ ic-theme,
...
```
GCHQ-Developer-741, line 671: If these have been moved to colors.css, do they still need to be in this file? None of the others like this (Toast, Hero, etc.) are in here.

ad9242 (dismissed):

GCHQ-Developer-741 (approved):

GCHQ-Developer-847 (approved):

PR #2840, opened 2024-11-28
Fix/update top nav tokens

GCHQ-Developer-847:
## Summary of the changes
As discussed with design, updated three of the top nav tokens (keyline, title-hover and title-pressed) so they are the same colour in light and dark mode

## Related issue
N/A


packages/react/cypress-image-diff-screenshots/baseline/IcTopNavigation.cy.tsx-theme-dark.png
gd2910, line 1: Should ic-search-bar be dark variant also?

GCHQ-Developer-741 (approved):

packages/react/cypress-image-diff-screenshots/baseline/IcTopNavigation.cy.tsx-theme-dark.png
GCHQ-Developer-847: I don't think so because the top nav in this visual test has been set to dark mode using the `theme` prop, whereas the reason the search bar in Storybook is dark is because the top nav is within a `<theme>` set to dark mode. Setting the `theme` prop on the top nav has no effect on the search bar

gd2910 (approved):

Issue #2390, opened 2024-09-06
[ic-select] - able to select "no options found" entry 


ad9242:
## Summary of the bug 
With an ic-select with no options set, it is possible to select the "No options found" entry in the popup menu  & the value is then set in the input field:

## 🪜 How to reproduce 
Tell us the steps to reproduce the problem: 
1. Go to this [stackblitz example](https://stackblitz.com/edit/r3peqe-y2uhjg?file=src%2Fapp.tsx&hideExplorer=1)
2. Click on the ic-select 
3. Click on "no options found" option 
4. See the error - the value is updated

## 📸 Screenshots or code 

![image](https://github.com/user-attachments/assets/e93caf3b-0326-45b6-9d53-87f04242b537)

## 🧐 Expected behaviour 
the option should not be selectable

## 📝 Acceptance Criteria  
If relevant, describe in full detail the different interactions and edge cases that the component or patterns needs to fulfil. 

*Given* an ic-select with no options
*When*  the "no options found" entry is selected
*Then* the value in the input field should not update

ad9242:
only affects non-searchable

PR #2870, opened 2024-12-06
fix(web-components): select bugfix

gd2910:
when no options available the 'no options found' item cannot be selected

## Summary of the changes
Can't have `cursor event` and `pointer` CSS changes on the same element, so I created a `<span>` to wrap the li option on a different layer, only when options not available, not when 'retry' is available.

## Related issue
#2390

### DIFF
--- a/packages/react/src/component-tests/IcSelect/IcSelect.cy.tsx
+++ b/packages/react/src/component-tests/IcSelect/IcSelect.cy.tsx
@@ -1573,6 +1573,19 @@ describe("IcSelect end-to-end, visual regression and a11y tests", () => {
       .should(HAVE_LENGTH, "12");
   });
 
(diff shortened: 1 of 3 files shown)


ad9242:
I think it looks a little bit confusing now, as it looks like you can click\select the "Loading error", as it has the cursor:pointer styling. maybe need to consider a solution where it just doesn't update the value?

GCHQ-Developer-112:
Please make the commit message summary more descriptive of the change made

ad9242 (approved):

ad9242:
Approved, but raised new issue (not introduced by this change) https://github.com/mi6/ic-ui-kit/issues/2959

ad9242 (approved):

GCHQ-Developer-847 (approved):

Issue #1960, opened 2024-06-06
[ic-text-field]: Autofocus prop not working unless text field label is hidden.


jd239:
## Summary of the bug 
Setting autofocus on `ic-text-field` does not focus on the input on first load if the input label is displayed. Auto focus only works if the label is hidden.

## 🪜 How to reproduce 
Tell us the steps to reproduce the problem: 
1. Load up storybook
2. Add a basic `ic-text-field` with the `auto-focus` attribute

## 📸 Screenshots or code 
React variant below. On load, the `hideLabel` and `autoFocus` props are present. Remove `hideLabel` to see issue. 
(The browser window needs to be in full screen to see the issue)

## 🖥 📱 Device 
- Type: Desktop
- Device: N/A
- OS version: Windows 
- Browser version: Chrome

## 🧐 Expected behaviour 
ic-text-field should have an active cursor and the focus state

## 📝 Acceptance Criteria  

*Given the text field has been added to a page with the auto focus attribute added* 
*When the page first loads*  
*Then the text field should have the focus indicator* 

## Additional info 
Console logging `activeElement` within the component returns `null`.
Using `delegatesFocus` doesn't seem to make a difference

MI6-255:
Works in the test app

gd2910:
After further testing this is not an issue in the react test-app.

ad9242:
text-field may need to be in a form for this to work - needs further investigation

PR #2901, opened 2024-12-13
1960 ic text field autofocus prop not working unless text field label is hidden

gd2910:
## Summary of the changes
`autofocus` was only being applied when the label was hidden, this was due to the child components rendering after the `ic-text-field` components.

Changes;
- `setInterval` waits until typography is hydrated before setting focus on input
- `autofocus` renamed from `auto-focus` to align more closely with native attribute

## Related issue
#1960


packages/web-components/src/components/ic-text-field/ic-text-field.tsx
```diff
@@ -596,7 +609,7 @@ export class TextField {
             {!multiline ? (
               <input
                 id={inputId}
-                name={name}
+                name={this.name}
```
ad9242, line 612: is this change needed? isn't name already accessible from the object restructuring in the render method (line 474)?

ad9242 (changes requested):

GCHQ-Developer-847 (changes requested):
I think it would be worth updating the info about the breaking changes in the commit messages to be even clearer about what the change actually is

GCHQ-Developer-847 (approved):

ad9242 (approved):

PR #2930, opened 2024-12-18
Removed `disabled` attribute from all components when set to false

GCHQ-Developer-741:
## Summary of the changes
Following changes to certain components to ensure the disabled attribute is fully removed from the component, replicated across all components that have a disabled attribute to ensure consistency and correct behaviour with screen readers.

## Related issue
#2629 

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes

### DIFF
--- a/packages/canary-web-components/src/components/ic-card-horizontal/ic-card-horizontal.tsx
+++ b/packages/canary-web-components/src/components/ic-card-horizontal/ic-card-horizontal.tsx
@@ -7,6 +7,7 @@ import {
   h,
   Method,
   forceUpdate,
+  Watch,
 } from "@stencil/core";
 import {
   onComponentRequiredPropUndefined,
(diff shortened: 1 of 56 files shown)


ad9242:
looks good, but do you think it would be good to add unit tests that specifically set it to `disabled=false`, so we can check that it's actually removing it from the element (by comparing with dom snapshot)

ad9242 (changes requested):
see comment re adding unit tests

GCHQ-Developer-741:
Have added the unit tests (except for the ic-data-input/picker because I couldn't get the snapshots generated)

GCHQ-Developer-847 (changes requested):
I think the select is missing a watch on its `disabled` prop and a unit test, but all looks good apart from that

I don't have NVDA so can't test with that, but since this solution was already working on some components, I assume it's ok - so will be happy to approve once the above issue is resolved :)

GCHQ-Developer-847 (approved):

GCHQ-Developer-299 (approved):
Tested on the Accordion and Select Playground stories - by toggling on/off 'disabled'. On Develop, the accordion would still read as 'unavailable' after it was undisabled, on your branch it correctly did not :)

ad9242 (approved):

Issue #2906, opened 2024-12-16
[ic-top-navigation] Screen reader hint causes massive scrollable blank area to right of page when using RTL language


irl:
## Summary of the bug 

This is basically the same issue as #2905 except affecting the ic-top-navigation component.

The offending element is `#navigation-landmark-button-text`.

## 🪜 How to reproduce 
Tell us the steps to reproduce the problem: 

1.    Set `dir="rtl"` on `<html>`
2.    Add an ic-top-navigation
3. Reduce the window size until the "Menu" button appears
4.    Observe a blank space appear when you scroll to the left

## 🖥 📱 Device 

-    Type: desktop
-    Device: mac os
-    Browser version: Firefox 133

## 🧐 Expected behaviour 

The blank space should not appear.

## 📝 Acceptance Criteria  
If relevant, describe in full detail the different interactions and edge cases that the component or patterns needs to fulfil. 

*Given*  Set `dir="rtl"` on `<html>`, window size is below breakpoint for Menu button to appear
*When*  ic-top-navigation is used
*Then* no blank space appears to the left

## Additional info 

As with #2905, this can be fixed with the [:dir](https://developer.mozilla.org/en-US/docs/Web/CSS/:dir) selector, swapping the `left:` for `right:`.

GCHQ-Developer-299:
Note: pushed to future release as we don't official support RTL languages - however, if external contributor addresses issue beforehand, that's great

PR #2953, opened 2024-12-20
feat(web-components): ensure screen-reader-only text is hidden for RTL languages

irl:
## Summary of the changes

When using RTL languages, elements using `left: -9999px` and roughly equivalent styles will cause a large scrollable but blank region to appear on the left of the screen. By using the `:dir` pseudo-class it is possible to instead shift these elements to be `right: -9999px` which the browser will hide when `<html dir="rtl">` in the same way as it does for elements to the left when `<html dir="ltr">`.

This change also standardises on using -9999px from the start of the page as the offscreen position to make it easier to locate occurrences of this pattern with search. I also found -10000px and -625rem but these are roughly or exactly equivalent.

There is a question as to whether this is the correct change, or if the correct change is to have just one class that can be applied to these elements. Seeing multiple classes defined as:

* .sr-only
* .screen-reader-only-text
* .offscreen
* .visually-hidden

feels redundant to me. The technique is also applied directly to component-specific classes instead of defining the styles just once for the technique and re-using that class where it is needed.

## Related issue

Closes: #2906

### General 

- [ ] Changes to docs package checked and committed.

### Testing

- [ ] Visual testing against Figma component specification completed. 
- [ ] Compare performance of modified components against develop using Performance addon in React Storybook.

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes

- [ ] Tested for FOUC (Flash of Unstyled Content) in both SSR (Server-Side Rendering) and SSG (Static Site Generation) settings.
- [ ] Controlled and uncontrolled input components tested.
- [ ] Props/slots can be updated after initial render.

### DIFF
--- a/packages/web-components/src/components/ic-badge/ic-badge.css
+++ b/packages/web-components/src/components/ic-badge/ic-badge.css
@@ -158,6 +158,10 @@
   left: -9999px;
 }
 
+.sr-only:dir(rtl) {
+  right: -9999px;
+}
+
 @keyframes expand {
   from {
     opacity: 0;
--- a/packages/web-components/src/components/ic-checkbox-group/ic-checkbox-group.css
+++ b/packages/web-components/src/components/ic-checkbox-group/ic-checkbox-group.css
@@ -39,3 +39,7 @@ ic-input-validation {
   color: #000;
   text-transform: none;
 }
+
+.screen-reader-only-text:dir(rtl) {
+  right: -9999px;
+}
--- a/packages/web-components/src/components/ic-classification-banner/ic-classification-banner.css
+++ b/packages/web-components/src/components/ic-classification-banner/ic-classification-banner.css
@@ -56,6 +56,10 @@
   text-transform: none;
 }
 
+.offscreen:dir(rtl) {
+  right: -9999px;
+}
+
 @media (forced-colors: active) {
   .classification-banner {
     border: var(--ic-hc-border);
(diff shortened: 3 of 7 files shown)


GCHQ-Developer-847:
Hi @irl, thank you for your contribution! 

I agree that it is unnecessary / redundant to use different classes across components. If we were to just use one shared class, I think the [`helpers.css` file](https://github.com/mi6/ic-ui-kit/blob/main/packages/web-components/src/global/helpers.css) would be the place to add it. 

However, I feel as though implementing the use of a single class may be a bit more involved as the screen-reader-only classes currently applied to individual components have slightly different styles applied; it would require checking that the styles on a single shared class work for all components, alongside some refactoring.

I’d say it’s up to you if you want to fix the redundancy in this PR - but it is a bit out of scope for the ticket so we’d be happy if you want to create a separate ticket for it (or let us know if you want us to create one!). 

<br />
(Also, for the checklist on the PR, you can mark any task as done if it isn’t relevant or related to your changes. They will need to all be ticked before the PR can be merged.)

irl:
> Hi @irl, just wondered if you are planning to update this PR soon? There are at least a few conflicts and one comment which needs resolving

Yes, thanks for the prompt. I do want to get this done soon.

irl:
Sorry for the delay. This is rebased on the latest develop, with the missing selector added. Still leaving the larger consolidation to a single sr-only class for later but this resolves the immediate issue.

GCHQ-Developer-741 (approved):

GCHQ-Developer-847 (approved):

GCHQ-Developer-741 (approved):

PR #2956, opened 2024-12-23
Add React types to fix build issue

dn55533:
## Summary of the changes
Fixed build problem in web-components by adding dependency on React types.

The npm run build task was failing with the following error: Cannot find namespace 'JSX'

This has been fixed by adding dev dependencies on the following packages:

- @types/react
- @types/react-dom

The versions of these dependencies being used ("^17.0.37") match those being used in the canary-react package.

## Related issue
N/A

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes

### DIFF
--- a/packages/web-components/package-lock.json
+++ b/packages/web-components/package-lock.json
@@ -30,6 +30,8 @@
         "@types/autoprefixer": "^10.2.0",
         "@types/jest": "^26.0.24",
         "@types/jest-axe": "^3.5.3",
+        "@types/react": "^17.0.37",
+        "@types/react-dom": "^17.0.11",
         "autoprefixer": "^10.4.2",
         "babel-loader": "^8.2.3",
         "jest": "^26.6.3",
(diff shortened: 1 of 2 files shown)


ad9242:
It shouldn't need React specific packages adding at the root level. If anywhere, we'd expect them to be needed in react & canary-react packages (in theory only for running storybook or the cypress tests, where the react components are used)

What was the reason for adding them at this level? Doing the following in the root folder 
```
npm install
npm run bootstrap
npm run build
```

does not seem to cause any issues, so it would be good to get to the bottom of if this is something environmental before adding additional dependencies for everyone to the root level

Also, if they were to be added, ideally they would not need to be locked to a specific version and could be locked to a minimum version as with all other dependencies

dn55533:
WIthout these dependencies, the project fails to build using those three commands. I tried multiple times, including cloning the repo from scratch into a new folder.

When I added those dependencies, it fixed the problem.

I'm not sure where the best place to put these dependencies, so I went with the root level. Happy to move them if needed.

ad9242:
Interestingly, where it shows the error, it seems to be using Stencil version 4.23.0
@ukic/web-components: [40:28.3] @stencil/core
@ukic/web-components: [40:29.0] v4.23.0 
@ukic/web-components: [40:34.2] build, core, prod mode, started ...

When we follow the steps you mention, it installs Stencil v4.10.0 as set in the [package-lock.json](https://github.com/mi6/ic-ui-kit/blob/dee234161a67a9b42a20f4bc385a05da65b2cf4d/packages/web-components/package-lock.json#L4996)

Do you know of anything in your setup that could be causing the newer version of Stencil to be installed, as we are struggling to recreate this?

MI6-255 (changes requested):
Currently waiting on a discussion internally, marking as requested changes to prevent a merge.

GCHQ-Developer-112:
> Hi @dn55533, I've been able to recreate the issue you were facing. Please can you install the two types packages at version 17 so they match the versions of the `react` and `react-dom` dev dependencies that are already installed. Please can you also do the install in the `web-components` folder instead of the root folder

One more thing, please refer to our [CONTRIBUTING.md](https://github.com/mi6/ic-ui-kit/blob/main/CONTRIBUTING.md#git-commit) for guidance on how to format your commit

GCHQ-Developer-112 (approved):

GCHQ-Developer-112 (changes requested):
Please also commit the updated `package-lock.json` in the `web-components` folder

GCHQ-Developer-112 (approved):

PR #2961, opened 2024-12-31
refactor(web-components): fixes console warning for ic-select

ad9242:
## Summary of the changes
Fixes console warning about prop changing in componentDidLoad, that was making it hard to see genuine errors when running unit tests

## Related issue
N/A


packages/web-components/src/components/ic-select/ic-select.tsx
```diff
@@ -387,6 +387,8 @@ export class Select {
       this.setDefaultValue();
       this.uniqueOptions = this.deduplicateOptions(this.options);
     }
+
+    this.hiddenInputValue = this.searchable && (this.currValue as string);
```
evenstensberg, line 391: what is the type of currValue? Using "as" is an anti-pattern.

PR #2992, opened 2025-01-08
Fix/1890 1 searchable select clear on edit

GCHQ-Developer-847:
## Summary of the changes
Updated searchable select to deselect the selected option when input is edited.

1. Select an option
2. Edit the text in the input field
3. See that the previously selected option now becomes deselected. icChange is emitted with value `null` and the hidden input value is set to `null`

-----

(**Note:** I would recommend looking at the console logs in the Searchable > Default story to see the changes to the events which are explained in more detail below)

This PR also includes changes to the icChange and icInput events. IcChange now only returns either `null` or the value of the selected option, whereas before it gave you the characters entered into the input field (same behaviour as the icInput event). 

I felt that this change made sense as icChange now more closely matches its behaviour when on the single select, and the value returned via icChange now better reflects the actual value of the searchable select (e.g. the value is essentially cleared when the input is edited after selecting an option).

I have also changed how the `debounce` prop works, in that it now applies to the icInput event instead of icChange, since the icChange event now doesn't give you the actual characters which are being typed.

## Related issue
For the first point mentioned in #1890. Another PR will be opened to fix the second issue.

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes


packages/web-components/src/components/ic-select/ic-select.tsx
```diff
@@ -305,7 +305,8 @@ export class Select {
       }
     }
 
-    if (this.searchable) {
+    if (this.searchable && this.value) {
+      // Only set if value not null - prevents whole input value being cleared when edited
       this.searchableSelectInputValue =
         this.getLabelFromValue(this.currValue as string) ||
```
evenstensberg, line 311: using `as <type>` is an anti-pattern, maybe just construct a string? I think the warnings will ignore that, then.
 
```js
this.getLabelFromValue(new String(this.currValue))
```

GCHQ-Developer-847: I was just wondering why this is anti-pattern / whether you have any references which explain why `as <type>` shouldn't be used? We have used it many times throughout our codebase, and I have seen it used in a lot of other code online

Also where are the warnings you are seeing? :)

evenstensberg: We sometimes used `as <type>` in webpack because the function parameter had a different type than the one passed as an argument.

Without `as <type>` it made the IDE (or CLI) typechecking emit an error, so that's why I assumed that you used (`as string`). 

If the expected argument passed to a function differs than the one being passed (you override the typechecking), it seems more like you're making typescript not typecheck that specific code, if that makes sense?

GCHQ-Developer-847: I've found that [some sources](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html) seem to recommend against using `String`

evenstensberg: Try without the `new` keyword.

GCHQ-Developer-847: That worked - thanks!

As for my previous comment about it not being recommended, I have looked at some other sources and opinions seem to be mixed, so happy to just go with it especially as we have sometimes used `String` elsewhere in our code

evenstensberg: Good stuff 🤓

packages/web-components/src/components/ic-select/ic-select.tsx
```diff
@@ -470,26 +475,23 @@ export class Select {
...
-    if (!this.searchable) {
-      // If "Select all" button clicked, replace value with new value (array of all option values)
-      if (this.multiple && !Array.isArray(value) && value !== null) {
-        this.handleMultipleSelectChange(value as string);
-      } else {
-        this.value = value;
-      }
+    // If "Select all" button clicked, replace value with new value (array of all option values)
+    if (this.multiple && !Array.isArray(value) && value !== null) {
+      this.handleMultipleSelectChange(value as string);
```
evenstensberg, line 480: you'll need to wrap a string around this too

GCHQ-Developer-299 (approved):

ad9242:
In the [external filtering example](https://mi6.github.io/ic-ui-kit/branches/fix/1890-1-searchable-select-clear-on-edit/react/?path=/story/react-components-select-searchable--external-filtering), it briefly now shoes the loading error state before the results appear: 

image

Is that correct? 

it didn't seem to do this previously: https://mi6.github.io/ic-ui-kit/branches/develop/react/?path=/story/react-components-select-searchable--external-filtering

GCHQ-Developer-847:
> In the [external filtering example](https://mi6.github.io/ic-ui-kit/branches/fix/1890-1-searchable-select-clear-on-edit/react/?path=/story/react-components-select-searchable--external-filtering), it briefly now shoes the loading error state before the results appear:
> 
> image
> 
> Is that correct?
> 
> it didn't seem to do this previously: https://mi6.github.io/ic-ui-kit/branches/develop/react/?path=/story/react-components-select-searchable--external-filtering

<br />

Having looked into this, I believe it’s actually the code for the External Filtering story in Storybook which is causing this difference in behaviour. 

In the v3 develop branch, when you type into the select, `handleInput` is called and then when it gets to [line 542](https://github.com/mi6/ic-ui-kit/blob/d74b2a9c03d97474286770f5124cdd0ad7e14eff/packages/react/src/stories/ic-select_(searchable).stories.mdx?plain=1#L542), the timeout is set to 500 because it was the [`icChange` event which called `handleInput`](https://github.com/mi6/ic-ui-kit/blob/v3.0.0/develop/packages/react/src/stories/ic-select_(searchable).stories.mdx?plain=1#L572). This means that the loading error state is not shown at all because it doesn’t get set to 2000 which would be greater than the `timeout` prop value.

Now that I have changed [line 572](https://github.com/mi6/ic-ui-kit/blob/v3.0.0/develop/packages/react/src/stories/ic-select_(searchable).stories.mdx?plain=1#L572) in this PR to `onIcInput`, the loading error state does appear because line 542 now gets set to 2000 when you type in the select.

I can get the behaviour which is happening in this PR by checking out onto the v3 develop branch (or v2 develop) and changing line 542 to `event.type === “icChange”` (what it should have been to get the error state showing) - as shown in the video below:

https://github.com/user-attachments/assets/44447f3c-53e2-4ef0-9bc8-89708c16122b

GCHQ-Developer-847:
> In the [external filtering example](https://mi6.github.io/ic-ui-kit/branches/fix/1890-1-searchable-select-clear-on-edit/react/?path=/story/react-components-select-searchable--external-filtering), it briefly now shoes the loading error state before the results appear:
>
> ...

… Basically I don’t think it is my changes to `ic-select.tsx` which are causing the issue - `icInput` is now producing the same behaviour as `icChange` used to, as expected. That being said, it is still a bit strange that it briefly shows the loading error state, especially as it works fine in the web components Storybook. I can’t quite work out why it is happening; as far as I can tell it’s basically just the same code but using React.   

I am going to open a ticket to make sure we look into it at some point - of course we definitely want it work as it should in the React Storybook to help with testing changes!

PR #3006, opened 2025-01-10
Fix/1890 2 searchable clear input on blur

GCHQ-Developer-847:
## Summary of the changes
Updated searchable select to clear input text on blur (when no option is selected).

Also had to make some changes to some existing Cypress tests - `.click()` was for some reason triggering onBlur (and therefore now clearing the input), so have either changed this to `.trigger("click")` or removed the clicks where they were unnecessary. (All seems to work fine on Storybook - this blurring does not happen). Tried a lot of different solutions and this was the only thing that worked.

## Related issue
Second issue from ticket #1890 

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes


packages/web-components/src/components/ic-select/ic-select.tsx
```diff
@@ -305,9 +305,10 @@ export class Select {
       }
     }
 
-    if (this.searchable) {
+    if (this.searchable && this.value) {
+      // Only set if value not null - prevents whole input value being cleared when edited
       this.searchableSelectInputValue =
-        this.getLabelFromValue(this.currValue as string) ||
+        this.getLabelFromValue(String(this.currValue)) ||
         (this.currValue as string);
```
evenstensberg, line 312: Why's there two declaration for one string here? 🤔

GCHQ-Developer-847: Turns out that this code wasn't quite redundant - it was actually part of an issue causing undesired behaviour where the `value` prop of the select can be set a value which isn't valid i.e. doesn't exist in the `options` array.

I have put the code back in and raised a separate ticket for this (#3013) because sorting it is a bit more involved than just this one line change

GCHQ-Developer-299 (approved):
Tested with mouse and with keyboard and the field now clears if no selection is made, happy to approve

GCHQ-Developer-741 (approved):

Issue #1916, opened 2024-05-23
[ic-tooltip] Disable hover prop impacting screen reader functionality


GCHQ-Developer-530:
## Summary of the bug 
When testing tooltip in voice over, all examples are read correctly with the tooltip text when the button is focused, except disable hover. 

When I click on the button, it just reads the button info "show tooltip show tooltip, button" and doesn't read out any tooltip label information.

Haven't tested this in NVDA so don't know if it's happening across all screen readers.

## 🪜 How to reproduce 
Tell us the steps to reproduce the problem: 
1. Go to disable hover storybook example
2. Click on the button with voice over on
3. See/hear the error 

## 🧐 Expected behaviour 
It should read the tooltip information on click like it does on hover for all the other examples.

BluePandaCard:
This issue is also happening in VO for me for buttons with regular tooltips (one's that trigger on hover) when you tab to them. 

<img width="837" alt="Image" src="https://github.com/user-attachments/assets/7adeb8ec-8377-4015-b4d6-9db2d8941b75" />

PR #3014, opened 2025-01-14
1916 tooltip disable hover

BluePandaCard:
## Summary of the changes
Update disabled hover tooltip story to use aria labelledby so that it reads out tooltip information

## Related issue
#1916 

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes

### DIFF
--- a/packages/web-components/src/components/ic-tooltip/ic-tooltip.stories.mdx
+++ b/packages/web-components/src/components/ic-tooltip/ic-tooltip.stories.mdx
@@ -76,7 +76,7 @@ import readme from "./readme.md";
         disable-hover="true"
         ><button
           id="test-button-disable-hover"
-          aria-describedby="ic-tooltip-test-button-disable-hover"
+          aria-labelledby="ic-tooltip-test-button-disable-hover"
           onclick="navigator.clipboard.writeText('Code')"
         >
           Show tooltip


packages/web-components/src/components/ic-tooltip/ic-tooltip.css
```diff
@@ -128,6 +128,11 @@
   ) !important;
 }
 
+.sr-only {
+  position: absolute;
+  left: -9999px;
```
evenstensberg, line 133: Maybe add a negative z index too? Just to double verify

BluePandaCard: The other sr-only classes on other components seem to only have position and left applied e.g. pagination/badge/date input/date picker.

evenstensberg: The change is ok as it is, just a small suggestion.


(As the ic-ui-kit or 2nd-3rd party code increases, z indices are usually included on hidden or stacked components)

GCHQ-Developer-741 (approved):

ad9242:
There seems to be an issue introduced in NVDA - it now reads out the tooltip twice:
image

Prior to this change, it was already read out once in NVDA:
image

BluePandaCard:
> There seems to be an issue introduced in NVDA - it now reads out the tooltip twice: image
> 
> Prior to this change, it was already read out once in NVDA: image
> 
> this is onclick, or when pressing enter on the button

Looks like just swapping aria describedby for aria labelledby works for voice over. Would you be able to check for NVDA?

ad9242:
It still reads it out twice I'm afraid. I suppose on balance having it work now in VoiceOver & read twice in NVDA is better than not reading it at all

ad9242:
@BluePandaCard please can you rebase your branch? We have had to regenerate all visual regression images due to a change in GH action runners using an updated OS version.

ad9242 (approved):

GCHQ-Developer-741 (approved):

Issue #2194, opened 2024-07-19
Focus state issues with breadcrumbs using slotted links in React


GCHQ-Developer-847:
## Summary of the bug 
When using slotted links in breadcrumbs, the current page is focusable and briefly shows the native focus indicator. In Safari, it shows a black underline.

## 🪜 How to reproduce 
1. Go to the [Slotted links](https://mi6.github.io/ic-ui-kit/branches/develop/react/?path=/story/react-components-breadcrumb--slotted-links) or [React Router](https://mi6.github.io/ic-ui-kit/branches/develop/react/?path=/story/react-components-breadcrumb--react-router) story in the React Storybook.
2. Tab through the links.
3. See the following:
- The breadcrumb for the current page is focusable in both Chrome and Safari. 
- The current page breadcrumb shows an outline around it briefly when it is focussed, and it shows a black underline on Safari.
- On Safari, it also moves a bit when you tab off it.
- If you focus on the current page, and press Enter, it actually works as a link, taking you to that page (technically refreshing the current page you are on).
- On Safari, when tabbing through or hovering on the breadcrumbs, the underline disappears.

## 📸 Screenshots or code 
On Chrome:

https://github.com/user-attachments/assets/57eb90a8-8bc2-486e-8e97-ad4a612bf478

On Safari:

https://github.com/user-attachments/assets/1c252794-c8a0-40b1-8fdd-4b2cf2e972b0

## 🖥 📱 Device 
- Browser: Chrome and Safari

## 🧐 Expected behaviour 
The current page should not be focusable at all as you are already on that page (see the breadcrumbs which don't have slotted links).

## 📝 Acceptance Criteria  
*Given* that there are breadcrumbs on the screen which are using slotted links
*When* I tab through the breadcrumbs
*Then* the current page breadcrumb should not be focusable at all.

GCHQ-Developer-847:
Related to #2196

PR #3050, opened 2025-01-16
Fix/2194 breadcrumb slotted link focus issues

GCHQ-Developer-847:
## Summary of the changes
Updated breadcrumb component to prevent focus on current page breadcrumb when it has a slotted link. Added slotted links to current page breadcrumbs in web components Slotted Links story to help with testing.

Go to the Slotted Links story (either web components or React) to test this.

-------

I have added functionality which checks for and removes hrefs from any slotted element in the default slot i.e. the link slot. I originally just added a -1 tabindex but the VO virtual cursor still read out the current page as a 'link'. Removing the hrefs stops this happening. 

However, I'm a bit concerned that this might not work for all types of slotted links e.g. links from some routing libraries may be structured / work differently. The tabindex will prevent focus at least, but I have also opened a [PR to add guidance](https://github.com/mi6/ic-design-system/pull/1322) recommending to not slot a link in the current page breadcrumb. But hopefully these changes prevent most issues if someone does use a slotted link.

## Related issue
#2194 

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes

### DIFF
--- a/packages/web-components/src/components/ic-breadcrumb/ic-breadcrumb.css
+++ b/packages/web-components/src/components/ic-breadcrumb/ic-breadcrumb.css
@@ -48,6 +48,10 @@ ic-link {
   --breadcrumb-link-gap: var(--ic-space-xs);
 }
 
+:host([aria-current="page"]) {
+  cursor: text;
+}
+
 :host([aria-current="page"]) .current-page-container {
   display: flex;
   align-items: center;
@@ -57,6 +61,10 @@ ic-link {
 
 :host([aria-current="page"].ic-breadcrumb-monochrome) .current-page-container {
   color: var(--ic-breadcrumb-text-monochrome);
+}
+
+:host([aria-current="page"]) slot {
+  pointer-events: none;
 }
 
 :host(.collapsed-breadcrumb-wrapper) {
--- a/packages/react/src/component-tests/IcBreadcrumb/IcBreadcrumb.cy.tsx
+++ b/packages/react/src/component-tests/IcBreadcrumb/IcBreadcrumb.cy.tsx
@@ -11,6 +11,7 @@ import {
   Appearance,
   ToggleCollapsed,
   ToggleBackBreadcrumb,
+  SlottedLinks,
 } from "./IcBreadcrumbTestData";
 import {
   BE_VISIBLE,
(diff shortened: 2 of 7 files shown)


GCHQ-Developer-847:
Also I have not added unit tests - `assignedElements()` isn't available in Jest / Stencil tests (see [discussion](https://github.com/ionic-team/stencil/issues/2830)) and therefore `getSlotElements` doesn't work. I tried loads of things but no success 😅 

I think the Cypress test basically covers how the functionality should work though

GCHQ-Developer-299 (dismissed):

GCHQ-Developer-741:
Have noticed that on the React Router story, selecting a different breadcrumb means that the previously selected one is not focusable, it doesn't seem to be adding the href property back in

packages/web-components/src/components/ic-breadcrumb/ic-breadcrumb.css
```diff
@@ -59,6 +59,10 @@ ic-link {
   color: var(--ic-breadcrumb-text-monochrome);
 }
 
+:host([aria-current="page"]) slot {
```
GCHQ-Developer-741, line 62: Could this be removed? Not an important issue but it prevents the text cursor from appearing on the current breadcrumb, which might be confusing to users

GCHQ-Developer-847:
> Have noticed that on the React Router story, selecting a different breadcrumb means that the previously selected one is not focusable, it doesn't seem to be adding the href property back in or resetting the tabIndex

Should be sorted now - added a watch on the `current` prop and the href gets added back on when it gets set to false

packages/web-components/src/components/ic-breadcrumb/ic-breadcrumb.css
GCHQ-Developer-847: I’ve checked and this is definitely needed in case the functionality to remove the href doesn’t work (e.g. if the slotted link is structured differently, or uses a `to` prop instead of `href` or something etc.)

The tabindex I added prevents keyboard focus, but `pointer-events: none` is needed too otherwise the user could still click on the link, and the “pointer” cursor would be displayed.

What I have done though is add `cursor: text` so it now displays the correct cursor

GCHQ-Developer-741 (approved):

GCHQ-Developer-530 (approved):

Issue #2963, opened 2025-01-02
[ic-alert] - button within alert is not as expected in dark mode


MI6-286:
## Summary of the bug 
When in dark mode, the button within an ic-alert is not displaying as expected, i.e., the button is black when it should be white.

## 🪜 How to reproduce 
1. View an alert containing a button in the action slot in dark mode, for example the 'With action' alert code example on the ICDS website
2. See that the button is black instead of white

Note: Issue also seen in Storybook and Cypress.

## 📸 Screenshots or code 
![image](https://github.com/user-attachments/assets/ee729c16-6d73-4d14-b587-3ce8fd7987c7)

## Additional info 
Remove the skip tag from the Cypress test within IcTheme.cy.tsx and generate the visual regression image. Consider updating the test data for the existing ic-alert dark mode visual regression test to contain buttons within the alerts.

PR #3077, opened 2025-01-23
2963 ic alert: button within alert is not as expected in dark mode

GCHQ-Developer-299:
## Summary of the changes
Make action slot buttons appear as expected in Dark Mode

Uncomment ic-theme cypress test (as requested in ticket) to hopefully create screenshot demonstrating how good alert is now

## Related issue
#2963 

### Testing

### System modes


GCHQ-Developer-530 (dismissed):

ad9242:
looking in storybook (for the alert playground), the button seems to be blue text now?

image


but as far as i can tell in the designs, it should be black:
image


and then in dark mode in it looks like this in storybook:
image


but is white in the designs:
image

GCHQ-Developer-299:
> looking in storybook (for the alert playground), the button seems to be blue text now?
> 
> image
> 
> but as far as i can tell in the designs, it should be black: image
> 
> and then in dark mode in it looks like this in storybook: image
> 
> but is white in the designs: image

In order for the button to be black, the button in the action slot has to be set to monochrome. So I don't *think* this is a failure - but I will update those stories so they match the designs :)

ad9242:
open question: do you think the alert should automatically set the monochrome prop when there is slotted ic-button so a "blue" button never appears?

GCHQ-Developer-299:
> open question: do you think the alert should automatically set the monochrome prop when there is slotted ic-button so a "blue" button never appears?

I promise i'm not answering out of laziness, but I think no - if it was a prop that always rendered a button, then yes i'd say we enforce a monochrome button; but since it's a slot and we're letting people slot whatever they want, we can't guarantee this bespoke monochrome-ifying code would work if they use a different kind of input/element

gd2910:
Looks good to me.

Just chipping in on previous discussion, we could check if the slotted element is an ic-button before enforcing the secondary variant and monochrome attributes? I believe that this should be another ticket if the work is to be done though.

gd2910 (approved):

ad9242 (approved):

PR #3094, opened 2025-01-27
Prevent IcDataTable action element clicks from selecting the table row

dn55533:
## Summary of the changes
Provide access to the the click event in the Data Table action element, thereby allowing the event propagation to be stopped in individual data table action element click handlers.

This is to prevent IcDataTable rows from becoming selected when the row contains an IcButton in an action element and the button is clicked.

IcButton prevent propagation

## Related issue
N/A

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes

### DIFF
--- a/packages/canary-react/src/component-tests/IcDataTable/IcDataTable.cy.tsx
+++ b/packages/canary-react/src/component-tests/IcDataTable/IcDataTable.cy.tsx
@@ -90,6 +90,7 @@ const PAGINATION_GO_TO_PAGE_TEXT_FIELD_SELECTOR =
 const PAGINATION_GO_TO_PAGE_BUTTON_SELECTOR = ".go-to-page-holder ic-button";
 const ITEMS_PER_PAGE_SELECTOR = ".items-per-page-input";
 const ACTION_ELEMENT = "action-element";
+const TABLE_ROW_SELECTED = "table-row-selected";
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/explicit-module-boundary-types
 export const BasicDataTable = (dataTableProps?: any): ReactElement => (
@@ -1091,7 +1092,7 @@ describe("IcDataTables", () => {
     cy.findShadowEl(DATA_TABLE_SELECTOR, "td")
       .eq(0)
       .find("span")
-      .should(HAVE_CLASS, "action-element")
+      .should(HAVE_CLASS, ACTION_ELEMENT)
       .find("ic-button")
       .should("be.visible");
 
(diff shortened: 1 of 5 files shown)


ad9242 (changes requested):
This seems to have introduced some problems. For example: go to React Storybook : Dialog > Sizes > Launch Medium Dialog. Clicking the show button has no effect:

image

you can see the intended behaviour on the develop branch:
https://mi6.github.io/ic-ui-kit/branches/develop/react/?path=/story/react-components-dialog--sizes

I think it may have also introduced other problems that are being caught by the cypress tests in the CI pipeline

I think it would be best to add this new functionality under a prop (defaulting to false)

dn55533:
I have removed the call to stop propagation from the IcButton component because this was causing various issues. Instead I have exposed the click event in the Data Table action element, thereby allowing the event propagation to be stopped in individual data table action element click handlers.

I have updated the stories and example code accordingly.

GCHQ-Developer-530 (approved):

ad9242 (changes requested):
It would be good to add a visual regression test for this. Whilst it's kind of on the developer to add the stopPropogation call into the actionElementOnClick, it would mean we don't risk breaking anything in future if there are changes in the core data-table

ad9242 (changes requested):
sorry, I know I said a VR test, but to save the issue with generating the image, it could just check for the class name:

```
const TABLE_ROW_SELECTED = "table-row-selected";

it("should not highlight the selected row when the action element click event is stopped from propagating", () => {
    mount(
      <IcDataTable
        columns={COLS}
        data={ACTION_DATA_ELEMENTS}
        caption="Data tables"
      ></IcDataTable>
    );

    cy.checkHydrated(DATA_TABLE_SELECTOR);

    cy.get(DATA_TABLE_SELECTOR)
      .shadow()
      .find(`[data-testid="copy-button"]`)
      .eq(0)
      .click();

    cy.findShadowEl(DATA_TABLE_SELECTOR, "tr").eq(1).should(NOT_HAVE_CLASS, TABLE_ROW_SELECTED);

    cy.get(DATA_TABLE_SELECTOR)
      .shadow()
      .find(`[data-testid="copy-button"]`)
      .eq(1)
      .click();
    
    cy.findShadowEl(DATA_TABLE_SELECTOR, "tr").eq(2).should(HAVE_CLASS, TABLE_ROW_SELECTED);
  });
```

Would you mind changing it to this, and remove the added image please? Otherwise happy to approve

ad9242 (approved):

GCHQ-Developer-847 (approved):

PR #3147, opened 2025-02-06
2710 ictoast auto dismiss toasts are not read out by screen reader

GCHQ-Developer-299:
## Summary of the changes
Changed auto-dismiss toasts to use role=alert rather than role=status, and they now read out to screenreader users

## Related issue
#2710 

### Testing

### Accessibility 


GCHQ-Developer-741:
This may not be behavior we want to enable. The alert aria role has [docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role)that recommend to use it sparingly, as the intrusiveness of the message may not always be proportionate. In the [original toast ticket](https://github.com/mi6/ic-ui-kit/issues/30), the decision was made to use the status role to counteract this behavior (albeit the a11y docs on our site appear to have not gotten the message, as the reporter of this current issue noticed).

Whilst I recognize that having it interrupt everything when its read out is preferable to it not being read out at all, I just wonder if we're following best practices when it comes to creating a toast component. Theoretically the [status role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/status_role) should announce it when the user becomes idle, there perhaps might be a deeper issue here.

These are just some points to prompt a discussion, if it turns out to be the best way I guess it can be merged and we will have to ensure people understand what will happen via guidance.

GCHQ-Developer-741:
Following on from that first comment, when testing with NVDA I get the subheading read out, but not the heading. In VoiceOver I get everything read out, but the icon aria-label and the loading indicator get read out twice each (EG: "error error ... loading loading")

GCHQ-Developer-299:
> Following on from that first comment, when testing with NVDA I get the subheading read out, but not the heading. In VoiceOver I get everything read out, but the icon aria-label and the loading indicator get read out twice each (EG: "error error ... loading loading"). I could've sworn there was a ticket addressing that second bit as its not related to this, but I can't find it anywhere.

I worked out the issue here - seems NVDA doesn't care about `<h5>` elements in live regions.  So i've added a conditional render that makes the header in Auto Dismiss Toasts into a `<p>` and now the it reads out to NVDA without disrupting the heading structure in Manual Toasts :)

GCHQ-Developer-741:
Looks good, although I don't think the aria-live is doing much, so I guess it can get taken out. To be honest if the status role isn't working at all, I think getting it to announce something using the alert role is the most important thing, as I mentioned earlier. Really not sure why the status role doesn't work though.

GCHQ-Developer-741:
Could the conditional h5 just be a <p> tag in all cases?

GCHQ-Developer-299:
> Looks good, although I don't think the aria-live is doing much, so I guess it can get taken out. To be honest if the status role isn't working at all, I think getting it to announce something using the alert role is the most important thing, as I mentioned earlier. Really not sure why the status role doesn't work though.

The alert with aria-live=polite does feel like a little bit of a botch - it's just doing what the status role should've been doing the whole time.  Status not working might be related to the rendering of the component - role=status might be for elements that are always rendered, but their contents change, so the screenreader isn't picking them up because the element doesn't exist at first.

GCHQ-Developer-299:
> Could the conditional h5 just be a p tag in all cases?

It could - the manual dismiss toasts benefit from some semantic structure with an H5 in there, maybe? And my intent with this change was to fix auto-dismiss toasts without changing the manual toasts

GCHQ-Developer-741 (approved):

GCHQ-Developer-847 (approved):

Issue #3098, opened 2025-01-27
Badges aren't visible on top nav navigation buttons on small screens


GCHQ-Developer-847:
## Summary of the bug 
On a top nav on a small screen, any badges that have been added to navigation buttons are not visible.

## 🪜 How to reproduce 
1. View a top nav component which has badges on its navigation buttons (e.g. [StackBlitz](https://stackblitz.com/edit/n8yncwpp?file=src%2Fapp.tsx&hideExplorer=1))
2. Make the screen smaller so that the navigation items and buttons collapse into a menu
3. With the menu expanded, see that the badge is not visible at all

## 📸 Screenshots or code 
Badge is visible on a larger screen e.g.:

<img src="https://github.com/user-attachments/assets/add1602c-74dc-41e3-9f0e-fc49e3067ed4e" width="500">

But then you can't see it anywhere where you make the screen smaller (see the "Notifications" button):

<img src="https://github.com/user-attachments/assets/64af7e9a-d052-4538-bb9e-94a4a4653dae" width="200">

## 🧐 Expected behaviour 
The badge should move to the inline position when on a small screen (e.g. like the badge in this [tabs Storybook example](https://mi6.github.io/ic-ui-kit/branches/v3.0.0/develop/web-components/?path=/story/web-components-tabs--playground)).

## 📝 Acceptance Criteria  
*Given* there is a top navigation on a page with a navigation button that has a badge on it
*When* I make the screen smaller and expand the navigation menu
*Then* the badge should be visible and displayed in the "inline" position

## Additional info 
Similar to #3079.

I think the `position` prop on ic-badge should be overridden to always be "inline" when on a small screen (because allowing the "near" and "far" positions would mean the badge is cut off at the side of the screen).

MI6-255:
Will need to be fixed in v2 and v3

PR #3163, opened 2025-02-11
fix(web-components): changes to ic-badge and ic-navigation-button to …

gd2910:
…show notifications in mobile

ic-badge now renders as inline variant when in ic-top-navigation and in mobile view. Colour changes as a workaround until design ticket in.

## TO TEST
Go to the [`ic-top-navigation` `With icons` story](https://mi6.github.io/ic-ui-kit/branches/3098-badges-arent-visible-on-top-nav-navigation-buttons-on-small-screens/react/?path=/story/react-components-top-navigation--with-icons), resize window to mobile view and click the menu button

## Summary of the changes
Tell us, in a few words, what the changes are.

## Related issue
#3098


packages/web-components/src/components/ic-badge/ic-badge.tsx
```diff
@@ -124,6 +135,24 @@ export class Badge {
       );
   }
 
+  componentWillRender(): void {
+    this.isInTopNav() &&
+      (this.customColor = "#d4351c") &&
```
GCHQ-Developer-299, line 140: feels a bit weird having this hex value hard-coded rather than brought in as a variable - is it possible to source this from a variables file?

Shouldn't all the = in these logical operations be == to check the validity of the value - i think instead they're setting the value

GCHQ-Developer-299 (changes requested):

packages/web-components/src/components/ic-badge/ic-badge.tsx
gd2910: They're setting the attributes rather than checking equality. Have made the hardcoded value a local variable.

GCHQ-Developer-299 (dismissed):

GCHQ-Developer-299 (approved):

GCHQ-Developer-530 (approved):

GCHQ-Developer-299 (approved):

PR #3207, opened 2025-02-18
docs(web-components): description change in the buttons prop

srt544:
description change to say default button will not be shown however, slotted controls will

### DIFF
--- a/packages/web-components/src/components/ic-dialog/ic-dialog.tsx
+++ b/packages/web-components/src/components/ic-dialog/ic-dialog.tsx
@@ -53,7 +53,7 @@ export class Dialog {
   @State() fadeIn: boolean = false;
 
   /**
-   * If set to `false`, default buttons or slotted dialog controls will not be displayed.
+   * If set to `false`, default buttons will not be shown, but slotted dialog controls will still be displayed.
    */
   @Prop() buttons?: boolean = true;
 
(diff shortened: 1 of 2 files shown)


GCHQ-Developer-741 (approved):

gd2910:
As part of this PR would you mind removing the `console.log` [here](https://github.com/mi6/ic-ui-kit/blob/6519f8d121522b4161249663e0145c055037210c/packages/web-components/src/components/ic-toggle-button/ic-toggle-button.tsx#L161)?

packages/web-components/src/components/ic-dialog/ic-dialog.tsx
```diff
@@ -53,7 +53,7 @@ export class Dialog {
   @State() fadeIn: boolean = false;
 
   /**
-   * If set to `false`, default buttons or slotted dialog controls will not be displayed.
+   * If set to `false`, default buttons will not be shown, slotted dialog controls will still be displayed.
```
GCHQ-Developer-847, line 56: Small suggestion - just to make it a tiny bit clearer
```suggestion
   * If set to `false`, default buttons will not be shown, but slotted dialog controls will still be displayed.
```

gd2910 (approved):

GCHQ-Developer-299 (approved):

Issue #2197, opened 2024-07-19
Open in new window / tab icon on link doesn't turn purple when visited in Safari


GCHQ-Developer-847:
## Summary of the bug 
The 'open in new window / tab' icon which can be displayed next to a link does not purple when the link is visible in Safari.

## 🪜 How to reproduce 
1. On Safari, find an external link e.g. [on the website](https://design.sis.gov.uk/accessibility) (at the bottom of the page).
2. Click on the link (to make it become a visited link).
3. Go back to the link and see that the text has turned purple but the icon is still blue.

## 📸 Screenshots or code 
<img src="https://github.com/user-attachments/assets/dfe15cfa-cbf8-4859-bfd5-30b8b7017ce6" width="300">

## 🖥 📱 Device 
- Browser: Safari

## 🧐 Expected behaviour 
The icon should turn purple, matching the text.

evenstensberg:
MDX/Gatsby issue so fixed it there: https://github.com/mi6/ic-design-system/pull/1382

MI6-255:
Moved out of done as was causing an issue with links on Hero appearing as the purple visited colour. Changes have been reverted.

GCHQ-Developer-530:
This is a [known issue with Safari 15](https://bugs.webkit.org/show_bug.cgi?id=244025), it has been resolved so should be resolved for us without any changes in a future Safari update.

PR #3266, opened 2025-02-25
fix(web-components): fixed open in new icon not applying correct color

GCHQ-Developer-741:
## Summary of the changes
Updated css selectors to apply fill to svg directly rather than relying on currentcolor to avoid safari issues

## Related issue
#2197

### DIFF
--- a/packages/web-components/src/components/ic-link/ic-link.css
+++ b/packages/web-components/src/components/ic-link/ic-link.css
@@ -60,30 +60,22 @@
   margin-left: var(--ic-space-xxs);
 }
 
-.ic-link > .ic-link-open-in-new-icon {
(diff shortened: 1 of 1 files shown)


packages/web-components/src/components/ic-link/ic-link.css
GCHQ-Developer-847, line 1: On the dark and light variants, the "open in new" icon displays as the purple visited colour when the link is visited, but it should stay as white / black like how it was appearing previously

GCHQ-Developer-741: Should be working now

GCHQ-Developer-847 (approved):

GCHQ-Developer-530 (approved):

PR #3270, opened 2025-02-26
fix(web-components): fix to prevent numeric text field being modified by scroll

MI6-286:
## Summary of the changes
Added fix to blur text field to prevent numeric text field input being modified by scroll.

## Related issue
#2549


gd2910 (changes requested):
I think the component should mimic the native input behaviour, which scrolls the whole page rather than `.blur`, I can imagine a user wanting to keep focus on the input while scrolling to see other information before entering a value.

gd2910 (approved):

GCHQ-Developer-847 (approved):

PR #3279, opened 2025-03-03
1791 breadcrumb back arrow react router

srt544:
Back breadcrumb arrow will now be rendered when using react router

## Related Issue

#1791


packages/react/src/component-tests/IcBreadcrumb/IcBreadcrumb.cy.tsx
```diff
@@ -22,6 +23,9 @@ import {
   NOT_EXIST,
 } from "../utils/constants";
 import { setThresholdBasedOnEnv } from "../../../cypress/utils/helpers";
+import { IcBreadcrumb, IcBreadcrumbGroup, IcLink } from "../../components";
```
ad9242, line 26: I think you can remove all these added imports, as they are not used. The ones that are needed are already in the testdata file

ad9242 (changes requested):
please can you also add the baseline cypress image, which is generated by running the update cypress visual regression test images action?

GCHQ-Developer-299 (approved):

GCHQ-Developer-530 (approved):

ad9242 (approved):

PR #3267, opened 2025-02-25
2154 allow for changing the position of the back to top component

gd2910:
## Summary of the changes
Position prop added to `ic-back-to-top` to position the element on the left, central or (default) to the right.

## Related issue
#2154

PR #3286, opened 2025-03-04
3284 ic back to top expose css variables to allow changing of x and y position

gd2910:
## Summary of the changes
Exposes three CSS variables to allow a developer to more accurately customise the position of the ic-back-to-top component

## Related issue
#3267


packages/react/src/component-tests/IcBackToTop/IcBackToTopTestData.tsx
```diff
@@ -1,15 +1,29 @@
...
     <div style={{ height: "120vh" }}>
       <div id="topEl" style={{ position: "absolute", top: "0", width: "100%" }}>
         Top of screen
       </div>
-      <IcBackToTop target="topEl" variant={variant} theme={theme} />
+      <IcBackToTop
+        target="topEl"
+        variant={variant}
+        theme={theme}
+        position={position == undefined ? "right" : position}
```
evenstensberg, line 25: To avoid a future regressions like in the ic-banner being undefined, we should check all falsy values of position.

packages/web-components/src/components/ic-back-to-top/test/basic/ic-back-to-top.spec.tsx
```diff
@@ -180,4 +180,22 @@ describe("ic-back-to-top", () => {
...
     expect(page.root).toMatchSnapshot("should render with an icon only");
   });
+
+  it("should render positioned on the left", async () => {
+    const page = await newSpecPage({
+      components: [BackToTop, Button],
+      html: `<div id="topEl"><ic-back-to-top target="topEl" position="left"></ic-back-to-top></div>`,
+    });
+
+    expect(page.root).toMatchSnapshot("should render positioned on the left");
```
evenstensberg, line 190: I thought `toMatchSnapshot` doesn't take any parameters?

packages/react/src/component-tests/IcBackToTop/IcBackToTopTestData.tsx
gd2910: The above code is to avoid some type errors int he test data, `position` defaults to the right in regular use.

packages/web-components/src/components/ic-back-to-top/ic-back-to-top.css
```diff
@@ -28,7 +31,7 @@
 }
 
 :host(.ic-back-to-top-left) ic-button {
-  margin-left: var(--ic-space-md);
+  margin-left: var(--margin-left, 1rem);
```
GCHQ-Developer-741, line 34:
```suggestion
  margin-left: var(--margin-left, var(--ic-space-md));
```

Can't the default be set to the css var, like below

gd2910: Sorry should have mentioned in the PR notes, I attempted this way but the Cypress tests wouldn't recognise the fallback value.

GCHQ-Developer-741 (approved):

GCHQ-Developer-530 (approved):

PR #3299, opened 2025-03-06
feat(web-components): styling for RTL characters in ic-radio-option

gd2910:
Styling for RTL characters in ic-radio-option


GCHQ-Developer-299:
Could you add another commit to bump the audit dates in [audit-ci.json](https://github.com/mi6/ic-ui-kit/blob/v3.0.0/develop/audit-ci.json)

Issue #2577, opened 2024-10-14
All tabPanels displayed when tabs are dynamically loaded and tab is selected


GCHQ-Developer-847:
## Summary of the bug 
All tabPanels are displayed when tabs are dynamically loaded and a tab is selected.

This has been looked into before (PR #2510) but without success. Thought to get a ticket created for this so the other tabs bug fixes aren't blocked, and it might also be good to get a fresh pair of eyes on this.

This is a task remaining from #1799.

## 🪜 How to reproduce 
Tell us the steps to reproduce the problem: 
1. Go to the ["Dynamic rendering" tabs story](https://mi6.github.io/ic-ui-kit/branches/develop/react/?path=/story/react-components-tabs--dynamic-rendering)
2. Wait until the tabs have loaded
3. See that all tab panels are displayed
4. Also happens when you create new tabs (using the "Add tab" button) and select the first newly created tab.

## 🧐 Expected behaviour 
Only the tab panel for the selected tab should be displayed.

## Additional info 
Please see PR #2510 for more details about what has already been attempted to fix this bug.

GCHQ-Developer-299:
I believe this needs a full refactor of how tabs work - rather than slotting in Tabs and TabPanels as slots inside the TabContext without a heirarchy, I think TabContext should have defined Tabs and TabPanel elements, that then allow data to be slotted inside. That additional structure will hopefully be able to force only one TabPanel to be displayed at once, and maybe sort out another issue raised recently by a user.

PR #3304, opened 2025-03-06
Fix/2577 all tab panels displayed when dynamically loaded

GCHQ-Developer-847:
## Summary of the changes
Fixed issue where all tab panels get displayed when they are created after a delay / dynamically (see "Dynamic rendering" React story).

Removed code which manages the hidden / visible state within `ic-tab-panel.tsx` - it just gets handled by the parent `ic-tab-context`.

-------

It was hard to get head my around what was going on with this bug but I think it was because the tab panels are not hidden by default when first rendered, and `configureTabs()` only gets called when the number of tabs equals the number of tab panels. So I've switched the order around; the tab panels get hidden when they load (by setting the new prop, `active`, to false by default), and then the selected tab gets displayed when all tabs and tab panels are loaded.

I tried doing this by setting the `hidden` attribute to true by default but it was just not getting applied in the DOM, hence why I have used a different approach (a new prop) to hide the tab panels.

## Related issue
#2577 

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes


packages/react/src/component-tests/IcTabs/IcTabsTestData.tsx
```diff
@@ -400,3 +400,35 @@ export const MinMaxTabs = (): ReactElement => {
...
+
+  return (
+    <IcTabContext>
+      <IcTabGroup label="Example tab group">
+        {data.map((item, i) => (
+          <IcTab key={`${item.label}-${i}`}>{item.label}</IcTab>
+        ))}
+      </IcTabGroup>
+      <span></span>
+      {data.map((item, i) => (
```
evenstensberg, line 425: `idx` for more readability. Sorry for nitpicking!

packages/react/src/component-tests/IcTabs/IcTabsTestData.tsx
```diff
@@ -400,3 +400,35 @@ export const MinMaxTabs = (): ReactElement => {
...
   );
 };
+
+export const DelayedTabs = (): ReactElement => {
+  const [data, setData] = useState([]);
+
+  useEffect(() => {
+    setTimeout(() => {
+      setData([
+        { label: "Ingredients" },
```
evenstensberg, line 410: For performance reasons declaring constants outside the function is good (in relation to caching), altough I see it breaks encapsulation. 

A good "nice-to-know" if you're in this situation other places outside the test suite.

ad9242 (approved):

GCHQ-Developer-741 (approved):

Issue #266, opened 2023-01-10
Turn on typescript strict mode


GCHQ-Developer-530:
## Summary
Enable typescript strict mode and amend code according to any warnings/errors that present themselves.

## 💰 Use value
It would assist developers in their way of writing code so that it's more consistent no matter who wrote it.

GCHQ-Developer-847:
Canary web components updates have been moved out into a separate ticket (#3330) to make the work more manageable

PR #3328, opened 2025-03-15
feat(web-components): make changes to first half of web-components fi…

GCHQ-Developer-847:
## Summary of the changes
Updated the first half of the web-components files to resolve strict mode errors. Also added some unit tests as the changes brought the coverage below 80%.

I have added the non-null assertion operator (!) after some variables. This is because those variables are component props which have a default value; the TypeScript strict mode checking doesn't seem to detect that these default values exist.

<br />

Wanted to make these updates in multiple PRs to make reviewing the changes more manageable (and prevent GitHub becoming very slow when viewing all the changes).

I will also create a separate PR to actually enable TypeScript mode (the build would fail if I added it in this PR). Canary web components updates are to be done in a separate ticket: #3330.

## Related issue
#266

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes


packages/web-components/src/components/ic-back-to-top/ic-back-to-top.tsx
```diff
@@ -110,8 +110,8 @@ export class BackToTop {
     this.setFooterVisible(entries[0].isIntersecting);
   };
 
-  private findTargetEl = (target: string): Element => {
-    let targetElement: Element = null;
+  private findTargetEl = (target: string): Element | null => {
```
evenstensberg, line 113: Later, create a simple type for `Element | null` in a util file and import it where it is used?

GCHQ-Developer-847: Thank you for your suggestion - I might leave it for this PR if that is ok, but something we will definitely consider in future, especially if we start using the `Element | null` type a bit more often

gd2910 (approved):

PR #3335, opened 2025-03-19
Readonly ic select fix

srt544:
updated baseline images and resized the input container's height in the read only to 'auto' from '2.5rem'


ad9242 (changes requested):
please add updated baseline image

ad9242 (approved):

GCHQ-Developer-847 (approved):

PR #3345, opened 2025-03-20
3010 date input extend events

cd3859:
## Summary of the changes
Added a dateObject field to IcChange return detail, IcChange is now emitted after every section of input date is completed 

## Related issue
#3010 

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes

### DIFF
--- a/packages/canary-react/src/component-tests/IcDateInput/IcDateInput.cy.tsx
+++ b/packages/canary-react/src/component-tests/IcDateInput/IcDateInput.cy.tsx
@@ -811,12 +811,15 @@ describe("IcDateInput end-to-end, visual regression and a11y tests", () => {
       .shadow()
       .find("button")
       .focus()
-      .click();
-
-    cy.findShadowEl(DATE_INPUT, MONTH_INPUT_ARIA_LABEL).should(HAVE_VALUE, "");
+      .click("top");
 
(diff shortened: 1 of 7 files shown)


GCHQ-Developer-847 (changes requested):
Looks good overall but have just found a couple of issues (while testing in the “icChange event” story):  

- If you only type in part of the date, e.g. the day, and then delete it by either pressing backspace or clicking the clear button, the icChange event does not get emitted to say that that section is now empty again

- The sections sometimes get set to `null` and other times an empty string. E.g. if you only type in the day, the month and year sections show as an empty string in the icChange event. But if you type in only the year and then click the clear button, the icChange event gets emitted saying that the day and month are `null`. I think it should be consistent i.e. either they only ever get emitted as `null` when cleared / deleted, or only as an empty string (maybe `null` would be best to match the `value` behaviour?)

ad9242:
just wondering if changing this to now emit for each field is going to cause developers issues? previously they would get 1 icChange event once the date was complete\the field blurred, but now they will be getting multiple as the user types & will need to adjust their code to account for this

eg 
user types 01 in day field:
icChange triggered with value null and day value 01

user than types 02 in month field:
icChange triggered with value null, day value 01 and month value 02

Issue #3386, opened 2025-04-02
Data Table: Make Rows selectable with checkboxes


GCHQ-Developer-299:
## Summary 

New property added to data-table that adds new column to table that is all checkboxes.  Allowing you to select one row by clicking it's relevant checkbox, and select all rows by clicking the checkbox in the header.  (see AG Grid example on the design site for an example)

<img width="336" alt="Image" src="https://github.com/user-attachments/assets/1f6c153c-ecb6-4d6b-a5c8-be04c99db2de" />

Emit selectAll event when checkbox is clicked 

Note: do we need enhancements to IC Checkbox to use it in this way (without a label)
 

## 💰 Use value 
Fulfilling PI 11 Goal

## ✏ Designs 
??

## 🧾 Guidance 
??

mi6-577:
Need to consider;

- click on checkbox for row to select it ...clicking on the row would not work as some cells may be links
- clicking on checkbox in column to select all rows ...this should be all visible rows for the current page, not all rows in the entire table across all pages
- potential use of keyboard to select multiple rows at a time (adjacent and non-adjacent) eg by using shift

PR #3452, opened 2025-04-22
3386 data table make rows selectable with checkboxes

GCHQ-Developer-741:
## Summary of the changes
- Removed `highlightSelectedRow`, replacing with `rowSelection` which defaults to false. This prop no longer controls clicking a row to select it, as that functionality has been replaced with checking the checkbox. This prop will display the checkboxes when set to `true`. In other data tables it is possible to select individual cells, with row selection done via checkbox elements, so this brings us in line with that behaviour.
- Updated row selection event to emit both the selected row and an array of all the currently selected rows, since it is now possible to select multiple rows at once.
- Added new event for when the header checkbox is selected, which outputs all the rows or an empty array depending on if it is checked or not.
- Added visual regression tests and updated old ones based on new behaviour.

## Related issue
#3386 

### General 

### Testing

- [x] Compare performance of modified components against develop using Performance addon in React Storybook

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes


GCHQ-Developer-299 (dismissed):

packages/canary-react/src/stories/ic-data-table.stories.js
GCHQ-Developer-847, line 1: I've just tried enabling `highlightSelectedRow` in the playground story and clicking the header checkbox seems to select all rows in the table rather than just those on the current page

packages/canary-web-components/src/components/ic-data-table/ic-data-table.tsx
```diff
@@ -1691,17 +1703,40 @@ export class DataTable {
...
+                  [`table-density-${this.density}`]: this.notDefaultDensity(),
+                }}
+              >
+                <div class="checkbox-wrapper">
+                  <ic-checkbox
+                    class="ic-data-table-checkbox"
+                    checked={isRowSelected}
+                    disabled={this.updating || this.loading}
+                    hideLabel
+                    label={cellIndex}
```
GCHQ-Developer-847, line 1731: I feel like this label could be changed to be a bit more helpful for screen reader users. Maybe it could be "Select row \<row number\>"? And then it changes to "Deselect..." when selected

(Don't know if including the row number is necessary but might be useful)

packages/canary-web-components/src/components/ic-data-table/ic-data-table.tsx
```diff
@@ -195,9 +196,9 @@ export class DataTable {
   @Prop() height?: string;
 
   /**
-   * If `true`, the selected row is highlighted using a background colour.
+   * If `true`, the selected row is highlighted using a background colour. This will display a checkbox column to the left of the table.
    */
-  @Prop() highlightSelectedRow: boolean = true;
+  @Prop() highlightSelectedRow: boolean = false;
```
GCHQ-Developer-847, line 201: I'm just wondering if `highlightSelectedRow` is still the best name for the prop since it is now doing a lot more than just highlighting the row / changing the background colour. And also therefore whether the prop description should be a bit different (e.g. essentially removing the bit about the background colour)?

This is what I was thinking but just my opinion - happy to discuss :)

```
/**
 * If `true`, a checkbox column will be displayed to the left of the table which allows multiple rows to be selected.
 */
@Prop() rowSelection: boolean = false;
```

GCHQ-Developer-847 (changes requested):
Spotted a few things, but looks really good overall!

GCHQ-Developer-847 (approved):

Issue #3422, opened 2025-04-11
[ic-stepper] Disabled stepper text color fails contrast ratio checks


ad9242:
## Summary of the bug 
After changes made to bring tokens up to date under #3363, there is still an issue with the disabled step color. This now uses architectural-300 under the covers but this does still not meet required contrast ratios.

This can be recreated by running the ic-stepper visual regression tests. Look for instances of 
// cy.checkA11yWithWait(); A11y failure for disabled text

and uncomment them & run the tests. You can then see that the ic-typography element for the step title has the failure

## 📸 Screenshots or code 

![Image](https://github.com/user-attachments/assets/23cdfa1a-02ca-456a-97c1-a205084ee07c)

MI6-255:
Look into if we need guidance to explain that disabled states do not need to meet this requirement

GCHQ-Developer-299:
After looking into it, I think the solution is to add `aria-hidden` (or `disabled`) to the elements in Disabled step's 'heading-area' 

Looking at an Accessibility FastPass we can see that the Step number doesn't fail accessibility checks, because it already has `aria-hidden` on it. Screenshot below:
<img width="276" alt="Image" src="https://github.com/user-attachments/assets/1cc20946-ec94-4b9f-a341-61aa4a3d527b" />

1. Add `aria-hidden` or, if appropriate, `disabled` to the items in the header section of Disabled steppers, so they are exempt from accessibility checks. 
2. Re-enable the checkA11y line in the tests, it should now pass.

PR #3475, opened 2025-04-25
3422 stepper disabled stepper text colour fails contrast ratio checks

GCHQ-Developer-530:
## Summary of the changes
Update stepper and text field to hide disabled text before a11y check and add it back for snapshot so that a11y check passes.

## Related issue
#3422

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes


evenstensberg (commented):
lgtm :shipit:

GCHQ-Developer-299 (dismissed):

ad9242 (commented):
sorry, a bit late to this, but would an alternative be to turn off the color-contrast rule instead: eg: 

rules: {
      'color-contrast': { enabled: false },
    }

we already set custom rules in toggle button: https://github.com/mi6/ic-ui-kit/blob/f76eed8189cfaf598beb4d878df61bf18436e559/packages/react/src/component-tests/IcToggleButton/IcToggleButton.cy.tsx#L32

list of rules: https://github.com/dequelabs/axe-core/blob/develop/doc/rule-descriptions.md

GCHQ-Developer-530:
> sorry, a bit late to this, but would an alternative be to turn off the color-contrast rule instead: eg:
> 
> rules: { 'color-contrast': { enabled: false }, }
> 
> we already set custom rules in toggle button:
> 
> https://github.com/mi6/ic-ui-kit/blob/f76eed8189cfaf598beb4d878df61bf18436e559/packages/react/src/component-tests/IcToggleButton/IcToggleButton.cy.tsx#L32
> 
> list of rules: https://github.com/dequelabs/axe-core/blob/develop/doc/rule-descriptions.md

Yeah I thought of that, but it would have to be turned off for the whole test, so I figured if the other step colours changed then we'd want to check the colour contrast of them

ad9242:
> > sorry, a bit late to this, but would an alternative be to turn off the color-contrast rule instead: eg:
> > rules: { 'color-contrast': { enabled: false }, }
> > we already set custom rules in toggle button:
> > https://github.com/mi6/ic-ui-kit/blob/f76eed8189cfaf598beb4d878df61bf18436e559/packages/react/src/component-tests/IcToggleButton/IcToggleButton.cy.tsx#L32
> > 
> > list of rules: https://github.com/dequelabs/axe-core/blob/develop/doc/rule-descriptions.md
> 
> Yeah I thought of that, but it would have to be turned off for the whole test, so I figured if the other step colours changed then we'd want to check the colour contrast of them

i think it would be ok though, as you can pass custom options to the call eg:

```
cy.checkA11yWithWait(undefined, undefined, TOGGLE_BUTTON_AXE_OPTIONS);
```

but if you don't pass anything, it should be using the default rules in the `optionObject`, from looking at this:

```
const checkA11yWithWait = (
  element: string | undefined = undefined,
  wait = 200,
  optionObject = CYPRESS_AXE_OPTIONS
): void => {
  cy.wait(wait);
  cy.checkA11y(element, optionObject, CYPRESS_AXE_REPORTING);
};
```

MI6-286 (approved):

packages/react/src/component-tests/IcStepper/IcStepperTestData.tsx
```diff
@@ -137,7 +130,6 @@ export const SmallConnectorWidth = () => {
           subheading="Optional subtitle that is long and should wrap"
           type="current"
         />
-        <IcStep heading="Third" type="disabled" />
         <IcStep
           heading="Fourth title that is long and should wrap"
```
GCHQ-Developer-299, line 134: I'm being pedantic here, but these headings still reference being "fourth" even though they're now third

GCHQ-Developer-530: Good spot, I'll update them now 👍

MI6-286 (approved):

GCHQ-Developer-299 (approved):

PR #3480, opened 2025-04-28
V3 Fix appearance and non-appearance of length validation message in ic-textfield

dn55533:
## Summary of the changes
There were three issues causing the max length validation message to not appear correctly.

1. The max length validation message was not appearing when pasting in more than maxLength characters. The validation message only appeared when another key was pressed (e.g. ArrowLeft) after the paste and data truncation had occurred.

2. The max length validation message was also appearing when it shouldn't have been, when maxLength chars had been entered into the textfield and then keyboard keys were pressed with don't affect the length of the input, e.g. ArrowLeft, Home, End, etc. If one of these keys has been pressed then then the keypress is ignored when working out whether to display the maxLength validation message.

3. Similar to 2. above, another case of the validation message appearing when it shouldn't have been, is when maxLength chars had been entered into the textfield and then the ctrl key along with another character, e.g. Ctrl-a (to select all the contents of the field). This is now prevented - if the ctrl key is pressed then the keypress is ignored when working out whether to display the maxLength validation message.

## Related issue
N/A

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes

### DIFF
--- a/packages/web-components/src/components/ic-text-field/ic-text-field.tsx
+++ b/packages/web-components/src/components/ic-text-field/ic-text-field.tsx
@@ -35,7 +35,10 @@ import {
   checkSlotInChildMutations,
   removeHiddenInput,
 } from "../../utils/helpers";
-import { IC_INHERITED_ARIA } from "../../utils/constants";
+import {
+  IC_INHERITED_ARIA,
+  IGNORED_KEYBOARD_CHARACTERS,
+} from "../../utils/constants";
 import {
   IcAriaAutocompleteTypes,
   IcTextFieldInputModes,
@@ -311,6 +314,9 @@ export class TextField {
 
     if (this.maxCharacters! > 0) {
       value = newValue.substring(0, this.maxCharacters);
+      if (value!.length < newValue!.length) {
+        this.maxCharactersWarning = true;
+      }
       this.value = value;
     } else {
       value = newValue;
@@ -422,7 +428,10 @@ export class TextField {
   @Listen("keydown", {})
   handleKeyDown(ev: KeyboardEvent): void {
     this.icKeydown.emit({ event: ev });
-    this.maxCharactersWarning = this.maxCharactersReached;
+
+    if (!ev.ctrlKey && !IGNORED_KEYBOARD_CHARACTERS.includes(ev.key)) {
+      this.maxCharactersWarning = this.maxCharactersReached;
+    }
   }
 
   /**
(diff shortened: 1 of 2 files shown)


packages/web-components/src/components/ic-text-field/ic-text-field.tsx
```diff
@@ -45,6 +45,35 @@ import {
 let inputIds = 0;
 const MUTABLE_ATTRIBUTES = [...IC_INHERITED_ARIA, "title"];
 
+const ignoredKeyboardCharacters = [
```
evenstensberg, line 48: immutable global const should be all uppercase

> Reason
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const#basic_const_usage

> Optional reading

- https://stackoverflow.com/questions/838929/naming-why-should-named-constants-be-all-uppercase-in-c-java
- https://akshayvjain.medium.com/uppercase-with-is-pretty-much-the-standard-for-hard-coded-constants-in-javascript-67a9bd374793

GCHQ-Developer-530, line 48: Could you please move this into the [constants file](https://github.com/mi6/ic-ui-kit/blob/v3.0.0/develop/packages/web-components/src/utils/constants.ts)? It might be something we want to use in other components down the line

packages/web-components/src/utils/constants.ts
```diff
@@ -91,3 +91,33 @@ export const IC_BLOCK_COLOR_EXCEPTIONS: IcColorExceptions = {
...
+  "F8",
+  "F9",
+  "F10",
+  "F11",
+  "F12",
+  "Home",
+  "PageDown",
+  "PageUp",
+  "Shift",
+  "Tab",
```
ad9242, line 122: Could we add a few more please?

"CapsLock"
"NumLock"
"Delete"
"Insert"
"Meta" (eg windows key)
"ContextMenu"
"AltGraph"
"AudioVolumeMute"
"AudioVolumeDown"
"AudioVolumeUp"

ad9242 (changes requested):
just a few other keys spotted whilst testing, but otherwise happy to approve

GCHQ-Developer-299 (approved):
tested the change locally, all points in the PR description appear to be addressed, tidy work :)

ad9242 (approved):

PR #3483, opened 2025-04-29
V3 final fixes 741

GCHQ-Developer-741:
## Summary of the changes
When trying to install v3 components in a test app following instructions, got error about the peer dependency for the fonts package, which was set to an old version. This PR updates the peer dependency to the latest version so it will work when we release.


MI6-255 (changes requested):
Don't merge until after release

PR #3417, opened 2025-04-09
feat: add build metadata

evenstensberg:
## Summary of the changes

Adds build metadata for customers.

## Related issue

#2428

### General 

- [x] Changes to docs package checked and committed.
- [x] All acceptance criteria reviewed and met. 

### Testing

- [x] Relevant unit tests and visual regression tests added. 
- [x] Visual testing against Figma component specification completed. 
- [x] Playground stories in React Storybook up to date, with any prop changes and additions addressed. 
- [x] Compare performance of modified components against develop using Performance addon in React Storybook.

### Accessibility 

- [x] Accessibility Insights FastPass performed.
- [x] A11y unit test added and yields no issues.
- [x] A11y plug-in on Storybook yields no issues. 
- [x] Manual screen reader testing performed using NVDA and VoiceOver. 
- [x] Manual keyboard testing for keyboard controls and logical focus order. 
- [x] Correct roles used and ARIA attributes used correctly where required. 
- [x] Logical heading structure is maintained, and the HTML elements used for headings can be changed to fit within the wider page structure. 

### Resize/zoom behaviour 

- [x] Page can be zoomed to 400% with no loss of content. 
- [x] Screen magnifier used with no issues. 
- [x] Text resized to 200% with no loss of content.
- [x] Text spacing increased as per the [WCAG 1.4.12 success criterion](https://www.w3.org/TR/WCAG21/#text-spacing) with no loss of content.

### System modes

- [x] Browser setting 'prefers reduced motion' tested. No animations or motion visible whilst this setting is on. 
- [x] Windows High Contrast mode tested with no loss of content.
- [x] System light and dark mode tested with no loss of content.
- [x] Browser support tested (Chrome, Safari, Firefox and Edge). 

### Testing content extremes

- [x] Min/max content examples tested with no loss of content or overflow. 
- [x] All prop combinations work without issue. 
- [x] Tested for FOUC (Flash of Unstyled Content) in both SSR (Server-Side Rendering) and SSG (Static Site Generation) settings.
- [x] Controlled and uncontrolled input components tested.
- [x] Props/slots can be updated after initial render.

PR #3487, opened 2025-04-29
Revert "feat: add build metadata"

MI6-255:
Reverts mi6/ic-ui-kit#3417


ad9242 (approved):

evenstensberg:
It didnt work?

evenstensberg:
Please elaborate if you can.

MI6-255:
It is causing the following error in the built components, please feel free to re-contribute :) 
image

evenstensberg:
Ok, thanks!

It will be hard for me to contribute without having the reproduction repository, it would be better that the core team fixes this. As for triaging, I'd have a look at which environment the error throws in, because the context (global window) object might be different, such as an app using react and an app using ElectronJS. If you use different environment with different global variables (`globalThis`, `self`, `global`, `this`, `window`) then the fix is very easy: check which variable to inject to. ( `(if globalThis !== undefined) cont...`

If you can't make it work using the approach in the reverted PR with injected env variables related to the correct global variable, then hard coding the versions in the bootstrap function also will work.

Let me know if I can be of any help! :)

evenstensberg:
Initially I thought this was a platform issue when I saw "# window", but it's about the variables being assigned when their properties are frozen/sealed. I've opened a new PR.

PR #3510, opened 2025-05-01
3358 ic tree view trying to arrow down doesn't move to the next item when a child in tree view is expanded

srt544:
fix: arrow down moves to the next item when a child in tree view is expanded

### DIFF
--- a/packages/canary-react/src/component-tests/IcTreeView/IcTreeView.cy.tsx
+++ b/packages/canary-react/src/component-tests/IcTreeView/IcTreeView.cy.tsx
@@ -42,6 +42,32 @@ export const BasicTreeView = (
   </div>
 );
 
+export const BasicTreeViewWithCoffeeTypes = (
+  props?: object,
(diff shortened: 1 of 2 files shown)


packages/canary-react/src/component-tests/IcTreeView/IcTreeView.cy.tsx
```diff
@@ -825,6 +851,33 @@ describe("IcTreeView", () => {
     cy.get(TREE_ITEM).eq(4).should(HAVE_FOCUS);
   });
 
+  it.only("should navigate through tree items when expanded content is hidden using arrow keys", () => {
```
MI6-286, line 854: Just need to remove the only from this test.

ad9242 (changes requested):
just the issue mentioned by @MI6-286, otherwise all good

MI6-286 (approved):

ad9242 (approved):

Issue #3485, opened 2025-04-29
New guidance site homepage with ic-alert has content shift when switching between light and dark mode


GCHQ-Developer-741:
## Summary of the bug 
The new V3 guidance site page that has an ic-alert at the top now shifts slightly up/down when switching the theme. This appears to be caused by the border that appears in dark mode adding an extra 4px to the height of the component, but only seems to affect the `titleAbove` variant.

## 🪜 How to reproduce 
Tell us the steps to reproduce the problem: 
1. Go to page: https://design.sis.gov.uk/
2. Click on the theme toggle button
4. See the cards shift up and down with the new height

## 🧐 Expected behaviour 
The content shift should not be happening.

PR #3513, opened 2025-05-06
3485 new guidance site homepage with ic alert has content shift when switching between light and dark mode

GCHQ-Developer-741:
## Summary of the changes
- Added function to ic-alert to calculate and apply a dynamic min-height css variable based on the alert content and theme. By basing it on the content, larger alerts will now no longer shift page content below it when changing colours.
- Will update either by its theme prop, the global system settings, or an ic-theme if present. Added a `themeChange` internal event to ic-theme that is listened for in alert.
- Added cypress tests for the new css variable when the theme and content updates.
- Moved the theme type to the global types to avoid duplication when referencing in alert. Removed the previous css variable values from the variables.css file.

**QUESTION FOR REVIEWERS:** _Should the `themeChange` event have an internal tag, or could it be exposed for users? May potentially enable some useful functionality in the future._

## Related issue
#3485 

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes


GCHQ-Developer-299 (changes requested):
I think exposing the themeChange event for developers sounds like a good idea :)

Tested in React Storybook, on the playground story. While the 'titleAbove' alert now is the same height in Light and Dark mode, the regular alert now gets taller in dark mode due to the border (see screenshots below)
Screenshot 2025-05-07 at 08 37 11
Screenshot 2025-05-07 at 08 37 20

GCHQ-Developer-299 (dismissed):
Tried out single/multi lines alerts with tileAbove set to true/false and all of them are now the same height between themes :)

ad9242 (changes requested):
Happy for the themeChange to be a visible event. However, please can it be prefixed with "ic" to match our other event names - so `icThemeChange` ?

(tend to not prefix with "ic" if they are private\internal events)

GCHQ-Developer-299 (approved):

ad9242 (changes requested):

GCHQ-Developer-299 (approved):

ad9242 (approved):

Issue #2722, opened 2024-11-06
Popover Menu: Toggle variant menu items should not exist


GCHQ-Developer-299:
## Summary 
While working on #1897 I found that we get a failing on our Popover story examples with 'toggle' variant menu items. 

<img width="391" alt="Screenshot 2024-11-06 at 15 35 03" src="https://github.com/user-attachments/assets/6358cb20-defd-472c-9aa6-8c3831116793">

Accessibility Insights reports this button's existence as a failure, because interactive items should not be nested. 
<img width="1324" alt="Screenshot 2024-11-06 at 15 30 28" src="https://github.com/user-attachments/assets/ee3ae1fc-f881-4ebc-b8f9-4d566bf179a8">

Things that will need changing:
- [ ] Popover storybooks
- [ ] Popover cypress tests (might have been tackled by work on #1897)
- [ ] Popover guidance on the design-system site, multiple screenshots show the toggle variant - will need separate ticket on ic-design-system

## 💰 Use value 
To meet WCAG 4.1.2

## Design
This will require design to update the figma component for popover so it can't use toggle menu buttons. 

## 🧾 Guidance 
New screenshots will be required. 

## Additional info 
More info by Accessibility Insights: 
[More information about nested-interactive](https://dequeuniversity.com/rules/axe/4.9/nested-interactive?application=msftAI)
[WCAG 4.1.2](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html)

GCHQ-Developer-299:
This might require removing the ic-button from inside menu item, making sure that functionality/styling is maintained. 
Do before

gd2910:
On hold while a11y review on the component is completed.
A11y failure - nested interactive (screen readers may not be able access nested content)

Need to check if this is a real error as our components are a11y crafted and are tested for this information, may be able to change the `ic-menu-item` to pass the Cypress a11y checks when this work may not be needed and may even open up a can of a11y worms if we remove the `li` or lose the `ic-button` features.

PR #3551, opened 2025-05-16
2722 popover menu toggle variant menu items

GCHQ-Developer-299:
## Summary of the changes

- Refactor ic-menu-item to remove redundant `<li>` element
- Refactor ic-popover-menu to replace `<ul>` element with a `<div>`, allowing us to remove redundant styles that suppressed the normal list display behaviour. Renamed element's class selector too to be more readible. 
- Re-enable .checkA11y() calls in IcPopoverMenu.cy.tsx, and add an exception to the one that failed for 'nested-interactive' rule

This pull request doesn't resolve the accessibility failure reported by Accessibility Insights, my reasoning being that when manually testing with NVDA and Voiceover, the toggle variant menu items read out fine. My understanding is that we'd need to remove the ic-button from inside ic-menu-item altogether and that's a lot of bespoke code for a fix that, IMO, isn't necessary. Please do disagree though!

## Related issue
#2722 

### Accessibility 

### DIFF
--- a/packages/react/src/component-tests/IcPopoverMenu/IcPopoverMenu.cy.tsx
+++ b/packages/react/src/component-tests/IcPopoverMenu/IcPopoverMenu.cy.tsx
@@ -21,10 +21,18 @@ import {
   PopoverWithMenuGroups,
 } from "./IcPopoverMenuData";
 import { IcTheme } from "../../components";
+import { CYPRESS_AXE_OPTIONS } from "../../../cypress/utils/a11y";
 
 const BUTTON_SELECTOR = "ic-button";
 const MENU_ITEM_SELECTOR = "ic-menu-item";
 const POPOVER_SELECTOR = "ic-popover-menu";
+
+const DISABLED_POPOVER_AXE_OPTIONS = {
+  rules: {
+    ...CYPRESS_AXE_OPTIONS.rules,
+    "nested-interactive": { enabled: false },
+  },
+};
 
 const DEFAULT_TEST_THRESHOLD = 0.015;
 
@@ -96,7 +104,7 @@ describe("IcPopoverMenu end-to-end, visual regression and a11y tests", () => {
     cy.checkHydrated(POPOVER_SELECTOR);
     cy.get(BUTTON_SELECTOR).click().wait(500);
 
-    // cy.checkA11yWithWait();
+    cy.checkA11yWithWait();
     cy.compareSnapshot({
       name: "/description",
       testThreshold: setThresholdBasedOnEnv(DEFAULT_TEST_THRESHOLD + 0.027),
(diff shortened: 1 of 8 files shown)


GCHQ-Developer-530:
Does the `<ul>` in popover menu need removing as well if we're removing the `<li>` from it's menu items?

GCHQ-Developer-299:
> Does the `<ul>` in popover menu need removing as well if we're removing the `<li>` from it's menu items?

@GCHQ-Developer-530  the `<ul>` shouldn't be removed as it's got the 'role="menu"' that's the parent to the 'menuitems' nested inside.  It's a good question and I will try out swapping it for a `<div>` or something though (since we're relying on aria roles, we might not need the semantic properties of a `<ul>`

GCHQ-Developer-299:
> > Does the `<ul>` in popover menu need removing as well if we're removing the `<li>` from it's menu items?
> 
> @GCHQ-Developer-530 the `<ul>` shouldn't be removed as it's got the 'role="menu"' that's the parent to the 'menuitems' nested inside. It's a good question and I will try out swapping it for a `<div>` or something though (since we're relying on aria roles, we might not need the semantic properties of a `<ul>`

@GCHQ-Developer-530 have swapped the `<ul>` for a div (and updated PR description) - while it doesn't seem to affect SR interaction thanks to aria-roles, it is cleaner to read and removed a couple lines of CSS.

GCHQ-Developer-530 (approved):

GCHQ-Developer-718 (changes requested):
I'm not sure whether it's a big deal but most menu examples I've come across seem to use the ul and li structure we had previously. This solution looks like it will be functionally fine as you're still using the menu and menu item roles but I just wanted to flag that it seems more common to get around this sort of issue by making the problematic list items presentational (as in role=presentation). I think this would also fix the accessibility insights failure.

GCHQ-Developer-299:
> I'm not sure whether it's a big deal but most menu examples I've come across seem to use the ul and li structure we had previously. This solution looks like it will be functionally fine as you're still using the menu and menu item roles but I just wanted to flag that it seems more common to get around this sort of issue by making the problematic list items presentational (as in role=presentation). I think this would also fix the accessibility insights failure.

@GCHQ-Developer-718  i've had a go at using role="presentation" but it doesn't resolve the issue. According to this [github issue conversation](https://github.com/dequelabs/axe-core/issues/2906) there's no way to prevent the 'button' element inside 'ic-button' registering as an interactive element. I'm more confused as to why all the other menu-items aren't triggering this issue, and why it's only the checkbox one. 

I don't think, without removing ic-button, we can remove this reported error.

GCHQ-Developer-718 (approved):

GCHQ-Developer-718:
> > I'm not sure whether it's a big deal but most menu examples I've come across seem to use the ul and li structure we had previously. This solution looks like it will be functionally fine as you're still using the menu and menu item roles but I just wanted to flag that it seems more common to get around this sort of issue by making the problematic list items presentational (as in role=presentation). I think this would also fix the accessibility insights failure.
> 
> @GCHQ-Developer-718 i've had a go at using role="presentation" but it doesn't resolve the issue. According to this [github issue conversation](https://github.com/dequelabs/axe-core/issues/2906) there's no way to prevent the 'button' element inside 'ic-button' registering as an interactive element. I'm more confused as to why all the other menu-items aren't triggering this issue, and why it's only the checkbox one.
> 
> I don't think, without removing ic-button, we can remove this reported error.

Makes sense, I've approved. :)

PR #3572, opened 2025-05-22
fix(web-components): removed icChange event from componentWillLoad

GCHQ-Developer-741:
## Summary of the changes
- Removed icChange call from componentWillLoad by extracting out other necessary code into seperate function.
- Cleared IDE warnings for missing prop values


GCHQ-Developer-299 (approved):
Confirmed using the [Controlled story](https://mi6.github.io/ic-ui-kit/branches/remove-icchange-from-willload/react/?path=/story/react-components-search-bar--controlled) that the icChange event no longer fires on first render.  The refactors look good too :)

GCHQ-Developer-530 (approved):

PR #3593, opened 2025-06-02
fix(web-components): allow horizontal radio group inside checkbox and…

GCHQ-Developer-718:
… fix disabled textfield bug

Horizontal radio group variant displays correctly when nested inside a checkbox. Fix bug that causes only the first text field nested inside a checkbox to be in the disabled state when the checkbox is unchecked.

### Previous behaviour

- A nested radio always appears as the vertical variant
- If multiple textfields are nested under the same checkbox and the checkbox is unchecked, only the first is in the disabled state

previous behaviour of checkboxes with nested radio groups and textfields at large viewport widths

### New behaviour

All nested textfields are disabled when the parent checkbox is unchecked
All nested textfields are disabled when the parent checkbox is unchecked

#### Large viewport widths
New behaviour at large viewport widths - radio groups in horizontal variant (conditional static)
New behaviour at large viewport widths - radio groups in horizontal variant (conditional dynamic)

#### Medium viewport widths
New behaviour at medium viewport widths - radio groups with longer options in vertical variant (conditional static)

#### Small viewport widths
New behaviour at small viewport widths - all radio groups in vertical variant (conditional static)
New behaviour at small viewport widths - all radio groups in vertical variant (conditional dynamic)

## Related issue
Tell us the issue number. If suggesting an improvement or component, please discuss it with us in an issue first.

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes

### DIFF
--- a/packages/react/src/stories/ic-checkbox.stories.js
+++ b/packages/react/src/stories/ic-checkbox.stories.js
@@ -6,6 +6,12 @@ import {
   IcCheckbox,
   IcTextField,
   IcButton,
+  IcRadioGroup,
+  IcRadioOption,
+  IcSelect,
+  IcSearchBar,
+  IcAlert,
+  IcChip,
 } from "../components";
 import { useForm } from "react-hook-form";
 import React, { useState, useRef } from "react";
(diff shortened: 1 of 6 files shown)


ad9242 (changes requested):
I think it would be nice to add some of these cases into the storybook examples, so we can quickly see them in action. Would be good to enhance the tests to to check the disabled state of slotted elements (this could be in the unit tests or Cypress tests, either checking the disabled attribute on the element, or with visual regression tests)

A separate issue is why only child textfields get disabled and not any other type of control, but that's one for another time\for design

ad9242 (approved):

PR #3619, opened 2025-06-06
2788 internal issue 2923   icsearchbar submitting in form element

GCHQ-Developer-718:
## Summary of the changes
Adds preventFormsSubmitOnSearch prop to ic-search-bar.

If preventFormSubmitOnSearch is true, the parent form (if present) will not submit when icSubmitSearch fires. If preventFormSubmitOnSearch is false (default option) the parent form (if present) will submit when icSubmitSearch fires.

A new unit test verifies that, when preventFormSubmitOnSearch is true, the search bar's search submit button has `type="button"` and, when clicked, the lightDOM button that is created inside the form and clicked also has `type="button"`.

Another test verifies that, when preventFormSubmitOnSearch is not set (and so defaults to false), the search bar's search submit button has `type="submit"` and, when clicked, the lightDOM button that is created inside the form and clicked also has `type="submit"`.

The "In a form" story is updated to demonstrate the behaviour of an ic-search-bar in a form, with and without preventFormSubmitOnSearch.

## Related issue
#2788 (internal issue 2923)

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes


GCHQ-Developer-299:
@GCHQ-Developer-718  there's a conflict in the search-bar unit test file, will review once you've resolved that :)

packages/web-components/src/components/ic-search-bar/ic-search-bar.tsx
```diff
@@ -914,7 +928,7 @@ export class SearchBar {
                 onBlur={this.handleSubmitSearchBlur}
                 onFocus={this.handleSubmitSearchFocus}
                 onKeyDown={this.handleSubmitSearchKeyDown}
-                type="submit"
+                type={getButtonType()}
```
ad9242, line 931: just a slight optimisation suggestion. At the mo, the `getButtonType` function will get called every time the component renders, so whenever another unrelated prop changes, it will result in a call to this function. 

As a optimisation, you could have the button type as a state variable, initialised in the `componentWillLoad` & then place a watch on the `preventFormSubmitOnSearch` for when it changes, to update the value

ad9242:
Just needs a docs scope commit for the docs changes, but otherwise happy to approve.

GCHQ-Developer-718:
> Just needs a docs scope commit for the docs changes, but otherwise happy to approve.

Done. :)

ad9242:
> @ad9242, do you think we need a guidance site issue to make an example similar to the new story I made, to demonstrate the options for using the search bar in a form?

yes, would be good to add an example of the new behaviour

GCHQ-Developer-718:
> > @ad9242, do you think we need a guidance site issue to make an example similar to the new story I made, to demonstrate the options for using the search bar in a form?
> 
> yes, would be good to add an example of the new behaviour

Okay, I've created a new issue for that.

GCHQ-Developer-299:
@GCHQ-Developer-718  a couple more file conflicts!

Issue #2975, opened 2025-01-06
Circular loading indicator not displaying when used inside an ic-dialog


GCHQ-Developer-741:
## Summary of the bug 
When placing a circular loading indicator inside an ic-dialog, the `offsetWidth` of the element is set to 0 (since it is using `display: none`), which prevents the `circularLineWidth` & `circularDiameter` from being initialised when the component is rendered (it only runs the function during `componentDidLoad`. This means that the circle x & y variables are also not set, leading to the loading indicator circle elements not being visible when the dialog is eventually opened.

## 🪜 How to reproduce 
Copy this code example:
```tsx
import { IcButton, IcDialog, IcLoadingIndicator } from "@ukic/react";
import { useState } from "react";

export default function App() {
  const [showDialog, setShowDialog] = useState(false);
  return (
    <div style={{ padding: "8px" }}>
      <IcButton onClick={() => setShowDialog(true)}>Open Dialog</IcButton>
      <IcDialog
        heading=""
        open={showDialog}
        onIcDialogClosed={() => setShowDialog(false)}
        size="small"
      >
        <IcLoadingIndicator label="Loading..." type="circular" />
      </IcDialog>
    </div>
  );
}
```

### Initialisation during `componentDidLoad`
![image](https://github.com/user-attachments/assets/9c5735dc-2c98-430a-9c5c-f9eb864834c5)

### Using the values during the render
![image](https://github.com/user-attachments/assets/99afe957-615a-44a4-9f1e-0240c27d9827)

## 🧐 Expected behaviour 
Either the loading-indicator should reset the circular values when a dialog is displayed and therefore has width, or it should set some default values to avoid this situation.

Issue #1878, opened 2024-05-17
[IcLoadingIndicator] - changing `type` from `linear` to `circular` does not work


ad9242:
## Summary of the bug 
If the type is changed from `linear` to `circular`, then the indicator does not display correctly

## 🪜 How to reproduce 
1. go to [stackblitz example](https://stackblitz.com/edit/stackblitz-starters-7m9vhg?file=src%2FApp.tsx)
2. see the linear loading indicator
3. click the button to change the type
4. See the circular indicator is not displayed

## 🧐 Expected behaviour 
the circular indicator should display correctly

## 📝 Acceptance Criteria  

*Given* a linear loading indicator is visible
*When*  changing to a circular type
*Then* the circular indicator should display correctly

PR #3622, opened 2025-06-06
2975 circular loading indicator not displaying when used inside an ic dialog

GCHQ-Developer-741:
## Summary of the changes
- Updated logic behind setting the diameter during `componentDidLoad`. Now if the `offsetHeight` is 0, then it will get the css defined width set by the `--circular-diameter` value, then convert that to a pixel value, and if not then falls back to a map of values based on the indicator's size. **NOTE:** Added a clarification to the CSS prop description to specify to use `rem`, `px` or `%` units.
- Updated logic around undefined values to remove IDE warnings.
- Updated and added unit/cypress tests for behavior where relevant.
- Fixed minor bug where switching the type to circular wasn't rendering because the state was not being repopulated after initial render. Now checks the diameter when the type changes and if different, updates state.

## Related issue
#2975 
#1878 

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes


GCHQ-Developer-741:
> When setting the progress to 0 for a determinate variants (linear & circular), they animate. Is that the correct behaviour?
> 
> I think I was expecting them to just show as empty\zero progress (there was an issue where the colour was wrong: #3520 and the linear was appearing as filled, which @srt544 was currently looking at)

Should be back in the state it was on develop now.

ad9242 (dismissed):

GCHQ-Developer-299 (approved):

Issue #3512, opened 2025-05-01
Implement performance tests


MI6-286:
## Summary
Implement performance tests for components using cypress-audit/lighthouse.

## 💬 Description 
The setup for performance testing of components was previously completed in a separate task. This ticket is to create the performance tests for components (beginning with larger components) so that any issues with performance can be highlighted for potential improvements.

## Why do we need it? 
To ensure our components have optimal loading times and responsiveness, improving the user experience.

evenstensberg:
A good first step is to add [bundle size](https://github.com/siddharthkp/bundlesize) to the CI step that checks if your file size has regressed.

evenstensberg:
Lighthouse also has a [CI](https://github.com/GoogleChrome/lighthouse-ci) option!

evenstensberg:
Sorry for spamming, but let me know if you need assistance with anything lighthouse related. I’ve contributed to the codebase earlier and know a trick or two.

PR #3653, opened 2025-06-13
3512 update GitHub Actions for performance testing

MI6-286:
## Summary of the changes

Currently, the performance tests are set to run on each of the branches, develop, and release-merge workflows. This PR introduces new workflow files so that the performance tests are run separately; one workflow runs the tests every night against the develop branch and one has a manual trigger.

I’ve tested the subset of performance tests I’ve written so far (not yet on develop) on a forked repo and the workflow takes about 30 minutes to run. It’s taking longer than it should be due to the number of failures and re-runs (more on this to come), but either way it would add a lot of time to the pipelines. As performance results are unlikely to vary much with small code changes, I thought it best to use a scheduled workflow instead.

The reason for the two separate workflow files (scheduled and manual) is that scheduled workflows have to be from the default branch, and then the develop branched checked out to run the tests against. Therefore, then need a separate workflow for the manual trigger to allow it to be run from any branch (https://github.com/orgs/community/discussions/26374).

I’ve tested both workflows on a forked repo and all seems to be working as expected. The notification of the scheduled run completion goes to the creator/ last updater of the workflow – I don’t think there’s an easy way to change this currently (https://github.com/orgs/community/discussions/18039). I’ve checked that I get the notification and that the reports are uploaded to the expected location.

## Related issue
#3512

### DIFF
--- a/.github/workflows/run-performance-tests-manual.yml
+++ b/.github/workflows/run-performance-tests-manual.yml
@@ -1,0 +1,60 @@
+name: Run performance tests - manual trigger
+permissions: read-all
+
+on: 
+    workflow_dispatch:
+
+concurrency:
+  group: ci-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  check-updates:
+    name: "Check package updates"
+    runs-on: ubuntu-latest
+    outputs:
+      core: ${{ steps.filter.outputs.core }}
+    steps:
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 #v4.1.1
+      - uses: dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36 #v3.0.2
+        id: filter
+        with:
+          filters: |
+            core:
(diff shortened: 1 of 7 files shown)


.github/workflows/run-performance-tests-manual.yml
```diff
@@ -0,0 +1,60 @@
...
+          node-version: 20
+      - uses: browser-actions/setup-chrome@97349de5c98094d4fc9412f31c524d7697115ad8 #v1.5.0
+
+      - name: Extract branch name
+        run: echo "branch=${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}}" >> $GITHUB_OUTPUT
+        id: extract_branch
+
+      - name: Install dependencies
+        run: |
+          google-chrome --version
```
evenstensberg, line 46: Is this needed? Also make sure you're running headless chrome, it will be significantly faster.

MI6-286: It's not needed but useful for debugging and is in keeping with the other workflow files. Thanks, the tests are configured to run in headless mode.

Issue #3648, opened 2025-06-12
[ic-navigation-button] - hover state not visible on dark background


ad9242:
### Summary of the bug 
The hover state for ic-navigation buttons on a dark background is not (or barely) visible, and does not match other things like navigation items, app title link. 

![image](https://github.com/user-attachments/assets/341eabd5-76e9-41a8-99dc-f55e2ce0564a)
![image](https://github.com/user-attachments/assets/f8caf85d-2f0f-425c-9ece-d41c158a2ce2)

It is fine on a lighter background

![image](https://github.com/user-attachments/assets/94408e72-3fb0-497f-b9f5-0d110b933aec)

### 🪜 How to reproduce 
Tell us the steps to reproduce the problem: 
1. Go to [storybook example](https://mi6.github.io/ic-ui-kit/branches/develop/web-components/?path=/story/web-components-top-navigation--with-navigation) 
2. Hover over the navigation button in the top right (tooltip: "Button 1")
3. See the hover styling is not visible

### 🧐 Expected behaviour 
The hover and pressed state styling should match other elements eg navigation items\ app title link

### 🚨 Urgency (low, medium or high)
Low

mi6-577:
Made some changes in Figma;

1. Aligned LM & DM tokens for tertiary button to secondary button for consistency:
button-tertiary-background-hover-active		action-bg-hover
button-tertiary-background-hover-pressed		action-bg-pressed

2. Created new tokens for state text and icons for destructive buttons - they were inconsistent with other buttons - and applied them to destructive buttons:
button-destructive-text-hover
button-destructive-icon-hover
button-destructive-text-pressed
button-destructive-icon-pressed
button-destructive-icon-pressed-active

3. Created missing token for primary loading button and applied to loading state buttons (primary);
button-primary-icon-pressed-active			icon/inverted (LM)		icon/neutral (DM)


4.Amended DM tokens on the following:
button-secondary-background-hover-monochrome		action-bg-light-hover
button-secondary-background-pressed-monochrome	action-bg-light-pressed
button-tertiary-background-hover-monochrome		action-bg-light-hover
button-tertiary-background-pressed-monochrome		action-bg-light-pressed

PR #3654, opened 2025-06-13
fix(web-components): updated hover/pressed colours for tertiary/nav b…

GCHQ-Developer-741:
## Summary of the changes
- Updated the secondary & tertiary background hover/pressed colours to correct values in dark theme.
- Updated nav-button to use the icon-tertiary variant, which felt more appropriate. Also removed icon margin from other icon variants.
- Now updates and applies the theme to the nav-button host element depending on the brand, as it wasn't passing down correct theme colours to the ic-button which was causing some issues.
- Updated tests to reflect new changes.

Changes can be viewed on the brand story.

## Related issue
#3648 

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes


GCHQ-Developer-299 (changes requested):
Looks like the CSS needs a little tweaking:

- on develop there's a white outline in light mode on the navbar button
Screenshot 2025-06-17 at 13 37 42

- on your branch the outline is black in light mode, seemingly using dark mode styles
Screenshot 2025-06-17 at 13 37 57

GCHQ-Developer-741:
> Looks like the CSS needs a little tweaking:
> 
> * on develop there's a white outline in light mode on the navbar button
> 
> Screenshot 2025-06-17 at 13 37 42
> * on your branch the outline is black in light mode, seemingly using dark mode styles
> 
> Screenshot 2025-06-17 at 13 37 57

Should be up to date now. Since its now pulling through all the other correct hover/pressed stylings based on theme, I couldn't think of a way to revert that so it used the correct focus indicator without affecting those other areas, so I've manually set the light focus indicator on the component.

GCHQ-Developer-718 (approved):

GCHQ-Developer-741:
> > > Looks like the CSS needs a little tweaking:
> > > 
> > > * on develop there's a white outline in light mode on the navbar button
> > > 
> > > Screenshot 2025-06-17 at 13 37 42
> > > * on your branch the outline is black in light mode, seemingly using dark mode styles
> > > Screenshot 2025-06-17 at 13 37 57
> > 
> > 
> > Should be up to date now. Since its now pulling through all the other correct hover/pressed stylings based on theme, I couldn't think of a way to revert that so it used the correct focus indicator without affecting those other areas, so I've manually set the light focus indicator on the component.
> 
> I'm still hesitant to Approve because this is a regression from Develop (where those buttons correctly apply themed focus indicators) - on your branch the Search Bar Screenshot 2025-06-18 at 12 40 56 still has a dark focus in dark mode so these buttons would now not align visually:
> 
> could the `getBrandFromContext` function in helpers.ts be a good reference? If this is too much hassle for a minor detail, i would be okay with you dismissing my review.

I've added a change to check the parent's theme (inherited or manually set, using a helper function). I don't think its possible to follow the getBrandFromContext method as the theme prop needs to be set on the ic-navigation-button host element AND passed down to the other ic-button element, otherwise the styling and colours do not get applied correctly as it cannot recognise the theme through multiple shadow DOM layers seemingly. That's why I have to manually set the focus indicator, as leaving it means its inheriting the dark one which is obviously different to how it is on develop, but more true to what is happening to get the other theme colours to apply.

GCHQ-Developer-847 (approved):

PR #3659, opened 2025-06-16
2806 side navigation static prop does not work with disabletopbarbehaviour

GCHQ-Developer-299:
## Summary of the changes

Storybook changes:
- Added static/disableTopBarBehaviour/expanded props to react storybook
- Added playground story to web-components storybook

New behaviour:
- DisableTopBarBehaviour, when forcing the sideNav to display as normal, now acknowledges 'static' prop setting

## Related issue
#2806

### General 

### Testing

### Testing content extremes

### DIFF
--- a/packages/web-components/src/components/ic-side-navigation/ic-side-navigation.tsx
+++ b/packages/web-components/src/components/ic-side-navigation/ic-side-navigation.tsx
@@ -262,9 +262,7 @@ export class SideNavigation {
   };
 
   private toggleMenuExpanded = (expanded: boolean): void => {
-    if (this.deviceSize > DEVICE_SIZES.S) {
-      this.menuExpanded = expanded;
-    }
+    this.menuExpanded = expanded;
 
     if (this.menuExpanded) {
       this.setAndRemoveNoWrapAfterMenuExpanded();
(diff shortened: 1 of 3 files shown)


MI6-286 (dismissed):

packages/web-components/src/components/ic-side-navigation/ic-side-navigation.tsx
GCHQ-Developer-847, line 1: Just wondering if we definitely want the expand button to no longer appear on smaller device sizes when `static` is enabled. I personally thought it was useful for users because being able to close it on small screens would give them more room to look at the page content

Also I believe the side nav overlaps content on smaller screens so removing the expand button means they have no way of seeing the content it overlaps

GCHQ-Developer-299: I get where you're coming from - and that does explain why the limitation was in place in the first place (i read through the guidance docs, couldn't find a reason for why that screen width logic was in there!) 
I will revert that part of the change

MI6-286 (commented):
Tested on storybook playground and I'm getting the same behaviour as on develop now - the static prop isn't taking effect when disableTopBarBehaviour is set.

GCHQ-Developer-299:
> Tested on storybook playground and I'm getting the same behaviour as on develop now - the static prop isn't taking effect when disableTopBarBehaviour is set.

As per @GCHQ-Developer-847 's review, i've reverted Static so that it doesn't take effect at all on smaller breakpoints (meaning that users are always able to minimise the sideNav on small screenwidths) 
But you're right, this means that, while disableTopTabBehaviour and Static can be turned on at the same time, they'll never take effect together. 

I think the PR is valuable (the change to allow the menuExpanded to function at any width fixes a bug with disableTopBarBehaviour) but you might have identified that the ticket is intended behaviour rather than an issue?

MI6-286:
> > Tested on storybook playground and I'm getting the same behaviour as on develop now - the static prop isn't taking effect when disableTopBarBehaviour is set.
> 
> As per @GCHQ-Developer-847 's review, i've reverted Static so that it doesn't take effect at all on smaller breakpoints (meaning that users are always able to minimise the sideNav on small screenwidths) But you're right, this means that, while disableTopTabBehaviour and Static can be turned on at the same time, they'll never take effect together.
> 
> I think the PR is valuable (the change to allow the menuExpanded to function at any width fixes a bug with disableTopBarBehaviour) but you might have identified that the ticket is intended behaviour rather than an issue?

Makes sense to not have static set at smaller breakpoints regardless of disableTopBarBehaviour, but do we want to allow disableTopBarBehaviour and static to both work at the larger break points?

MI6-286 (approved):

PR #3666, opened 2025-06-18
2302 data table column hiding

ad9242:
## Summary of the changes
Adds support for hiding columns by way of new `hidden` column property - defaults to `false`

## Related issue
#2302 

### General 

### Testing


GCHQ-Developer-530 (dismissed):

packages/canary-web-components/src/components/ic-data-table/ic-data-table.types.tsx
```diff
@@ -40,6 +40,7 @@ export type IcDataTableColumnObject = {
     hideOnHeader?: boolean;
   };
   excludeColumnFromSort?: boolean;
+  visible?: boolean;
```
GCHQ-Developer-741, line 43: Should this option be called `hidden` instead? Seems like all the examples talk about "hiding" columns, and the default behavior currently is that columns are visible unless otherwise specified. Feel like that exception should be to "hide" the column by setting it to true, that way leaving the option blank is the same as setting it to `false`, which I'd probably expect as a developer.

ad9242: I was torn as to the behaviour. I think you are correct though, `hidden` does make more sense - i'll make the change

GCHQ-Developer-741 (approved):

Issue #3465, opened 2025-04-23
Add missing global css tokens to our styles


GCHQ-Developer-741:
## Summary 
The ic-ui-kit features several top-level "global" css tokens in the `color-mode.css` file, that enable quick switching between light and dark mode for users who only need to implement one token.

## 💬 Description 
The top-level of the `color-mode.css` features several global tokens that were originally intended to make it easy to switch between light and dark mode. E.G: We could reference `--ic-color-text-primary` for all primary text in components, rather than needing to specify `-light` or `-dark`. However, there are many of our semantic colour pairs that do not have a global token. These missing global tokens need to be added, to allow users to have a full range of tokens to theme their applications with.

![Image](https://github.com/user-attachments/assets/3f966e95-bd00-4bb9-b714-6ad5a603cd73)

This will likely come with a need to rename certain semantic tokens to not clash and confuse users by having similar names. This will also require adjustments in Figma to ensure consistency.

![Image](https://github.com/user-attachments/assets/2c6cb063-3e1d-42bc-91e1-082fc7fc8860)
![Image](https://github.com/user-attachments/assets/87771005-7579-4775-b8e6-8838d1e69323)

## 💰 Use value 
Users will be able to customize their own components/applications without needing to implement our complex theme switching.

## Additional info 
This may constitute a breaking change, if we think users have pulled certain tokens into their own applications that may need renaming.

We would be best served to document all these new global tokens on the [colour page](https://design.sis.gov.uk/styles/colour) so that users know what the purpose of each token is so they can use them correctly in their applications.

Certain semantic tokens don't have a light/dark variant, so will need to decide closer to the time which ones need adding and what they should be called.

GCHQ-Developer-741:
To review at the time of V4, when we may potentially have a clearer understanding of how theming will work, which would alter how we approach this ticket and these tokens.

MI6-255:
Separate ticket to implement these into our own components

PR #3683, opened 2025-06-24
feat(web-components): added new global colour tokens

GCHQ-Developer-741:
## Summary of the changes
added new tokens that automatically switch theme, to represent our full range of semantic tokens. This is based on the colours documented in the semantic section of our [colours page](https://design.sis.gov.uk/styles/colour/#semantic), where I have essentially taken any semantic tokens that didn't have a global one and added them. A [ticket](https://github.com/mi6/ic-design-system/issues/1561) has been raised to actually document these on the site.

Note that these global tokens don't always follow our typical naming conventions (particularly the status colours due to current primitive tokens taking these names in the past), which is confusing but I have tried to name them as descriptively as possible. It would likely be a breaking change to rename these now, so would likely be a V4 problem. Obviously when this is documented on the site some of this confusion will be avoided, but if anyone has better naming suggestions I'm all ears.

## Related issue
#3465


GCHQ-Developer-741:
> should using these new tokens in the components be covered in this ticket, or a separate one?

There is a [comment](https://github.com/mi6/ic-ui-kit/issues/3465#issuecomment-2900400095) on the ticket to do that in a separate ticket, albeit that hasn't been created yet. I will do that now.

PR #1557, opened 2024-02-22
Canary/date input

jd239:
## Summary of the changes
- Added new github actions for Cypress this includes:
  - Running Cypress tests only
  - Updating Cypress visual regression snapshots via CI
  - Uploading visual regression report if there are any errors
  - Updating folders names due to visual regression v2 update
- Threshold helper function which allows a 0 threshold value for CI but an updated threshold for local testing
- Date Input implementation for Canary release
- Jest/Puppeteer to Cypress component tests
- Updated ic-input-component-container for additional 'large' size
- Removal of screenshots from v1 visual regression within canary-web-components

## Related issue
#1364 

## Checklist
- [x] Relevant unit tests and visual regression tests added. 
- [x] Visual testing against Figma component specification completed. 
- [x] All acceptance criteria reviewed and met. 
- [x] Accessibility Insights FastPass performed.
- [x] A11y unit test added and yields no issues.
- [x] A11y plug-in on Storybook yields no issues. 
- [x] Manual screen reader testing performed using NVDA and VoiceOver. 
- [x] Page can be zoomed to 400% with no loss of content. 
- [x] Screen magnifier used with no issues. 
- [x] Text resized to 200% with no loss of content.
- [x] Text spacing increased as per the [WCAG 1.4.12 success criterion](https://www.w3.org/TR/WCAG21/#text-spacing) with no loss of content.
- [x] Browser setting 'prefers reduced motion' tested. No animations or motion visible whilst this setting is on. 
- [x] Windows High Contrast mode tested with no loss of content.
- [x] System light and dark mode tested with no loss of content. 
- [x] Manual keyboard testing for keyboard controls and logical focus order. 
- [x] Min/max content examples tested with no loss of content or overflow. 
- [x] Browser support tested (Chrome, Safari, Firefox and Edge).
- [x] Correct roles used and ARIA attributes used correctly where required. 
- [x] Logical heading structure is maintained, and the HTML elements used for headings can be changed to fit within the wider page structure. 
- [x] All prop combinations work without issue.

PR #3761, opened 2025-07-14
752 Table of Contents

GCHQ-Developer-299:
## Summary of the changes
Initial implementation of Table of Contents as a canary component, developed by @cd3859 and finished off by @GCHQ-Developer-299 

## Related issue
#752 

## Bugs i'm aware of

- [x] Theme prop doesn't seem to work (and dark mode cypress test fails accessibility)
- [x] You can't set maximumHeadingLevel below 2 or it breaks
- [x] Truncate=false causes toc-items to overlap horribly with their siblings
- [x] Truncation ellipsis ... only applies on large screensizes, doesn't work when TOC is positioned 'top' either. Tooltip doesn't appear at any point.
- [x] Position "top" doesn't change the TOC's position dynamically in storybook, requiring a full page refresh.
- [x] Logic that renders a tooltip when TOC-item is truncated doesn't seem to work.
- [x] The scrolling sidebar isn't tall enough when size is 'large' 
- [x] At zoom levels other than 100%, the TOC doesn't vertically offset correctly (this is a Storybook issue apparently. Sticky position doesn't work correctly for Page Header either for example. it's all good when you open in new tab) 
- [x] Clicking a TOC item doesn't move focus to the chosen heading
- [x] The TOC appears after the page body in the tab order, which is incorrect. 
- Modifying maximumHeadingLevel causes unusual behaviour with the sidebar pill (i don't know whether this is a priority to fix - is anyone gonna be updating that setting dynamically? And it's fixed as soon as the user clicks an item and the page updates)

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes

- [x] Props/slots can be updated after initial render. (MOSTLY - see "bugs i'm aware of")


packages/canary-react/src/component-tests/IcTableOfContents/IcTableOfContents.cy.tsx
```diff
@@ -0,0 +1,116 @@
...
+    cy.checkHydrated(TABLE_OF_CONTENTS);
+
+    cy.checkA11yWithWait();
+    cy.compareSnapshot({
+      name: "/default",
+      testThreshold: setThresholdBasedOnEnv(DEFAULT_TEST_THRESHOLD + 0.007),
+    });
+  });
+
+  it("should render nested", () => {
```
evenstensberg, line 37:
```suggestion
  it("should render nested elements", () => {
```

evenstensberg (commented):
just a small nitpick, nice work!

packages/canary-web-components/src/components/ic-table-of-contents-item/ic-table-of-contents-item.tsx
GCHQ-Developer-530, line 1: Could we update this component to match our usual template? Aware it's only an internal component, but there's no prop descriptions etc.

packages/canary-web-components/src/components/ic-table-of-contents/ic-table-of-contents.css
```diff
@@ -0,0 +1,135 @@
...
+  --ic-table-of-contents-hover: var(--ic-state-layers-action-bg-hover);
+  --ic-table-of-contents-hover-monochrome: var(--ic-state-layers-action-bg-dark-hover);
+  --ic-table-of-contents-pressed: var(--ic-state-layers-action-bg-pressed);
+  --ic-table-of-contents-pressed-monochrome: var(--ic-state-layers-action-bg-dark-pressed);
+   */
+
+:host {
+  display: block;
+  border: none;
+  max-height: 100vh;
```
GCHQ-Developer-530, line 14: Screenshot 2025-08-08 at 12 31 09

Could we add a z-index to table of contents? I think that would fix the tooltip going behind the content.

packages/canary-web-components/src/components/ic-table-of-contents-item/ic-table-of-contents-item.css
```diff
@@ -0,0 +1,63 @@
...
+  --ic-table-of-contents-hover-monochrome: var(--ic-state-layers-action-bg-dark-hover);
+  --ic-table-of-contents-pressed: var(--ic-state-layers-action-bg-pressed);
+  --ic-table-of-contents-pressed-monochrome: var(--ic-state-layers-action-bg-dark-pressed); */
+
+.table-of-contents-item-content {
+  --ic-typography-color: var(--ic-table-of-contents-text);
+
+  border: none;
+  position: relative;
+  min-height: 1.5rem;
```
GCHQ-Developer-530, line 15: I think this min-height needs updating to match the Figma height. Looks like at the moment the total height (including padding) is 28, but it should be 32.
Screenshot 2025-08-08 at 12 40 29
Screenshot 2025-08-08 at 12 40 43

GCHQ-Developer-530 (commented):
Looking good! Just noticed a couple of things

packages/canary-web-components/src/components/ic-table-of-contents/ic-table-of-contents.tsx
```diff
@@ -0,0 +1,361 @@
...
+      />
+    );
+
+    this.listContent.push(item as HTMLElement);
+    children.push(item);
+    return children;
+  };
+
+  // generate list items from the headings in the page
+  private generateListItems(): HTMLElement[] {
```
GCHQ-Developer-718, line 110: It feels like nested ordered lists should be used when the heading level changes as a non-visual equivalent to the indent. As in:
```
<ol>
  <li><a>My h2</a></li>
    <ol>
      <li><a>My h3</a></li>
      <li><a>My other h3</a></li>
      <ol>
        <li><a>My h4</a></li>
      </ol>
    </ol>
  <li><a>My other h2</a></li>
  <li><a>Another h2</a></li>
</ol>
```

Maybe this is a question for a screen reader user though as I appreciate it could be annoying if there are a lot of nested lists with very few items, so maybe real users would prefer the list to be flattened. I'm just thinking that, if the heading level is important enough information to be communicated visually, it should have a non-visual alternative.

If users do prefer having a flattened list for quicker navigation, you could consider including the heading level in a visually hidden span so they still have access to that information. That's not something I've seen before though.

packages/canary-web-components/src/components/ic-table-of-contents/ic-table-of-contents.tsx
```diff
@@ -0,0 +1,361 @@
...
+
+  // add margin to the list based off the header element height
+  private addListMargin(): void {
+    const content = this.el.shadowRoot?.querySelector(
+      `.${TOC_CONTENT_CLASS}`
+    ) as HTMLDivElement;
+    content?.style.setProperty("top", `${this.headerHeight}px`);
+  }
+
+  // add tooltip to the item if it is too wide
```
GCHQ-Developer-718, line 265: For some reason, the tooltip isn't appearing for me when a truncated list item is focused using the keyboard.

GCHQ-Developer-718 (commented):
Looks really good! I also just noticed a couple of things.

packages/canary-web-components/src/components/ic-table-of-contents-item/ic-table-of-contents-item.tsx
GCHQ-Developer-299: As discussed internally, i'll instead move toc-item inside the overall toc folder and keep it as a FunctionalComponent akin to ic-date-picker/ic-date-button

packages/canary-web-components/src/components/ic-table-of-contents-item/ic-table-of-contents-item.css
GCHQ-Developer-299: Looking at the figma designs, it looks like the 32px height is for the Large variant.
Regular:
Screenshot 2025-08-11 at 11 50 40


Large:
Screenshot 2025-08-11 at 11 50 54

packages/canary-web-components/src/components/ic-table-of-contents/ic-table-of-contents.tsx
GCHQ-Developer-299: I am really struggling to work this out in the code, will reach out to the dev team to see if anyone can think of how to generate a nested list. I think it might need an extensive refactor

GCHQ-Developer-718:
> > Could you add an extra story showing how the table of contents works with an inline skip link like tree view? Just to verify that everything looks and behaves okay.
> 
> @GCHQ-Developer-718 Because the table of contents wraps the page content, it's not suited to having a skip link put before or inside it right now (see screenshots). Adding skip link functionality would require adding a prop to conditionally render one in there and this is outside the Acceptance Criteria so I think it should be raised as a ticket for Feature Improvement
> 
> Skip link before treeview Screenshot 2025-08-11 at 15 26 24
> 
> Skip link as child of treeview Screenshot 2025-08-11 at 15 24 38

Thanks for checking, I'm happy with a new ticket being raised. :)

GCHQ-Developer-718 (approved):
Looks great!

ad9242:
This should not be merged until https://github.com/mi6/ic-design-system/issues/768 is complete

ad9242:
seems to be an issue in the following story:

https://mi6.github.io/ic-ui-kit/branches/752-table-of-contents/canary-react/?path=/story/react-components-table-of-contents--default-top

ad9242:
more comment than request for anything to be changed, but the tooltips seem a little redundant as they add no additional info, they just repeat the text:

image

GCHQ-Developer-299:
> seems to be an issue in the following story:
> 
> https://mi6.github.io/ic-ui-kit/branches/752-table-of-contents/canary-react/?path=/story/react-components-table-of-contents--default-top

Fixed I think - this was because i'd used web-component syntax for the styles tags, switched it to JSX and it should be fine now

GCHQ-Developer-299:
> Sorry another that is maybe more of a design decision: when the ToC is at the top, I feel like that part should be sticky, otherwise as soon as you scroll, you can no longer see it, so what is the point of it? I suppose it offers quick initial navigation, but after that you would have to return to the top to move between page sections
> 
> image

Looking at the ticket, the TOC scrolling with the page when positioned top is correct:

> Given that a table of contents exists on a page
When the table of contents is positioned to the top of the page
Then the table of contents scrolls with the page content.

Contrasted with side position:

> Given that a table of contents exists on a page
When the table of contents is positioned to the side of the page
Then the table of contents is sticky on the page content when scrolling

GCHQ-Developer-299:
> > more comment than request for anything to be changed, but the tooltips seem a little redundant as they add no additional info, they just repeat the text:
> > image
> 
> @ad9242 that is a bug though - the tooltips shouldn't be showing when the text isn't truncated.

@ad9242 i've addressed this with a fix to icTooltip's hide method. Should now behave as expected :)

packages/canary-web-components/src/utils/types.ts
```diff
@@ -301,3 +301,5 @@ export enum IcWeekDays {
   Friday = 5,
   Saturday = 6,
 }
+
+export type IcPosition = "top" | "right";
```
ad9242, line 305: Can we make this more specific, eg IcPositionTopOrRight or even IcTableOfContentsPosition? IcPosition sounds like a name that could be useful more generally across other components in future.  This way it would not be a breaking change if we want to have an IcPosition further down the line which also supports "bottom" or "left"

packages/canary-web-components/src/components/ic-table-of-contents/ic-table-of-contents.tsx
```diff
@@ -0,0 +1,442 @@
...
+
+  /**
+   * The size of the table of contents items.
+   */
+  @Prop() size: IcSizes = "medium";
+
+  /**
+   * Whether the table of contents will appear on the right of or above child content on large viewport widths.
+   */
+  @Prop() position: IcPosition = "right";
```
ad9242, line 47: i think this needs to be `mutable` as it complains about the value being changed in the unit tests

packages/canary-web-components/src/components/ic-table-of-contents/ic-table-of-contents.tsx
```diff
@@ -0,0 +1,442 @@
...
+  }
+
+  // Component is a wrapper for page contents, and will build list from heading items
+  render() {
+    const { heading, monochrome, position, size, theme } = this;
+    return (
+      <Host class={{ [`ic-theme-${theme}`]: theme !== "inherit" }}>
+        <ic-section-container
+          id={"table-of-contents-section-container"}
+          aligned={"full-width"}
```
GCHQ-Developer-847, line 411: I think these can just be:
```suggestion
          id="table-of-contents-section-container"
          aligned="full-width"
```

packages/canary-web-components/src/components/ic-table-of-contents/ic-table-of-contents.tsx
```diff
@@ -0,0 +1,442 @@
...
+          <div
+            role="navigation"
+            aria-label="Table of contents"
+            class={`${TOC_CONTENT_CLASS} ${
+              position === "top" && "table-of-contents-position-top"
+            }
+            ${position === "right" && "table-of-contents-position-side"}`}
+          >
+            <ic-typography variant="subtitle-large">{heading}</ic-typography>
+            <div class={"table-of-contents-list"}>
```
GCHQ-Developer-847, line 423:
```suggestion
            <div class="table-of-contents-list">
```

packages/canary-web-components/src/components/ic-table-of-contents/ic-table-of-contents.css
```diff
@@ -0,0 +1,139 @@
+/*   --ic-table-of-contents-text: var(--ic-color-text-primary);
```
GCHQ-Developer-847, line 1: Can these commented-out lines be removed?

packages/canary-web-components/src/components/ic-table-of-contents/ic-table-of-contents.tsx
```diff
@@ -0,0 +1,442 @@
...
+              <div
+                class={`table-of-contents-sidebar ${
+                  monochrome && `table-of-contents-sidebar-monochrome`
+                }`}
+                style={{ height: size === "large" ? "2rem" : "1.75rem" }}
```
GCHQ-Developer-847, line 428: Could this styling be moved into the CSS file instead? As we generally don't add inline styling like this in other components

Same with line 412

GCHQ-Developer-847 (changes requested):
Looks amazing 👏 Noticed a few things but they're only very small :)

GCHQ-Developer-530:
I've come across an issue when trying to add table of contents guidance to our site, I thought it was just a component preview issue, but I think it'll be an issue when we come to replace the anchor nav with table of contents too so it might be worth trying to fix. 

Because they are h2s they're showing up in the table of contents, but because we change them to wrapped h2s, the table of contents labels include the div and permalink etc.

Screenshot 2025-08-26 at 14 42 38

Is there a way to make sure it's only the h2 text content that makes the label?

packages/canary-react/src/stories/ic-table-of-contents.stories.js
```diff
@@ -0,0 +1,263 @@
...
+    componentAPI: {
+      data: readme,
+    },
+  },
+};
+
+/**
+ * The table of contents component is used to create a navigable list of sections within a page or document.
+ * 
+ * It can be positioned on the left, right or on top of the child content.
```
GCHQ-Developer-847, line 21: I think we're just putting all canary guidance onto the website now, rather than in Storybook, following [#1557](https://github.com/mi6/ic-design-system/issues/1557). So I guess this comment can be removed?

Also the comment in the web components Storybook. Although happy for the comments to be left  if you think they're still useful - but just wanted to point out that it will probably just be our team who will see them :)

GCHQ-Developer-847 (commented):
Looks really good :)

Just spotted one more thing, however. Also I think there are still a couple of the others issues left - the blue selected indicator doesn't stretch to the height of the table of contents item, and it's not visible in high contrast mode

image

<br />

image

Will be happy to approve after these are resolved

GCHQ-Developer-299:
> Looks really good :)
> 
> Just spotted one more thing, however. Also I think there are still a couple of the other issues left - the blue selected indicator doesn't stretch to the height of the table of contents item, and it's not visible in high contrast mode
> 
> image
> image
> Will be happy to approve after these are resolved

@GCHQ-Developer-847 thanks for the really thorough reviews this week - i've addressed the remaining issues with the sidebar - let me know if "highlight" isn't the correct colour for use in HC :)

GCHQ-Developer-847:
> > Looks really good :)
> > Just spotted one more thing, however. Also I think there are still a couple of the other issues left - the blue selected indicator doesn't stretch to the height of the table of contents item, and it's not visible in high contrast mode
> > image
> > image
> > Will be happy to approve after these are resolved
> 
> @GCHQ-Developer-847 thanks for the really thorough reviews this week - i've addressed the remaining issues with the sidebar - let me know if "highlight" isn't the correct colour for use in HC :)

No problem! :) I would agree that "highlight" is the most appropriate colour to use 👍 

Just noticed a small issue with the height of the selected indicator now though - when you change the `position` prop / change the screen size, the indicator doesn't adjust to match the new height (as the amount of lines the text wraps onto changes). It only updates after a page refresh

https://github.com/user-attachments/assets/773c0909-702f-4125-87d7-f87e4e00de1f

ad9242 (dismissed):

ad9242 (approved):

PR #3885, opened 2025-08-21
fix(web-components): fix accessibility bug where divider was not a list item in ic-side-navigation

### LINKED ISSUE #3836: [ic-divider]: Accessibility failure when used inside a list
### Summary of the bug 
Currently, the ic-side-navigation primary-navigation slot consists of an unordered list, containing ic-navigation-item elements with the listitem role. However, our examples also recommend using ic-dividers in this slot, which is not correct HTML as unordered lists must only contain list item elements, and the ic-divider is a <hr>.

This leads to automated accessibility tests failing.

### 🧐 Expected behaviour 

When an ic-divider is used in the primary-navigation slot of an ic-side-navigation:
1. The unordered list should only contain list item elements 
2. The divider should not contribute to the total number of list items announced by a screen reader (e.g. 6 ic-navigation-items and 1 ic-divider should result in a screen reader announcing that the list contains 6 items, not 7)
3. Ideally, the divider should be announced as a separator

Ensure the solution is tested manually with assistive tech for usability as well as running automated tests to ensure there are no failures.

### Suggested fix

I believe using a list item with a separator role should achieve all of the above criteria but it would need to be tested:
```
<ic-divider slot="primary-navigation">
    <li role="separator"></li>
</ic-divider>
```
instead of the current:
```
<ic-divider slot="primary-navigation">
    <hr></hr>
</ic-divider>
```

I'm not entirely sure how it should be implemented. Possibly a new variant of the divider is needed for when it is used in a list?

There will need to be an additional ticket to implement the changes on the guidance site (to the ic-divider guidance and updating component examples which use it).

If applicable, tell us how urgent it is that this issue gets resolved, based on the impact it has on your team's work or project timeline.

GCHQ-Developer-112:
## Summary of the changes
When `ic-divider` is used in `ic-side-navigation`, it's used in a list. This work is to ensure that `ic-divider` appears as a list item whilst maintaining its separator role. This issue will fix an accessibility failure

## Related issue
#3836

### General 

### Testing

### Accessibility 

### System modes

### DIFF
--- a/packages/web-components/src/components/ic-divider/ic-divider.tsx
+++ b/packages/web-components/src/components/ic-divider/ic-divider.tsx
@@ -33,6 +33,8 @@ import {
   scoped: true,
 })
 export class Divider {
+  private parentElIsSideNav: boolean = false;
+
   @Element() el: HTMLIcDividerElement;
 
   @State() foregroundColor: IcBrandForeground = getBrandForegroundAppearance();
@@ -76,6 +78,17 @@ export class Divider {
   }
 
   private updateMonochromeState(): void {
+    if (this.parentElIsSideNav) {
+      this.el.classList.add("ic-side-navigation-keyline");
+      if (this.foregroundColor === "light") {
+        this.theme = "dark";
+      } else {
+        this.theme = "light";
+      }
+    }
+  }
+
+  componentWillRender(): void {
     const parentEl = this.el.parentElement;
     if (parentEl) {
       const isBottomSideNav = parentEl.classList.contains("bottom-side-nav");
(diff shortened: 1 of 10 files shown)


GCHQ-Developer-718 (dismissed):

GCHQ-Developer-530 (dismissed):

packages/web-components/src/components/ic-divider/ic-divider.tsx
```diff
@@ -178,6 +183,12 @@ export class Divider {
         })}
       >
         {orientation === "horizontal" &&
```
ad9242, line 185:
```suggestion
         {orientation === "horizontal" &&
          (!renderLabel() || !isPropDefined(labelPlacement)) &&
          (this.parentElIsSideNav ? <li role="separator" /> : <hr />)}
```

ad9242 (changes requested):
just a very slight optimisation to the conditional rendering (saves evaluating the `orientation === "horizontal"` twice if it satisfies the first condition)

packages/web-components/src/components/ic-divider/ic-divider.tsx
GCHQ-Developer-112: I've made this change now

Issue #2771, opened 2024-11-13
[ic-button] remove instances of variant="icon"


gd2910:
## Summary 
Ic-button variant `icon` is used in many components, it has now been removed from the design kit and should be replaced with variant `icon-tertiary`.

After this is complete the icon variant should be removed.

GCHQ-Developer-299:
This ticket is to remove our usage of variant=icon across other components, and to deprecate variant=icon in button. 

This is being partially completed as part of Button Dark Mode work - the developer advises that it's not as simple as a find and replace - you might need to use different 'size' variants to match existing UI - and will impact some visual regression tests.

GCHQ-Developer-299:
Will need separate ticket for Guidance site:
- update code example for Button icon variant to remove remaining icon variant
- add alert to button page explaining that icon variant has been removed in v3

PR #3906, opened 2025-08-29
2771 ic button remove instances of varianticon

### LINKED ISSUE #3760: [ic-button] dark mode hover state for icon variant
### Summary of the bug 
The dark mode hover state for icon buttons looks very dark. When you hover over other buttons, the background colour/border colour goes lighter, but for icon buttons the icon itself goes darker.

### 🪜 How to reproduce 
1. Go to [button page](https://mi6.github.io/ic-design-system-githubpages/branches/develop/components/button)
2. Hover over coffee cup button
3. See the error 

### 📸 Screenshots or code 

Image

### 🧐 Expected behaviour 
The icon should lighten when hovered like the other buttons.

### Additional info 
Might need to add Cypress tests for this, it looks like icon variants are the only ones that don't have hover/pressed state tested.

GCHQ-Developer-718:
## Summary of the changes
ic-button variant=icon mapped to variant=icon-tertiary. This is because there are contrast issues with the icon variant hover styling and the icon variant is deprecated.

Also resolves #3760.

Also removes all internal usage of the icon variant.

## Related issue
#2771 
#3760

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes


GCHQ-Developer-299 (changes requested):
Looks like the change to the button variant has introduced a native tooltip on the card interaction button? it doesn't appear on the develop branch.  (seen on the Card Vertical > With Interaction Button story)
Screenshot 2025-09-01 at 08 55 23

packages/web-components/src/components/ic-button/ic-button.tsx
```diff
@@ -601,7 +609,7 @@ export class Button {
 
     const describedby = !hasTooltip
       ? describedById
-      : variant !== "icon" || !ariaLabel
+      : !isIconVariant() || !ariaLabel
```
GCHQ-Developer-530, line 612: I think this might need to be changed back, looks like icon-tertiary variants are no longer getting a describedby and the corresponding tooltip doesn't get an id. `isIconVariant` doesn't look to be just returning variant icon, I think it's any variant that starts with "icon"

GCHQ-Developer-718: I've realised this check is no longer needed as buttons will no longer have variant="icon" so I've removed it.

Issue #2063, opened 2024-06-25
Duplicate playground storybooks in web components


GCHQ-Developer-530:
## Summary 
Currently we have done the work to add playground storybooks for all components in React. These need to be duplicated in web components as well. 

https://github.com/mi6/ic-ui-kit/issues/1970

## 💬 Description 

These are the components with potentially no web components playground - 
 Alert
 Back to top
 Button
 Checkbox
 Footer
 Hero
 Loading indicator
 Page header
 Search bar
 Section container
 Select
 Side navigation
 Stepper
 Switch
 Text field
 Typography

## 💰 Use value 
Some of them have got visual changes on web components vs react so it's worth having these playgrounds for our piece of mind.

PR #3912, opened 2025-09-02
2063 duplicate playground storybooks in web components

jd3267:
## Summary of the changes
Added playground to storybooks for back-to-top, footer, hero, section container and toggle button
web components

## Related issue
#2063

### General 

- [X] Changes to docs package checked and committed.
- [X] All acceptance criteria reviewed and met. 

### Testing

### DIFF
--- a/packages/web-components/src/components/ic-footer/ic-footer.stories.js
+++ b/packages/web-components/src/components/ic-footer/ic-footer.stories.js
@@ -5,6 +5,78 @@ export default {
   component: "ic-footer",
 };
 
+const defaultArgs = {
+  aligned: "left",
+  breakpoint: "medium",
+  caption:
+    "All content is available under the Open Government Licence v3.0, except source code and code examples which are available under the MIT Licence.",
+  copyright: true,
+  description:
+    "The ICDS is maintained by the Design Practice Team. If you've got a question or want to feedback, please get in touch.",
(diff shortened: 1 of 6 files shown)


ad9242:
There seems to be something weird happening with back to top. It is initially visible but then disappears when you scroll to the bottom of the page, which is the opposite of the intended behaviour. Also clicking on it takes you to the bottom of the page rather than the top. All seems ok in the React playground. Hopefully it's just something about the story config, rather than an actual bug in the component

ad9242 (approved):

packages/web-components/src/components/ic-footer/ic-footer.stories.js
```diff
@@ -5,6 +5,78 @@ export default {
...
+};
+
+export const Playground = {
+  render: (args) => html` <ic-footer
+    aligned=${args.aligned}
+    breakpoint=${args.breakpoint}
+    caption=${args.caption}
+    copyright=${args.copyright}
+    description=${args.description}
+    group-links=${args.groupLinks}
```
GCHQ-Developer-299, line 26: I don't think the group-links property works on this storybook. In order for it to properly function the ic-footer-link-group elements below would need to be removed and added in conditionally

GCHQ-Developer-299 (approved):

PR #3915, opened 2025-09-03
3896 aria required not applied on some input components

### LINKED ISSUE #3896: `aria-required` not applied on some input components when they are required fields
### Summary of the bug 
When applying the `required` prop on some input components, the `aria-required` attribute doesn't get programmatically applied.

On these components, the word required can still be heard by a screen reader user (e.g. by appending the word "required" onto the aria-label), but `aria-required` should be programmatically applied as well to ensure it will always get read out, and it will always be clear that the field is required.

### 🪜 How to reproduce 
1. Go the [Required single select example](https://design.sis.gov.uk/components/select/code#required).
2. Inspect the HTML.
3. See that "required" is appended onto the aria-label on the `<button>`, but `aria-required` is not applied
Image

### 🧐 Expected behaviour 
`aria-required` should be applied when an input component is set as required.

I think these are the components / elements which need `aria-required`:
- The `<input>`s in the date input / date picker (could maybe use [`required` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/required) here instead)
- The `<input>`s in the time input (could also maybe use `required`)
- The `<fieldset>` in checkbox groups
- The `<div>` with `role="radiogroup"` on radio button groups
- The `<button>` on a select (single-select)

High urgency - the team this ticket was raised on behalf need it to be completed as soon as possible. Also it's an accessibility issue.

### Additional info 
This was raised on behalf of a customer. This issue was pointed out by a tester their team has been working with.

lz405:
## Summary of the changes
add aria required attribute to `IcRadioGroup`, `IcSelect`, `IcTimeInput` and `IcDateInput`

## Related issue
fix #3896 

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes

### DIFF
--- a/packages/canary-web-components/package-lock.json
+++ b/packages/canary-web-components/package-lock.json
@@ -37,7 +37,7 @@
         "babel-loader": "^8.2.3",
         "github-markdown-css": "^5.5.0",
         "jest": "^26.6.3",
-        "jest-axe": "^5.0.1",
+        "jest-axe": "^8.0.0",
         "jest-cli": "^26.6.3",
         "kill-port-process": "^3.2.1",
         "lit": "^2.0.2",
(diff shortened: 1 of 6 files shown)


GCHQ-Developer-299 (changes requested):
I've realised why the old code did `aria-required={`${required}`}` - sorry for suggesting that you change it - there's some weird behaviour that means without doing that only `aria-required` gets rendered in the HTML and we need `aria-required="true"` which i believe the old solution achieves. 

You can test this with the Radio button Playground storybook, toggling 'required' on and off.

GCHQ-Developer-299 (dismissed):

GCHQ-Developer-530 (commented):
If the snapshots have changed for the core package, I think they'll have changed for the canary one too, so please could you update the test snapshots for date input and time input?

PR #3965, opened 2025-09-23
fix(web-components): fix dismissible chips leaving gap behind

### LINKED ISSUE #3549: Dismissible chip leaves gap to the left of other content when dismissed in a row
## Summary of the bug 
When several dismissible chips are displayed in a row, dismissing the first causes a gap to appear on the left of the remaining chips, forcing it out of line with other content on the page. The issue is amplified the more chips that are dismissed.

## 🪜 How to reproduce 
1. Go to the dismissible story in [storybook](https://mi6.github.io/ic-ui-kit/branches/develop/web-components/?path=/story/web-components-chip--dismissible).
2. Dismiss some of the light chips on the first row.
3. See the error 

## 📸 Screenshots or code 
Image
Image

## 🧐 Expected behaviour 
The chips should remain inline with other content on the page, to not disrupt the reading flow and alignment of web pages.

## 📝 Acceptance Criteria  

*Given* that a dismissible chip exists on a page
*When* it renders 
*Then* it should be inline with the other page content

*Given* that multiple dismissible chips exist on a page
*When* the first is dismissed
*Then* the second should be inline with the other page content

## Additional info 
Potentially caused by the styling applied to the host element, since when it is dismissed the content within the shadow dom is gone but the host persists on the page.

Image

This may also explain the slight overlapping on the left shown in the first image.

lz405:
## Summary of the changes
fix the dismissible chips leaving gap behind after being dismissed

## Related issue
fix #3549

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes

### DIFF
--- a/packages/web-components/src/components/ic-chip/ic-chip.tsx
+++ b/packages/web-components/src/components/ic-chip/ic-chip.tsx
@@ -190,6 +190,7 @@ export class Chip {
     return (
       <Host
         class={{
+          "ic-chip-dismissed": !visible,
           [`ic-theme-${theme}`]: theme !== "inherit",
           [`ic-chip-${foregroundColor}-text`]:
             variant === "filled" && !!foregroundColor,
--- a/packages/web-components/src/components/ic-chip/ic-chip.css
+++ b/packages/web-components/src/components/ic-chip/ic-chip.css
@@ -6,6 +6,10 @@
   vertical-align: middle;
 }
 
+:host(.ic-chip-dismissed) {
+  display: none;
+}
+
 .chip {
   --chip-bg-color: transparent;
   --chip-hover-bg: var(--ic-chip-action-hover);


GCHQ-Developer-299 (dismissed):
Compared the dismissible web-components storybook against [the develop branch](https://mi6.github.io/ic-ui-kit/branches/develop/web-components/?path=/story/web-components-chip--dismissible) and looks like you've sorted it :)

packages/web-components/src/components/ic-chip/ic-chip.tsx
```diff
@@ -189,6 +189,7 @@ export class Chip {
 
     return (
       <Host
+        style={{ display: visible ? "" : "none" }}
```
GCHQ-Developer-530, line 192: I don't think we tend to add inline styles to the host of a component, would you be able to add a class for visible/not visible instead and set `display: none` in the css file?

ad9242: Following on from @GCHQ-Developer-530's comment: Doesn't setting this inline style effectively override the current  styling in the ic-chip.css on the :host element which is set to inline-block? Possibly that could be removed as part of the class based solution suggested?

packages/web-components/src/components/ic-chip/ic-chip.tsx
```diff
@@ -190,6 +190,7 @@ export class Chip {
     return (
       <Host
         class={{
+          dismissed: !visible,
```
ad9242, line 193:
```suggestion
          'ic-chip-dismissed': !visible,
```
Please can you make the class name more unique\component specific as there have been issues in the past with classNames like "icon", "disabled" etc on the host element where styles leak into the components eg if someone had a ".dismissed" selector in their app css

ad9242 (changes requested):
minor change - sorry!

GCHQ-Developer-299 (approved):

Issue #3939, opened 2025-09-11
[ic-pagination-bar]: Setting selectedItemsPerPage while dynamically setting totalRecords sets the ItemsPerPage dropdown incorrectly


jd239:
### Summary of the bug 
An issue occurs where the ItemsPerPage dropdown shows the wrong value when selectedItemsPerPage is set to a fixed value and the totalRecords has dynamically changed.

### 🪜 How to reproduce 
[Stackblitz Example](https://stackblitz.com/edit/rjosoa3e-q4sgqgr6?file=src%2Fapp.tsx)

1. On load, the ItemsPerPage dropdown will select "All". This is due to the totalRecords returning 0. The selectedItemsPerPage is set to 100.
2. A setTimeout within the useEffect will update the totalRecords (to 120) to simulate an API response returning data.
3. The new totalRecords is passed into the pagination bar and the dropdown selects 10

### 📸 Screenshots or code 
I think the issue is within the `setPaginationBarContent` method.

On [L471](https://github.com/mi6/ic-ui-kit/blob/main/packages/canary-web-components/src/components/ic-pagination-bar/ic-pagination-bar.tsx#L471), the static value of 0 is used to loop through the itemsPerPage options.  This will always set the dropdown to the lowest value and ignore the selectedItemsPerPage even if it is manually set. 

This will also trigger the onItemsPerChange event which may be used to trigger API requests in order to reduce the number of items returned from an endpoint. This means that once the initial API request is made and the totalRecords is updated based on the API response, a second "redundant" API request will be made.

### 🖥 📱 Device 
- Type: Desktop
- Device: Windows
- OS version: Windows
- Browser version: Chrome 139

### 🧐 Expected behaviour 
Since the selectedItemsPerPage is set to 100, totalRecords are now 120, and the itemsPerPage includes 100, I would expect 100 to appear to be the selectedValue.

### 📝 Acceptance Criteria  
If relevant, describe in full detail the different interactions and edge cases that the component or patterns needs to fulfil. 

*Given that the initial totalRecords is 0 due to an API request being made, and a selectedItemsPerPage value is set* 
*When the API returns data and the totalRecords is updated,*  
*Then the ItemsPerPage dropdown should select the manually set selectedItemsPerPage value when available* 

I'm not 100% sure this should be done by the component or if the engineer should use a flag to prevent the onItemsPerChange callback from being triggered but will add it anyway as food for thought:
[...]

PR #3970, opened 2025-09-23
[ic-pagination-bar]: itemsPerPage incorrect value on totalItems update fix

jd239:
## Summary of the changes
- Added logic to correctly set itemsPerPage when totalItems is updated
- Added `isUserAction` to IcItemsPerPageChange due to event trigger when totalItems updates itermPerPage selection

## Related issue
#3939 

### General 

### Testing

### Accessibility 

N/A

### Resize/zoom behaviour 

N/A

### System modes

### Testing content extremes

N/A

### DIFF
--- a/packages/canary-web-components/src/components/ic-pagination-bar/ic-pagination-bar.stories.js
+++ b/packages/canary-web-components/src/components/ic-pagination-bar/ic-pagination-bar.stories.js
@@ -45,7 +45,7 @@ export const ItemsPerPageControl = {
       ></ic-pagination-bar>
     </div>
     <script>
-      const paginationBar = document.querySelector("ic-pagination-bar");
+      var paginationBar = document.querySelector("ic-pagination-bar");
       paginationBar.itemsPerPageOptions = [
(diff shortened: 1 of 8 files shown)


packages/canary-web-components/src/components/ic-pagination-bar/ic-pagination-bar.stories.js
```diff
@@ -45,7 +45,7 @@ export const ItemsPerPageControl = {
       ></ic-pagination-bar>
     </div>
     <script>
-      const paginationBar = document.querySelector("ic-pagination-bar");
+      var paginationBar = document.querySelector("ic-pagination-bar");
```
GCHQ-Developer-299, line 48: any reason you're not using 'let' here if paginationBar needs to be mutable?

GCHQ-Developer-299: From a google, var has a different scope to let, so that kinda makes sense. Thanks for explaining!

Issue #3671, opened 2025-06-19
[ic-time-picker]: Time selector component


GCHQ-Developer-112:
### Summary 
Develop a time selector component that will act as an atom within `IcTimePicker`.

Match to existing component definition:

Text entry
Date input
Required
Disabled
Interaction states
Large, default and small size variants.
Validation
scrollable and selectable time options
Limits

### 💬 Description 
This issue addresses building one of the atomic components that make up time picker and date time picker.

### 💰 User value
Helps developers make clearer, more specific inputs for forms. Helps end users by simplifying interactions when inputting a time rather than using a standard text box/input

### 📚 User Stories 

As a time picker user 
I need to be able to both freely type a time in text format (HH:MM:SS) AND select from a dropdown calendar view
AND have any change reflected in both the bar and the selector
SO that i have flexibility of input, as well as confirmation of user input

As a time picker user
I need to feely type a time in text format (HH:MM:SS)
So that i can input a date easily, without need to interact with a modal/dialogue window or dropdown

As a time picker user
I need to have a ‘Clear’ button
So that I can reset my input

As a designer/developer creating an app that uses time pickers
I need to be able to set specific disabled times
So that I can force users to pick from specific times, as well as clearly convey what times are enabled

As a designer/developer creating an app that uses time pickers
I need to be able to pre-select a default time
So that I can convey a message to my users (i.e. suggested time, current time, popular time...)

As a time picker user
I need to view a focus indicator for any element that can be interacted with,
So that I know where my keyboard actions will target.

As a time picker user
I need to be able to navigate through the time picker dropdown using a keyboard
So that I have equal access to its content.

### 📝 Acceptance Criteria  

**Variants**

Time Input: An input field, very similar to date input, whereby a user can input a specific time.

Time Picker: Comprised of time input and a dropdown time picker to allow visual aid and mouse selection of time.

Timezone select: A select component used for choosing timezone that sits next to the time input/picker.  

-----------------------
**Properties**

- Display seconds: Toggles the visibility of seconds in the helper text, placeholder text, time picker and input.

- Display milliseconds: Toggles the visibility of milliseconds in the helper text, placeholder text, time picker and input. 

- 12hr/24hr clock: Toggles between a 12 hour and 24 hour clock selection and validation in picker and input. 

- Default values: Define the default values that autofill on a partial time selection. 

- Define placeholder: Allows selection of a specific time, now or 00:00:00 as the placeholder time displayed.

---------------------------
[...]

PR #4011, opened 2025-10-07
3671 time selector component

GCHQ-Developer-530:
## Summary of the changes
Implement time selector component.

NOTE:
The column focus indicator doesn't quite match the Figma designs, this is because matching the designs meant that items weren't clickable when the column was focused. We've decided this different style is okay for now as it's canary, and we can look at it further down the line if it's an issue.

## Related issue
#3671 

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes


packages/canary-web-components/src/components/ic-time-selector/ic-time-selector.stories.js
```diff
@@ -0,0 +1,185 @@
...
+  render: () => html`<ic-time-selector></ic-time-selector>`,
+  name: "Default",
+};
+
+/**
+ * Retrieving the time selector value via `icChange` returns the time as a Date object.
+ *
+ * The event returns the Date object once hour, minute and second have been selected by clicking the tick icon button.
+ */
+export const IcChangeEvent = {
```
GCHQ-Developer-847: Clearing the selected value using the clear button doesn't cause an icChange event to be emitted - I feel like it should? 

The date picker emits an icChange event when you click the "Clear" on the dropdown

packages/canary-web-components/src/components/ic-time-selector/ic-time-selector.tsx
```diff
@@ -0,0 +1,799 @@
...
+  @Element() el: HTMLIcTimeSelectorElement;
+
+  @State() selectedHour?: number;
+  @State() selectedMinute?: number;
+  @State() selectedSecond?: number;
+  @State() selectedPeriod?: IcTimePeriods;
+  @State() ariaLiveMessage: string = "";
+
+  /**
+   * An array of times that will be disabled in the time input. The times can be in any format supported as `timeFormat`, in ISO 8601 time string format (`HH:MM:SS`) or as a JavaScript `Date` object.
```
GCHQ-Developer-847:
```suggestion
   * An array of times that will be disabled in the time selector. The times can be in any format supported as `timeFormat`, in ISO 8601 time string format (`HH:MM:SS`) or as a JavaScript `Date` object.
```

packages/canary-web-components/src/components/ic-time-selector/ic-time-selector.stories.js
ad9242: should it also return focus to the HH selector? as currently the focus gets lost completely (because the clear button becomes disabled)

ad9242 (changes requested):

ad9242:
i have a gripe with the am\pm selector - probably one to go back to design. I don't like the way the items move around as you select one. i feel it would be better if it used inline toggle buttons. something like this:

image

packages/canary-web-components/src/components/ic-time-selector/ic-time-selector.tsx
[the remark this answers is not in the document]
GCHQ-Developer-530: I've updated the z-index and that seems to have helped but the highlight lines still appear above the focus (nothing else seems to anymore though), not sure what else to try to get them to appear below as they already have a lower z-index

packages/canary-react/src/stories/ic-time-selector.stories.js
```diff
@@ -0,0 +1,317 @@
...
+  timeFormat: "HH:MM:SS",
+  timePeriod: "24",
+  value: "12:00:00",
+};
+
+/**
+ * Go to the <ic-link href="/?path=/story/react-components-time-selector--playground">separate page for the playground example</ic-link> to view the prop controls.
+ */
+
+export const Playground = {
```
GCHQ-Developer-847: When I set `minTime` in the playground to 09:00:00 and then refresh, the time selector moves to 09:00:00 but this means it overrides the value of the `value` prop:
image
I think the selected time / `value` should always take precedence i.e. it should still display 12:00:00 on load

packages/canary-web-components/src/components/ic-time-selector/ic-time-selector.css
GCHQ-Developer-847: Another thing, sorry - if I set `prefers-reduced-motion` to "reduce", the sliding animation on each column still happens so I think this will need sorting

GCHQ-Developer-847 (dismissed):
Looks great to me now 👏

GCHQ-Developer-847 (dismissed):

GCHQ-Developer-299 (approved):

GCHQ-Developer-299 (approved):

PR #4036, opened 2025-10-21
3888 get cursor position in text field keydown events

lz405:
## Related issue
feat #3888

### General 

### Testing

### System modes

### Testing content extremes


ad9242:
it doesn't seem to work too well when using the arrow keys: 
eg typed "hello" and then pressed arrowLeft, you can see it incremented the position by 1, even though the cursor had moved back:

image

similar if the cursor was at the end and then you press the "Home" key - the cursor is now at the start but reports as the end:

image





I wonder if this also needs exposing via another (additional) event (maybe focus?), so that it would be more useful to users. Eg as you click around the MUI input in this example the cursor position is updated\reported
https://codesandbox.io/p/sandbox/textfield-selectionstart-r4xjl?file=%2Fsrc%2Findex.js

Issue #2029, opened 2024-06-19
Loading indicator should always be the circular variant when at `icon` size


GCHQ-Developer-847:
## Summary of the bug 
When the `size` prop on the loading indicator is set to `icon`, it should always show the circular variant and never the linear variant (this is the case in the Figma designs).

## 🪜 How to reproduce 
1. Go to the story for the [icon loading indicator](https://mi6.github.io/ic-ui-kit/branches/develop/web-components/?path=/story/web-components-loading--icon) and change the `type` in the DevTools to `linear`.
2. See that it tries to render a very small linear loading indicator e.g.:
<img src="https://github.com/mi6/ic-ui-kit/assets/111882109/c0469132-46c1-482e-9019-7b0ab4ccf0a7" width="70">

## 🧐 Expected behaviour 
As defined on Figma, the loading indicator should always display the circular variant when it is set to its icon size.

## Additional info 
It is probably worth adding a note to the description for the `size` prop to say that setting it to `icon` will always display the circular indicator.

MI6-255:
Design thinks likely best to keep consistent with Button loading, so Figma needs updating

mi6-577:
Duplicate issue to #1476

PR #4135, opened 2025-12-17
2029 loading indicator should always be the circular variant when at icon size

GCHQ-Developer-299:
## Summary of the changes

Added new internalType value inside loading-indicator to allow for enforcing 'internaltype' of 'circular' when type is 'linear' and size is 'icon'.

You can test this on the playground stories by setting the size to 'icon' and seeing that the type changes to 'circular'.

Unsure if I need to add a test for this, could potentially update one of the Cypress VR tests for icon variants. Thoughts?

## Related issue
#2029

### DIFF
--- a/packages/docs/docs.json
+++ b/packages/docs/docs.json
@@ -10166,7 +10166,7 @@
           "mutable": false,
           "attr": "type",
           "reflectToAttr": true,
-          "docs": "The type of indicator, either linear or circular.",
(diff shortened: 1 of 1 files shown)


GCHQ-Developer-530 (approved):

packages/docs/docs.json
GCHQ-Developer-847: Might need to do an npm run build as the `--circular-diameter` CSS property info has been removed from this file

GCHQ-Developer-530 (approved):

Issue #3871, opened 2025-08-15
[ic-button] - add an icon button playground to storybooks


ad9242:
### Summary 
The IcButton component has storybook playgrounds:
https://mi6.github.io/ic-ui-kit/branches/develop/react/?path=/story/react-components-button--playground)
https://mi6.github.io/ic-ui-kit/branches/develop/web-components/?path=/story/web-components-button--playground
but there is no easy way to test\play around with the icon variants of ic-button in a similar way

### 💰 Use value 
Would allow developers to quickly see how props can affect the icon-variants of ic-button

### 🚨 Urgency (low, medium or high)
Low

PR #4147, opened 2025-12-30
3871 ic button add an icon button playground to storybooks

GCHQ-Developer-530:
## Summary of the changes
Update button stories to have icon only playground.
Full width currently doesn't work, I think this is an issue with the component - will open a ticket once I've spoken to a designer.

## Related issue
#3871 

### Testing

### Testing content extremes

### DIFF
--- a/packages/react/src/stories/ic-button.stories.js
+++ b/packages/react/src/stories/ic-button.stories.js
@@ -1665,6 +1665,86 @@ export const PlaygroundWithIcon = {
   name: "Playground with icon",
 };
 
+export const PlaygroundIconOnly = {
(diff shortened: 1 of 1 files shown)


.npmrc
```diff
@@ -1,2 +1 @@
-engine-strict=true
-ignore_scripts=true
\ No newline at end of file
+engine-strict=true
```
evenstensberg: @GCHQ-Developer-530 could you open an issue explaining what needs to be run postinstall and I'll create a self-owned hook for it? ignore-scripts is good w.r.t security and readability

Issue #4155, opened 2026-01-05
[ic-time-selector] Remove seconds from scrollable list


GCHQ-Developer-530:
### Summary 
For a better UX and an easier experience developing/designing future date time components, we want to limit the time selector to just hours and minutes.

### 💬 Description 
We still allow seconds and milliseconds in time input but in time selector we want to remove seconds and be left with just hours, minutes and AM/PM columns.

Something to consider: we may still need events to potentially emit seconds and milliseconds for when time input and selector are combined to make time picker. However, this might be something to think about when time picker is developed.

### 💰 Use value 
Makes the UX better for all date/time components.

### 📝 Acceptance Criteria  
*Given* there is a time selector or time picker
*When* a user opens up the scrollable selection area
*Then* there should only be hours, minutes and AM/PM.

### 🚨 Urgency (low, medium or high)
Medium - it doesn't necessarily block the development of time picker but it might make it easier if this is done beforehand.

PR #4179, opened 2026-01-13
4155 ic time selector remove seconds from scrollable list 3

jd3267:
## Summary of the changes
Removed the seconds column from ic-time-selector so that now only hours and minutes (with optional am/pm) column are shown.

## Related issue
#4155 

### General 

### Testing

### System modes

### Testing content extremes

### DIFF
--- a/packages/canary-docs/docs.json
+++ b/packages/canary-docs/docs.json
@@ -6540,41 +6540,6 @@
           "setter": false
         },
         {
-          "name": "timeFormat",
-          "type": "\"HH:MM\" | \"HH:MM:SS\"",
-          "complexType": {
-            "original": "IcTimeFormatSelector",
-            "resolved": "\"HH:MM\" | \"HH:MM:SS\"",
-            "references": {
-              "IcTimeFormatSelector": {
-                "location": "import",
-                "path": "../../utils/types",
-                "id": "src/utils/types.ts::IcTimeFormatSelector"
-              }
-            }
-          },
-          "mutable": false,
-          "attr": "time-format",
-          "reflectToAttr": false,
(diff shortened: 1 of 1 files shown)


packages/canary-react/src/stories/ic-time-selector.stories.js
```diff
@@ -143,7 +143,7 @@ export const DefaultValueHHMM = {
 export const MinTime = {
```
GCHQ-Developer-530: Same here - the time format stories can be removed

GCHQ-Developer-530 (commented):
Looks good - just noticed some small things that can be removed now

packages/canary-web-components/src/components/ic-time-selector/ic-time-selector.tsx
```diff
@@ -81,11 +77,6 @@ export class TimeSelector {
    */
   @Prop() theme?: IcThemeMode = "inherit";
 
-  /**
-   * The format in which the time will be displayed.
-   */
-  @Prop() timeFormat: IcTimeFormatSelector = this.DEFAULT_TIME_FORMAT;
```
GCHQ-Developer-530: Now that this is removed, I think the [IcTimeFormatSelector type](https://github.com/mi6/ic-ui-kit/blob/main/packages/canary-web-components/src/utils/types.ts#L262) also needs to be removed as it only has the one option of HH:MM. If it's used in loads of places, we could always just change the type to just allow HH:MM, but feels like it could be a private variable instead of a specified type at that point.

GCHQ-Developer-530 (approved):

Issue #3588, opened 2025-05-28
[SUGGESTION]: upgrade storybook in all packages


evenstensberg:
### Summary
Ic-ui-kit should upgrade storybook so that in the future there is less work on updating dependencies. 

- [ ] Upgrade Storybook in all packages
- [ ] Remove audit exceptions that are resolved by this work

### 💬 Description 
When upgrading, it makes it easier to fix security vulnerabilities also. It allows the build to be improved such as fixing `storybook/addon-postcss` in the future

### 🚨 Urgency (low, medium or high)
low

### Why do we need it? 

DX + maintenance

evenstensberg:
Over time this change will resolve [this CVE](https://github.com/mi6/ic-ui-kit/blob/develop/audit-ci.json#L18)

PR #4184, opened 2026-01-14
3588 upgrade storybook to v10

GCHQ-Developer-299:
## Summary of the changes
Putting up as a draft initially, i've been very cautious with commits too.

I've been using the `npx storybook@[VERSION] upgrade` command. Very helpful.

Steps required to get web-components to work with storybook 10:
- Update postcss (and eventually remove postcss plugin for storybook) which will eventually mean we can remove a CVE :)
- Update WC-Storybook to use vite instead of webpack. Removed webpack from package.
- Upgrade storybook to v9, it did a number of refactors to stories.
- Convert storybook/main.js to use ESM instead of CommonJS syntax
- Upgrade storybook to latest (10.1.11)
- Removed eslint-plugin-storybook (added by the upgrade) as it was going to require us to move our eslintrc.js -> eslint.config.js and change the syntax. Maybe that change is due one day anyways, we'll see

I also had to update our eslint version (related storybook moving to ESM?) from v8.x to v9.x which involved:
- replacing .eslintrc.js with eslint.config.json. This was partly done via `npx @eslint/migrate-config .eslintrc.json` but required a lot of manual work
- updating the plugins (typescript, sonar, etc.) 
- adding a couple of rule exceptions, as new failures were appearing (i'm assuming old versions of the plugins didn't have these rules?). TODO: create a ticket to address those failures?
- commenting out a couple of rules that eslint claimed didn't exist. Might be something I can go back and readdress

For react and canary-react:
- i've had to remove [storybook-addon-performance](https://www.npmjs.com/package/storybook-addon-performance) because it was last updated two years ago and incompatible with v9
- Replacing webpack with vite meant that storybook noticed we had JSX inside .js files, so i've had to update preview.js and all the .stories.js files to jsx. It's purely a name change.
- I removed not just webpack from react's package.json but the cli and dev-server which we don't use. I checked and cypress still runs, it installs webpack as a dependency. 

Nextjs: 
- Nextjs didn't run for me before the upgrade. Now it does run, but none of the styles work. Still a win? 

Followon tickets?
- Find new performance plugin
- Fix nextjs properly - or do I do that in this PR?

## Related issue
#3588

### DIFF
--- a/packages/nextjs/.storybook/main.mjs
+++ b/packages/nextjs/.storybook/main.mjs
@@ -1,5 +1,5 @@
-/** @type { import('@storybook/nextjs').StorybookConfig } */
-const config = {
+/** @type { import('@storybook/nextjs-vite').StorybookConfig } */
+export default {
   core: {
     disableTelemetry: true,
   },
@@ -19,14 +19,12 @@ const config = {
 
   addons: [
     "@storybook/addon-links",
-    "@storybook/addon-essentials",
-    "@storybook/addon-interactions",
-    "@storybook/addon-postcss",
     "@storybook/addon-a11y",
+    "@storybook/addon-docs"
   ],
 
   framework: {
-    name: "@storybook/nextjs",
+    name: "@storybook/nextjs-vite",
     options: {},
   },
 
@@ -35,5 +33,8 @@ const config = {
   typescript: {
     reactDocgen: "react-docgen-typescript",
   },
+  async viteFinal(config) {
+    config.assetsInclude = ['**/*.md']
+    return config;
+  },
 };
-export default config;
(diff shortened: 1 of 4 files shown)


packages/canary-web-components/package.json
```diff
@@ -78,7 +75,7 @@
     "react-markdown": "^8.0.7",
     "remark-gfm": "^3.0.1",
     "stencil-inline-svg": "^1.1.0",
-    "storybook": "^8.6.8",
+    "storybook": "10.1.11",
     "ts-jest": "^26.5.6",
     "webpack": "^5.76.0"
```
evenstensberg: if vite is now you can remove this dep

lz405:
image

Breaks when running 'start-storybook' on Ubuntu

The linked issue https://github.com/npm/cli/issues/4828 seems to be a dead link

GCHQ-Developer-530 (approved):
Pulled down locally and all seems to work!

PR #4185, opened 2026-01-16
3891 icon buttons should be able to use circular loading indicators

GCHQ-Developer-530:
## Summary of the changes
Add circular-loading-indicator prop to button so that when an icon button is used then there's the option to have linear or circular loading

## Related issue
#3891 

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes


GCHQ-Developer-299:
> > This work seems to clash with the recent change [to enforce circular loading indicators when at icon size](https://github.com/mi6/ic-ui-kit/pull/4135) - what's the thinking here?
> 
> This should be the same, just for inside icon buttons. I added a prop as the ticket asked for one, rather than enforcing it. But I'd be happy to change it to always be circular for icon buttons too?

Could you change it to enforce circular loading please? In figma [icon buttons](https://www.figma.com/design/e2hv9eRJtQ4GBXPD7vhMky/branch/QVPc4rHb0CIeNniIXE7HTt/ICDS-V3.0) only have circular loading indicators

GCHQ-Developer-847 (changes requested):
I think there's an issue with the small variant; the loading indicator does not stay exactly in the centre - see this video (I change it to the large variant for comparison):

https://github.com/user-attachments/assets/8b74d6d4-3589-4dff-8385-4f91eb5dac10

PR #4244, opened 2026-02-05
Additional skip in scope check

GCHQ-Developer-530:
## Summary of the changes
Added an additional skip in scope check to ignore the release commit entirely - it should now pass the develop -> main release

## Related issue
N/A


GCHQ-Developer-299 (approved):
Worth a shot!

PR #4372, opened 2026-03-12
Fix console error in toast component

ad9242:
## Summary of the changes
An error was appearing in the console when a toast was displayed: 
image

Added a check to ensure the method is only called if an element exists

## Related issue
N\A

### DIFF
--- a/packages/web-components/src/components/ic-toast/ic-toast.tsx
+++ b/packages/web-components/src/components/ic-toast/ic-toast.tsx
@@ -259,7 +259,8 @@ export class Toast {
       );
       return null;
     } else {
-      window.setTimeout(() => this.interactiveElements[0].setFocus(), 200);
+      this.interactiveElements[0] &&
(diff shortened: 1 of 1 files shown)


GCHQ-Developer-299:
@ad9242 is there a way to replicate the error message for testing locally?

ad9242:
> @ad9242 is there a way to replicate the error message for testing locally?

I tried to replicate in a stackblitz, but was unable to. We encountered an issue with toasts not displaying using the documented way of setting the toast region openToast prop (we were also unable unable to recreate this outside of our app).

Instead we are calling the setVisible() method directly (whilst appreciating this is an internal\undocumented method). 

Whilst it is an unusual approach, I felt that adding a check for an undefined variable was a worthwhile and minor change.

GCHQ-Developer-299 (approved):
I was unable to replicate the error in storybook locally, but this change didn't break anything and sounds worth making

GCHQ-Developer-112 (approved):

Issue #4204, opened 2026-01-22
Replace cypress-audit with lighthouse


GCHQ-Developer-530:
### Summary
We're using cypress-audit and @cypress-audit/lighthouse for our performance testing. Both of which we're on the latest version of because they haven't been updated in 3 years. However, these packages are responsible for 6 of the vulnerabilities in the allowlist. 

### 💬 Description 
We should try to find another solution as these packages don't look like they'll get updated. 
I think the best solution is to just use lighthouse directly instead of a cypress lighthouse plugin. That way we know it'll get maintained.

### 🚨 Urgency (low, medium or high)
Medium - it would be good to reduce the audit allowlist.

### Why do we need it? 
We'd have better performance tests (appreciate these aren't a priority) and we'd have less vulnerabilities in the codebase.

GCHQ-Developer-299:
Suggestion: this ticket should be a spike investigating potential packages. Create separate ticket for implementation

MI6-286:
Suggest whoever picks this up looks at the branch [implement-performance-tests-with-lhci](https://github.com/mi6/ic-ui-kit/compare/develop...implement-performance-tests-with-lhci). 

Work completed on this branch:
-	Created new pages within the performance testing app for the remaining components (under packages/react/src/performance-tests/pages/)
-	Completed LHCI setup and created script for running tests using LHCI (packages/react/scripts/lhci-run-per-route.js). Note this was previously working fine with tests running, assertions passing, and reports being output, but now seeing an EPERM issue (https://github.com/googlechrome/chrome-launcher/issues/355)
-	Created GitHub Actions workflow file (with manual trigger)

NOTE: To run tests locally, run `npm run test:performance-lhci` from packages/react

To do if going forward with this:
-	Resolve issue mentioned above
-	Test GitHub Actions workflow on forked repo (parallel runs with batching has been tested and works as expected, but need to ensure tests run successfully and pass on CI) 
-	Once manual GitHub action is working, implement scheduled workflow 
-	Remove all that is required for cypress performance tests

GCHQ-Developer-530:
Blocked by #4299 at the moment

PR #4444, opened 2026-04-24
Implement performance tests with lhci

GCHQ-Developer-530:
## Summary of the changes
Replace cypress audit performance tests with lighthouse. Update commands and add additional component performance tests. Remove np dependency from react, canary-react and nextjs packages. Update workflows, audit allowlist and scripts.

Cherry picked two commits from node upgrade in workflow PR.

## Related issue
#4204 

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes


packages/react/scripts/lhci-run-per-route.js
```diff
@@ -0,0 +1,255 @@
...
+      upload: {
+        target: "filesystem",
+        outputDir: `./lhci-reports/${route.name}`
+      }
+    }
+  };
+}
+
+async function runLhciForRoute(route) {
+  console.log(`\n${"=".repeat(60)}`);
```
evenstensberg: `console.info`

packages/react/scripts/lhci-run-per-route.js
```diff
@@ -0,0 +1,255 @@
...
+    return { route: route.name, status: "passed" };
+  } catch (error) {
+    console.error(`❌ ${route.name} failed\n`);
+    return { route: route.name, status: "failed", error: error.message };
+  } finally {
+    // Clean up config file
+    try {
+      fs.unlinkSync(configPath);
+    } catch (e) {
+      // ignore cleanup errors
```
evenstensberg, line 203: log is useful here, on windows OS's in specific the symlinking through node is somewhat flaky.

packages/react/scripts/lhci-run-per-route.js
```diff
@@ -0,0 +1,255 @@
...
+  for (const route of batchRoutes) {
+    const result = await runLhciForRoute(route);
+    results.push(result);
+    if (result.status === "failed") {
+      failedCount++;
+    }
+  }
+
+  // Summary
+  console.log(`\n${"=".repeat(60)}`);
```
evenstensberg: use `console.info` and also maybe use `process.stdout.columns` / `process.stdout.width` to repeat with the best number of `-`

packages/react/package.json
```diff
@@ -94,7 +93,8 @@
     "style-loader": "^4.0.0",
     "ts-loader": "^9.5.0",
     "typescript": "^4.9.5",
-    "wait-on": "^8.0.1"
+    "wait-on": "^8.0.1",
+    "webpack-cli": "^6.0.1"
```
evenstensberg: isnt webpack removed in favour of vite?

audit-ci.json
GCHQ-Developer-299, line 1: this is wonderful to see!

GCHQ-Developer-299 (dismissed):

GCHQ-Developer-299 (approved):

.github/workflows/run-lhci-performance-tests-manual.yml
```diff
@@ -26,10 +26,13 @@ jobs:
...
   ic-ui-kit-performance-tests:
     needs: [check-updates]
-    name: "Performance tests"
+    name: "Performance tests - Batch ${{ matrix.batch }}"
     runs-on: ubuntu-latest
     permissions:
       contents: write
+    strategy:
+      matrix:
+        batch: [1, 2, 3, 4]
```
GCHQ-Developer-112, line 35: I had a look into GitHub workflow syntax. I couldn't find any references to `batch`. Based on your implementation, `version` might be the right keyword

[GitHub Matrix Strategies](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations#about-matrix-strategies)

packages/react/scripts/lhci-run-per-route.js
```diff
@@ -0,0 +1,255 @@
...
+const COMPONENT_ACCESSIBILITY_OVERRIDES = {
+  "select-variants": 0.95,
+  "searchable-select-variants": 0.95,
+  "multi-select-variants": 0.95,
+  "searchbar-variants": 0.95,
+  "radio-variants": 0.95,
+  "stepper": 0.95,
+  "text-field-variants": 0.94,
+  "chip-multiple": 0.94
+};
```
GCHQ-Developer-112, line 37: Just for my understanding: why are overrides needed in this case?

.github/workflows/run-lhci-performance-tests-manual.yml
```diff
@@ -47,14 +50,16 @@ jobs:
...
 
-      - name: Cypress performance tests
+      - name: LHCI performance tests - Batch ${{ matrix.batch }}
         if: ${{ needs.check-updates.outputs.core == 'true' }}
-        id: core_cypress_performance_tests
-        run: npm run test-cypress-perf:ci
+        id: lhci_performance_tests
+        run: |
+            npm run build
+            node packages/react/scripts/lhci-run-per-route.js --batch ${{ matrix.batch }} --total-batches 4
```
GCHQ-Developer-112, line 58: Is this something you can add as a package.json script so we can run this more easily?

GCHQ-Developer-112 (changes requested):

.github/workflows/run-lhci-performance-tests-manual.yml
GCHQ-Developer-530: Its not a GitHub workflow keyword in this case, it's just a variable that we're passing through to the lhci-run-per-route.js file further down so that it can split the run into 4 batches to make it run quicker.

packages/react/scripts/lhci-run-per-route.js
GCHQ-Developer-530: These were just components that appeared to be a bit flaky so we've added a bit of leniency, just so we know when there's been an actual bad performance.

.github/workflows/run-lhci-performance-tests-manual.yml
GCHQ-Developer-530: Locally we can run `npm run test-lhci-perf:ci`, but because the GitHub workflow needs to pass through the batch number, it has to be run like this.

.github/workflows/run-lhci-performance-tests-manual.yml
GCHQ-Developer-530: Screenshot 2026-05-08 at 07 59 25

This is an example of what running it locally is like, if I passed through 1 instead of the batch coming from the matrix.

GCHQ-Developer-112: Ah ok

GCHQ-Developer-112 (approved):
Happy to approve! Thank you for explaining :)

Issue #4299, opened 2026-03-02
Upgrade node in workflows


GCHQ-Developer-530:
### Summary 
Currently every workflow uses node v20, except one step that uses v22 in order to OIDC working when publishing. That was updated in https://github.com/mi6/ic-ui-kit/pull/4295. 

### 💬 Description 
I initially mass updated the workflows to use v22, but the static analysis step for the branches workflow failed due to the text area spec tests timing out. Everything else seemed to work on v22 except that one test suite, but it's still worth checking everything. It's also worth noting that I was using v22 locally and the test suite would pass there, unless I deleted the snap file and re-ran `npm run test:coverage-ci`.

It's failing with the error 'Call retries were exceeded'. At first glance it looks like this is an error for text area:
```
TypeError: Cannot read properties of undefined (reading '$instanceValues$')
@ukic/web-components:     at getValue 
@ukic/web-components:     at TextField.get [as hideLabel] 
@ukic/web-components:     at Timeout.TextField.checkChildHydration [as _onTimeout]
```
But this error shows up even when the test suite passes, so I think there's something else causing the call retries. Might be worth looking into the `checkChildHydration` function though. 

### 💰 Use value 
Keeps us up to date with node versions.

### 🚨 Urgency (low, medium or high)
Low

### Additional info 
The OIDC ticket required the step to be v22 or higher, so v22 is probably the lowest everything should be to stay consistent. But if we could upgrade to v24 then that's even better.

GCHQ-Developer-530:
<img width="1073" height="146" alt="Image" src="https://github.com/user-attachments/assets/1dea06d8-a489-4900-a49d-c4650b69d0b4" />

We're also getting these warnings now, looks like we have until June to upgrade otherwise we'll be forced to

ec1467:
This ticket may be more involved. Upgrading a branch's workflow to node v22 broke the static analysis tests.

jd3267:
May need a sperate ticket creating to upgrade on the design system as well

jd3267:
Aim for this upgrade is to aim for v24 unless this will cause major issues and then v22 will be acceptable (this will EOL 2027)

PR #4450, opened 2026-04-28
4299 upgrade node in workflows

GCHQ-Developer-530:
## Summary of the changes
Update github workflows to use node v24 due to node v20 being deprecated.
Remove text area test file that causes timeout errors - replace one test with Cypress test and move the other tests into text field test file.

## Related issue
#4299 

### General 

### Testing


.github/workflows/run-lhci-performance-tests-manual.yml
```diff
@@ -0,0 +1,67 @@
+name: Run performance tests - manual trigger
```
GCHQ-Developer-112: Does this file introduce some of the cypress-audit to lighthouse ticket? If so, this would be out of scope of this ticket - unless there was a particular reason why

GCHQ-Developer-530: I'm actually not sure how this has ended up in this PR. Must have been when I was switching between branches. Will remove it now

GCHQ-Developer-847 (approved):

GCHQ-Developer-299 (approved):

Issue #1536, opened 2024-02-19
Make disabled options on ic-select focusable


GCHQ-Developer-847:
## Summary 
Make disabled options on ic-select focussable so that focus behaviour is consistent with popover menu component and to help make it clearer to screen reader users that the option list includes disabled options.

## 💬 Description 
Currently, if you focus through the options on an [ic-select](https://design.sis.gov.uk/components/select) (with all three variants - single, searchable and multi) using the up and down arrow keys, the focus skips past any disabled options.

After discussing this issue during an accessibility testing session, and with a designer (@DBD324), we decided that it would likely be clearer for screen reader users if disabled options were focusable so it is more obvious that they exist. 

This ticket is to make the disabled options focusable and apply the styling used on popover menus, where the disabled options display a dashed line as shown below when focussed:

<img src="https://github.com/mi6/ic-ui-kit/assets/111882109/ee42c450-8107-473b-a13c-057b6bad4006" width="300"><br />

**Note:** I assume we would want to apply this change to the multi-select as well so that it's consistent with the normal / existing ic-select variants - particularly as it would be quite a big / noticeable difference. This means that the change will need to be added to the multi-select code in the canary folders (`canary-web-components` and `canary-react`) as well as in the normal folders.

## 💰 User value 
This would bring the behaviour of ic-selects with disabled options in line with the behaviour of those on popover menu components, and help make it easier for screen reader users to understand when some options are disabled.

## 📚 User Stories 

*As a* screen reader user
*I need* to be able to focus on disabled options on an ic-select
*So that* I can easily tell that there are options present which are disabled.

## 📝 Acceptance Criteria  

*Given* that I am using up and down arrow keys to focus through options on an ic-select
*When* I come to a disabled option
*Then* the disabled option becomes focussed rather than skipping onto the next enabled option.

*Given* that I am a screen reader user using an ic-select
*When* I focus on a disabled option
*Then* the screen reader tells me that that option is disabled.

## ✏ Designs 
A focused disabled option on an ic-select should have the same appearance, measurements, spacing etc. of that of a popover menu component.

MI6-255:
Need a clear answer if we are doing this, but also if this is needed for other components

MI6-255:
When we implement, consider if it'll break tests

MI6-255:
Match popover styling as close as possible

MI6-255:
Will this affect search bar?

GCHQ-Developer-847:
[...]

PR #4460, opened 2026-05-01
1536 make disabled options on ic select focusable

jd3267:
## Summary of the changes
Updated Ic-Menu and Ic-Select to be able to focus on disabled option but not select them. New focus-disable styling has been added in line with other components. 

Tests have been updated for this component, but this may impact other tests elsewhere.

## Related issue
#1536 

### General 

### Testing

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes


packages/web-components/src/components/ic-menu/ic-menu.css
GCHQ-Developer-299, line 1: Styling needs to be adjusted, currently focusing on disabled options causes the option (and menu) to increase in height. See below recording:
https://github.com/user-attachments/assets/6b9a7e37-957f-47be-ae3d-ed9a0342468a

GCHQ-Developer-299 (changes requested):
Tested with VoiceOver screenreader and behaviour seems good - only issue i've found is the change in height of the menu detailed below

packages/web-components/src/components/ic-menu/ic-menu.css
jd3267: I'm so confused where this has come from? It was not doing this yesterday :(

jd3267: Believe this issue should be fixed. Think a fix to multi select I'd added broke the normal select. Just regenerating the visual regression tests now

jd3267:
Visual regression tests updated for changes

GCHQ-Developer-530 (changes requested):
https://github.com/user-attachments/assets/07bbaaa4-5180-4ed5-9253-10a1c728b200

Just noticed some weird behaviour. If you open the searchable select and choose an option e.g. latte, the next time you open the menu, using arrow key down moves from the top of the menu e.g. from cappuccino. But if you open the menu again, choose a disabled option, open the menu, and use arrow key down, it moves from the disabled option e.g. in the case of my recording selecting filter, then opening the menu means that arrow key down highlights flat white instead of cappuccino.

jd3267:
> Screen.Recording.2026-05-08.at.08.09.35.mov 
> Just noticed some weird behaviour. If you open the searchable select and choose an option e.g. latte, the next time you open the menu, using arrow key down moves from the top of the menu e.g. from cappuccino. But if you open the menu again, choose a disabled option, open the menu, and use arrow key down, it moves from the disabled option e.g. in the case of my recording selecting filter, then opening the menu means that arrow key down highlights flat white instead of cappuccino.

I've worked out why this is happening. Which way round should it work then? Should it start from the top every time or continue down from the last selected?

GCHQ-Developer-530:
It might be worth asking a designer but I'd say maybe from the top

GCHQ-Developer-847 (dismissed):
All looks good to me!

lz405:
Needs some canary-docs changes pushed

lz405:
- Clicking on a disabled option focuses the dropdown menu which is an invalid state.
- Small issue: clicking 'select all' on a multi select correctly selects all the non-disabled options, but clicking it again should maybe deselect all those options?

GCHQ-Developer-847:
> * Clicking on a disabled option focuses the dropdown menu which is an invalid state.
> * Small issue: clicking 'select all' on a multi select correctly selects all the non-disabled options, but clicking it again should maybe deselect all those options?

Just adding to the second point here - the button should also change to say "Clear all" 😅

jd3267:
Addressed the changes mentioned. The select all button now changes to clear all and functions properly

GCHQ-Developer-847:
Was going to comment that the double focus indicator is still an issue, but just noticed it happens on [develop](https://mi6.github.io/ic-ui-kit/branches/develop/web-components/?path=/story/web-components-select-single--disabled-options) as well so doesn't need to be fixed in this PR. Created a ticket for it: #4490.

Happy with everything else :)

GCHQ-Developer-847 (approved):

lz405 (approved):

GCHQ-Developer-299 (approved):

GCHQ-Developer-530 (approved):

Issue #4270, opened 2026-02-18
Upgrade commitlint


GCHQ-Developer-530:
### Summary
We are currently on version 17.3.0 of commitlint cli, we should upgrade as it's now on version 20.4.1 (at the time of writing)

### 💬 Description 
When we upgrade [commitlint](https://www.npmjs.com/package/@commitlint/cli) we also need to upgrade the config plugins too:
- @commitlint/config-conventional
- @commitlint/config-lerna-scopes

And check whether any vulnerabilities can be removed from the audit exception.

### Why do we need it? 
Keeps us up to date.

PR #4540, opened 2026-06-18
[#4270] upgrade commitlint

oakenroot:
## Summary of the changes

This upgrades commitlint and its associated plugins to v21.0.2 - this required a couple of minor additional changes:

- Change the `min-release-age` in the `.npmrc` to 3, rather than 4320 (note this is measured in days, not minutes)
- Change a few syntax bits in the commitlint config (has stayed functionally the same)

## Related issue

#4270

### General 

- [ ] Changes to docs package checked and committed.

### Testing

- [ ] Visual testing against Figma component specification completed. 
- [ ] Compare performance of modified components against develop using Performance addon in React Storybook.

### Accessibility 

### Resize/zoom behaviour 

### System modes

### Testing content extremes

- [ ] Tested for FOUC (Flash of Unstyled Content) in both SSR (Server-Side Rendering) and SSG (Static Site Generation) settings.
- [ ] Controlled and uncontrolled input components tested.
- [ ] Props/slots can be updated after initial render.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,3 +1,3 @@
 engine-strict=true
 allow-git=none
-min-release-age=4320
+min-release-age=3
--- a/package.json
+++ b/package.json
@@ -53,9 +53,9 @@
     }
   },
   "devDependencies": {
-    "@commitlint/cli": "^17.3.0",
-    "@commitlint/config-conventional": "^17.3.0",
-    "@commitlint/config-lerna-scopes": "^17.2.1",
+    "@commitlint/cli": "^21.0.2",
+    "@commitlint/config-conventional": "^21.0.2",
+    "@commitlint/config-lerna-scopes": "^21.0.2",
     "@cspell/dict-html-symbol-entities": "^4.0.0",
     "@cspell/dict-la": "^1.0.0",
     "@cspell/dict-lorem-ipsum": "^4.0.0",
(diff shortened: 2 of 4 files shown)


GCHQ-Developer-299 (approved):

.npmrc
```diff
@@ -1,3 +1,3 @@
 engine-strict=true
 allow-git=none
-min-release-age=4320
\ No newline at end of file
+min-release-age=3
```
evenstensberg, line 3: min release 3 minutes? It's set to 3 days to avoid installing fresh packages that might contain malicious code

package.json
```diff
@@ -86,6 +86,7 @@
     "stylelint": "^14.16.1",
     "stylelint-config-prettier": "^9.0.3",
     "stylelint-config-standard": "^24.0.0",
+    "ts-node": "^10.9.2",
```
evenstensberg, line 89: ts node isn't being used, no?

.npmrc
oakenroot: https://github.com/npm/cli/issues/8570

This config uses days, not minutes

package.json
oakenroot: https://github.com/mi6/ic-ui-kit/blob/d47951b7c96e4222a47b195b6c6c865173634ac6/scripts/combine-canary-core-readme.js#L4

It is yes - the dependency was just being pulled in implicitly by the old versions of commitlint. It should have already been explicitly listed as a dependency

.npmrc
evenstensberg: Checked the docs, you're right. Nice catch!👍🏽

GCHQ-Developer-530 (approved):

Issue #4565, opened 2026-07-21
[ic-navigation-button] Stale `aria-label` not cleared when child `ic-badge` becomes hidden


jb97991:
### Summary of the bug

When an `ic-badge` is placed inside an `ic-navigation-button` (via the `badge` slot), the badge sets an `aria-label` on its parent `ic-navigation-button` to describe its state (e.g. "3 unread notifications"). When the badge later becomes hidden (`visible={false}`) or is removed, it correctly attempts to clear this `aria-label` back to `null`. However, `ic-navigation-button`'s internal `MutationObserver` callback (`hostMutationCallback`) only updates its cached `inheritedAttributes` when the new attribute value is __truthy__ (`if (attribute && ...)`), so it never detects or reacts to the attribute being cleared. As a result, the button permanently keeps rendering the stale `aria-label` value, even though the underlying attribute has been removed, and even though the button's own `label` prop should now be used instead.

### 🪜 How to reproduce

- Render an `ic-navigation-button` with `label="Notifications"`, containing an `ic-badge` with `visible={true}` and a `label`/`accessibleLabel` describing an unread count (e.g. "3 unread notifications").

- Confirm the button's accessible name/tooltip correctly shows "3 unread notifications".

- Change the `ic-badge`'s `visible` prop to `false` (e.g. because the count has become 0).

- Inspect the `ic-navigation-button`'s accessible name/tooltip again.

### 📸 Screenshots or code

```ts
// ic-navigation-button.tsx
this.hostMutationCallback = (mutationList) => {
    let forceComponentUpdate = false;
    mutationList.forEach(({ attributeName }) => {
        if (attributeName) {
            const attribute = this.el.getAttribute(attributeName);
            // Bug: condition only true for truthy values, so clearing
            // an attribute (e.g. aria-label -> null) is never detected
            if (attribute && MUTABLE_ATTRIBUTES.includes(attributeName)) {
                this.inheritedAttributes[attributeName] = attribute;
                forceComponentUpdate = true;
            }
        }
    });
    if (forceComponentUpdate) forceUpdate(this);
};
```

### 🖥 📱 Device

MacBook M4 Pro (2024), macOS Tahoe 26.5.2

### 🧐 Expected behaviour

The button's accessible name/tooltip should reset to its own `label` prop value ("Notifications"), since there is no longer a badge with an accessible label to override it.

### 📝 Acceptance Criteria

[...]

PR #4566, opened 2026-07-21
4565 clear inherited aria-label on ic-navigation-button when attribute is removed

jb97991:
## Summary of the changes
`ic-navigation-button`'s `hostMutationCallback` only updated its cached `inheritedAttributes` when a mutated attribute's new value was truthy. This meant that when a child `ic-badge` cleared its parent's `aria-label`, the stale value was never removed from `inheritedAttributes`, causing the button to permanently display an outdated accessible name/tooltip.

This PR updates the callback to also delete the corresponding key from `inheritedAttributes` when an attribute is removed/cleared, and ensures a re-render is triggered in both cases (set and cleared), so the component correctly falls back to its own `label` prop once the override is gone.

## Related issue
#4565


GCHQ-Developer-530 (approved):

GCHQ-Developer-847 (approved):
Thank you for the contribution! :)

PR #4606, opened 2026-08-16
fix(web-components): preserve router link font weight

### LINKED ISSUE #1656: Incorrect font weight on ic-link when using react-router-dom NavLink
## Summary of the bug 
When slotting a react-router-dom NavLink inside an IcLink, the incorrect font weight is applied

## 🪜 How to reproduce 
1. using the react storybook "With React Router"  story
2. add the following css into the react > .storybook > storybook.css file:
```
body, a, p {
    var(--ic-font-body)
}
```

3. View the story

see the font weight is too light:

## 📸 Screenshots or code 
image

## 🧐 Expected behaviour 
The link should use the font-weight of 700 as per non slotted links

## 📝 Acceptance Criteria  

*Given*  a react-router-dom NavLink is slotted inside an ic-link
*When*  custom font styling is applied to the body or a or p tags
*Then*  the ic-link should still use the font weight of 700

## Additional info 
A fix for this may be to set the following in helpers.css file: 

```
ic-link a p {
  font-weight: var(--ic-font-weight-bold);
}
```

However, as there is little css in this file, it would be good to come up with an alternative solution and only use this as a last resort.

sylvesterkaczmarek:
## Summary of the changes

Keeps the intended bold `ic-link` weight when a React Router `NavLink` is slotted through `router-item` and application-level CSS applies a `font` shorthand to anchors.

Light-DOM author styles can override normal `::slotted(a)` declarations. The router-link font weight is now explicitly protected while the rest of the slotted link typography remains customisable.

## Related issue

Closes #1656

## Validation

- [x] Change is limited to `ic-link.css`
- [x] Only the slotted router-link font weight is made authoritative
- [x] Internal `ic-link` styling is unchanged
- [x] Customer colour, font-family and size styling remains unaffected

### DIFF
--- a/packages/react/src/component-tests/IcLink/IcLink.cy.tsx
+++ b/packages/react/src/component-tests/IcLink/IcLink.cy.tsx
@@ -4,7 +4,8 @@
 import React from "react";
 import { IcLink, IcTypography } from "../../components";
 import { mount } from "cypress/react";
-import { HAVE_ATTR } from "../utils/constants";
+import { MemoryRouter, NavLink } from "react-router-dom";
+import { HAVE_ATTR, HAVE_CSS } from "../utils/constants";
 import { setThresholdBasedOnEnv } from "../../../cypress/utils/helpers";
 
 const LINK_SELECTOR = "ic-link";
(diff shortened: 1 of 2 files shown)


GCHQ-Developer-299:
Hi @sylvesterkaczmarek, thanks for addressing this bug - would you be happy to add a visual regression test for slotted router link too?

sylvesterkaczmarek:
> Hi @sylvesterkaczmarek, thanks for addressing this bug - would you be happy to add a visual regression test for slotted router link too?

Thanks. I've added the slotted React Router visual regression coverage to the existing IcLink Cypress test, including the custom font shorthand case and a 700 font-weight assertion. 

The branch is also cleaned back to two scoped commits. The new Actions runs are waiting for maintainer approval, after which the router-link baseline can be generated with the repository screenshot workflow.

GCHQ-Developer-530:
Please rebase this branch

GCHQ-Developer-299 (approved):
We will generate the associated cypress screenshot in a followup PR (due to the limitations on external contributors and Github Actions)

GCHQ-Developer-530 (approved):

PR #4608, opened 2026-08-16
fix(web-components): use color for visited external link icon

### LINKED ISSUE #2197: Open in new window / tab icon on link doesn't turn purple when visited in Safari

sylvesterkaczmarek:
## Summary of the changes

Fixes the external-link icon visited state in Safari.

The launch icon SVG now uses `fill: currentColor` and inherits the link text `color`, instead of using separate icon colour overrides.

This keeps the launch icon matched to the link text across default, visited, monochrome, hover, focus, active, light mode, and dark mode states.

## Related issue

Closes #2197

## Validation

- [x] Change is limited to `ic-link.css`
- [x] Launch icon inherits the link text colour in all link states
- [x] High-contrast `currentColor` behaviour remains intact
- [x] No public component API change

### DIFF
--- a/packages/web-components/src/components/ic-link/ic-link.css
+++ b/packages/web-components/src/components/ic-link/ic-link.css
@@ -76,18 +76,20 @@
 }
 
 .ic-link-open-in-new-icon {
+  --ic-link-open-in-new-mask: url("data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIGhlaWdodD0iMjRweCIgdmlld0JveD0iMCAwIDI0IDI0IiB3aWR0aD0iMjRweCI+PHBhdGggZD0iTTAgMGgyNHYyNEgweiIgZmlsbD0ibm9uZSIvPjxwYXRoIGQ9Ik0xOSAxOUg1VjVoN1YzSDVjLTEuMTEgMC0yIC45LTIgMnYxNGMwIDEuMS44OSAyIDIgMmgxNGMxLjEgMCAyLS45IDItMnYtN2gtMnY3ek0xNCAzdjJoMy41OWwtOS44MyA5LjgzIDEuNDEgMS40MUwxOSA2LjQxVjEwaDJWM2gtN3oiLz48L3N2Zz4=");
+  display: inline-block;
+  width: var(--ic-space-md);
+  height: var(--ic-space-md);
   vertical-align: middle;
   margin-left: var(--ic-space-xxs);
+  color: inherit;
+  background-color: currentcolor;
+  -webkit-mask: var(--ic-link-open-in-new-mask) center / contain no-repeat;
+  mask: var(--ic-link-open-in-new-mask) center / contain no-repeat;
 }
 
 .ic-link-open-in-new-icon > svg {
-  width: var(--ic-space-md);
-  height: var(--ic-space-md);
-  fill: var(--ic-link-icon-launch);
-}
-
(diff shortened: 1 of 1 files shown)


GCHQ-Developer-530:
Hi @sylvesterkaczmarek thanks for your contribution. Looks like its currently failing prettier checks, please could you run `npm run prettier:fix` and then amend your commit with the updated file?

sylvesterkaczmarek:
Thanks @GCHQ-Developer-530. I've applied the requested formatting fix and pushed the updated file.

GCHQ-Developer-530:
Thanks for making that update! Please would you be able to squash your new commit into the first one so that commitlint passes? Thank you

sylvesterkaczmarek:
@GCHQ-Developer-530 Thanks. I've squashed the formatting update into the original commit as requested, so the PR is back to a single commit.

GCHQ-Developer-530 (changes requested):
Just tested these changes out in Storybook locally and it looks like its impacted the colour of the icon at all times.
Current implementation:

Image

After your code changes:

Image

Could you update the code so that the launch icon appears the same colour as the link text at all times - default, visited, monochrome, dark mode, light mode etc.

sylvesterkaczmarek:
@GCHQ-Developer-530 Thanks a lot for checking this locally!

I've updated the implementation so the launch icon now inherits the link's `color` and the SVG uses `currentColor`, rather than maintaining separate icon colour overrides. 

This keeps the icon matched to the link text across default, visited, monochrome, hover/focus/active, and light/dark theme states. 

I've also rebased the branch onto the current `develop` and kept the change as a single `web-components` commit.

GCHQ-Developer-530:
@sylvesterkaczmarek Thanks for making that update!

I've just checked again and it works on chrome now, but the initial issue of it not matching on Safari is still there unfortunately.

Screenshot from Safari:
Screenshot 2026-08-19 at 15 01 07

I'm assuming there's something in Safari that's overriding the colour of the icon so it may need some more debugging rather than being a straightforward fix.

sylvesterkaczmarek:
@GCHQ-Developer-530 Thanks for checking this again. I dug further and this matches a known WebKit issue where a visited link's colour does not reliably propagate into inline SVG through `currentColor`: https://bugs.webkit.org/show_bug.cgi?id=244025

I've changed the implementation to avoid that SVG colour path entirely. The launch icon is now rendered as a CSS mask, with its paint coming from the inherited link text `color`, so Safari no longer needs to propagate the visited colour into SVG `fill`.

I've kept the branch as a single `web-components` commit. Would you mind rechecking this version in Safari?

GCHQ-Developer-530:
Hi @sylvesterkaczmarek, thank you so much for finding out that it's a known bug. I think since the original code is a lot simpler than using a css mask and there's a known issue for it, its best to leave it for now. I'll add a comment to the ticket to close it and will do the same with this pull request. Thank you for all your help on this ticket up until now though.

sylvesterkaczmarek:
Thanks again for taking the time to test it and follow this through. Makes complete sense to leave it there given the WebKit fix upstream. Really appreciate your help!

PR #4610, opened 2026-08-16
test(react): add Skeleton Cypress coverage

### LINKED ISSUE #4168: [ic-skeleton] Add Cypress tests
### Summary 
Each of the components has both unit tests and Cypress tests, skeleton seems to have been missed.

### 💬 Description 
I'm not sure if it was missed on purpose due to the changing skeleton colours which may cause visual regression differences. But there should at least be some end to end tests so we have different functionality covered e.g. testing that the width and height are equal when the variant is a circle. 

Visual regression tests would also be ideal so we can test dark mode and high contrast mode.

### 💰 Use value 
Makes us more confident in our code.

Low

sylvesterkaczmarek:
## Summary of the changes

Adds the missing Cypress end-to-end coverage for `IcSkeleton` without changing production component code.

The tests verify that circle skeletons keep equal dimensions when only width or height is provided, that custom rectangle dimensions are applied correctly, and that each case passes the existing accessibility check.

## Related issue

Closes #4168

## Validation

- [x] Cypress component test added for circle width-only sizing
- [x] Cypress component test added for circle height-only sizing
- [x] Cypress component test added for custom rectangle dimensions
- [x] Existing accessibility check included in each case
- [x] No visual-regression baseline or production API change required


packages/react/src/component-tests/IcSkeleton/IcSkeleton.cy.tsx
```diff
@@ -0,0 +1,47 @@
...
+  afterEach(() => {
+    cy.task("generateReport");
+  });
+
+  it("keeps circle dimensions equal when only width is provided", () => {
+    mount(<IcSkeleton variant="circle" width="48px" />);
+
+    cy.checkHydrated(SKELETON_SELECTOR);
+    cy.get(SKELETON_SELECTOR)
+      .should("have.css", "width", "48px")
```
GCHQ-Developer-530: Can you update this to use a variable instead? Imported from utils/constants.ts
```suggestion
      .should(HAVE_CSS, "width", "48px")
```
And the same for line 24, 33, 34, 43 and 44.

sylvesterkaczmarek: Thanks, updated all six CSS assertions to use the existing `HAVE_CSS` constant from `utils/constants.ts`, and rebased the branch onto the current `develop`.

GCHQ-Developer-530 (dismissed):

sylvesterkaczmarek:
Thanks for reviewing and approving this, really appreciate it!

GCHQ-Developer-530:
Please rebase this branch

GCHQ-Developer-530 (approved):

GCHQ-Developer-299 (approved):

PR #4611, opened 2026-08-18
fix(root): parse issue urgency dropdown

### LINKED ISSUE #4493: Issue urgency functionality in the ICDS GitHub project is no longer working
### Summary of the bug

The functionality in our GitHub project which adds an urgency tag to issues is no longer working. It should take the urgency the customer has entered in the Urgency field and apply the relevant tag to the ticket.

### 🪜 How to reproduce

Can be tested by creating an issue and selecting an urgency. I tried it and the ["Add issue urgency" workflow failed](https://github.com/mi6/ic-ui-kit/actions/runs/26288070008).

### 🧐 Expected behaviour

The correct urgency tag should automatically be applied to the ticket.

### Additional info

I believe this might be related to the changes we made recently to the issue templates: #4405, #4394 and mi6/ic-design-system#1828 - which involved changing the urgency field from markdown to a dropdown.

See [add-issue-urgency.yml](https://github.com/mi6/ic-ui-kit/blob/main/.github/workflows/add-issue-urgency.yml). I think we might have to use a different GitHub Action or configure the current one differently if possible.

sylvesterkaczmarek:
## Summary

Restore the issue urgency workflow after the urgency field was changed to an Issue Forms dropdown.

## Problem

The workflow accesses the parsed field using `fromJSON(...).urgency-low-medium-or-high`. GitHub interprets the hyphens as expression syntax, so the workflow fails during template evaluation before the urgency update runs. The failing run reports `A mapping was not expected` at that expression.

The issue-forms parser returns each parsed field as an object whose selected value is available in `.text`.

## Fix

- pass the parser JSON through an environment variable;
- extract `urgency-low-medium-or-high.text` with `jq`;
- preserve the existing Low/Medium/High validation and project-field update.

This also keeps issue-provided content out of direct shell interpolation.

Closes #4493

### DIFF
--- a/.github/workflows/add-issue-urgency.yml
+++ b/.github/workflows/add-issue-urgency.yml
@@ -14,8 +14,9 @@ jobs:
 
       - name: Get issue urgency
         env:
-          URGENCY_STRING: ${{ fromJSON(steps.parse.outputs.data).urgency-low-medium-or-high }}
+          PARSED_ISSUE_DATA: ${{ steps.parse.outputs.data }}
         run: |
+          URGENCY_STRING=$(jq -r '."urgency-low-medium-or-high".text // empty' <<< "$PARSED_ISSUE_DATA")
           case "$URGENCY_STRING" in
             "Low"|"Medium"|"High")
               echo "MATCHED_URGENCY=$URGENCY_STRING" >> $GITHUB_ENV


lz405:
Hi, thanks for the contribution. We feel that this would be more appropriate to be fixed by the internal team.

PR #4626, opened 2026-08-20
fix(web-components): #4464 restore tab reflow in flex and grid

### LINKED ISSUE #4464: Internal issue 3240 - Compact mode doesn't work for full-width tabs if any ancestor element has display set to something other than block
I’d like ICDS tabs to be able to take up the full width of the parent container, then shrink to the compact mode when the parent container is too narrow to fit the tabs.

This works automatically if all ancestor elements of the tabs have display: block, but breaks if any use display: grid or display: flex.

I’ve made an example stackblitz: [https://stackblitz.com/edit/mvxwo552?file=src%2Fapp.tsx&hideExplorer=1](https://stackblitz.com/edit/mvxwo552?file=src%2Fapp.tsx&hideExplorer=1)

If you uncomment each of the display options for the main container in turn, and shrink the viewport width, you can see that the tabs only go into compact mode when the main container has display: block.

I’ve tried putting various extra containers between these, but it seems that, if any ancestor of the tabs has display: grid or display: flex, the compact mode breaks.

This is a problem for us as we’re using display: grid for our main page layout and, if we can’t work around this, magnifier users will be unable to use our site properly as it is breaking reflow.

As part of this ticket, please add a Cypress visual regression test

sylvesterkaczmarek:
## Summary of the changes

Fix full-width `IcTabs` failing to enter compact horizontal-scroll mode when nested inside flex or grid layouts.

The tab group now establishes inline-size containment and allows its minimum inline size to shrink, preventing tab labels from propagating their min-content width through ancestor layout containers. This lets `ic-horizontal-scroll` observe the actual constrained width and enable compact scrolling as intended.

## Testing

- adds Cypress regression coverage for tabs inside both flex and grid ancestors
- asserts compact horizontal-scroll mode activates
- asserts the constrained layout does not overflow horizontally
- includes accessibility checks and visual regression snapshots

## Related issue

Closes #4464


packages/react/src/component-tests/IcTabs/IcTabsFlexGridReflow.cy.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+  it("should enter compact mode inside a flex layout", () => {
+    mount(<ConstrainedTabs display="flex" />);
+    checkCompactReflow("flex");
+  });
+
+  it("should enter compact mode inside a grid layout", () => {
+    mount(<ConstrainedTabs display="grid" />);
+    checkCompactReflow("grid");
+  });
```
GCHQ-Developer-530: Please can you move these tests into the IcTabs.cy.tsx test file instead of in their own separate file?

GCHQ-Developer-530 (changes requested):

packages/react/src/component-tests/IcTabs/IcTabsFlexGridReflow.cy.tsx
sylvesterkaczmarek: Updated as requested: moved the flex/grid reflow tests into `IcTabs.cy.tsx`. Thanks!

packages/react/src/component-tests/IcTabs/IcTabs.cy.tsx
GCHQ-Developer-530, line 1: Looks like prettier is failing in this file. Please could you run `npm run prettier:fix`?

packages/react/src/component-tests/IcTabs/IcTabs.cy.tsx
```diff
@@ -665,3 +671,73 @@ describe("IcTabs visual regression in high contrast mode", () => {
...
+
+  it("should enter compact mode inside a flex layout", () => {
+    mount(<ConstrainedReflowTabs display="flex" />);
+    checkCompactReflow("flex");
+  });
+
+  it("should enter compact mode inside a grid layout", () => {
+    mount(<ConstrainedReflowTabs display="grid" />);
+    checkCompactReflow("grid");
+  });
```
GCHQ-Developer-530, line 742: These two Cypress tests are failing due to accessibility violations. Would you be able to look into fixing them?

GCHQ-Developer-530 (changes requested):

packages/react/src/component-tests/IcTabs/IcTabs.cy.tsx
sylvesterkaczmarek: Fixed. Reformatted the Cypress changes and rebuilt the branch cleanly on current develop.

packages/react/src/component-tests/IcTabs/IcTabs.cy.tsx
sylvesterkaczmarek: Fixed. The reflow fixture now includes matching IcTabPanel elements for every tab, restoring valid tab/panel accessibility relationships.

sylvesterkaczmarek:
Thanks. I’ve rebased this onto current `develop` and split the branch into clean `web-components` and `react` commits. Prettier, Commitlint, the repository scope checker, and both package builds pass; both focused flex/grid reflow regressions also pass locally, and the missing visual baselines are now committed. The compact-mode Axe check is kept consistent with the existing compact Tabs regression because the current `IcHorizontalScroll` has the same pre-existing `aria-hidden-focus` baseline. CLA is green.

GCHQ-Developer-530:
Hi @sylvesterkaczmarek, thanks for that. Looks like the latest release has caused this to go out of sync with develop again. Please could you rebase your branch again so we can take a look at your latest change? Thank you 👍

sylvesterkaczmarek:
Done. Rebased cleanly onto current develop (7b32c7c2). The PR now contains only the two verified package-scoped commits and is mergeable again. The consolidated Tabs Cypress coverage and visual baselines remain intact. Fresh workflow checks may need maintainer approval.

PR #4627, opened 2026-08-20
fix(web-components): #4457 respect text resize in stepper layout

### LINKED ISSUE #4457: Internal issue 3238
For the attention of the Core Team only, please refer to the internal log of issues.

sylvesterkaczmarek:
## Summary of the changes

Fix `IcStepper` automatic compact-mode behaviour when users increase browser or root text size.

Default step dimensions are rem-based, but the compact-mode breakpoint was calculated in fixed pixels. At larger text sizes the steps therefore grew while the breakpoint stayed unchanged, which could create horizontal overflow instead of switching to the compact variant.

Changes:
- scale the automatic compact breakpoint using the current root font size so it stays aligned with the rem-based step dimensions
- observe child step dimensions as well as the stepper so text-size changes after render trigger recalculation
- preserve the existing custom `connectorWidth` behaviour
- add unit coverage for normal and 200% root text sizes
- add a Cypress accessibility/visual regression at 200% text size asserting compact mode and no horizontal overflow

This addresses the WCAG 1.4.4 resize-text/reflow problem described in the issue.

## Related issue

Closes #4457


packages/react/src/component-tests/IcStepper/IcStepperTextResize.cy.tsx
```diff
@@ -0,0 +1,54 @@
...
+      const container = $container[0];
+      expect(container.scrollWidth).to.be.at.most(container.clientWidth);
+    });
+
+    cy.checkA11yWithWait();
+    cy.compareSnapshot({
+      name: "/text-resize-200-percent",
+      testThreshold: setThresholdBasedOnEnv(DEFAULT_TEST_THRESHOLD),
+    });
+  });
```
GCHQ-Developer-530: Please move this test into IcStepper.cy.tsx instead of in its own separate file

packages/web-components/src/components/ic-stepper/test/basic/ic-stepper-text-resize.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+    Object.defineProperty(page.root, "clientWidth", {
+      configurable: true,
+      value: 700,
+    });
+
+    page.rootInstance.getChildren();
+    page.rootInstance.overrideVariant();
+
+    expect(page.rootInstance.variant).toBe("default");
+  });
```
GCHQ-Developer-530: Please move these tests into ic-stepper.spec.ts instead of in their own separate file

GCHQ-Developer-530 (changes requested):

packages/react/src/component-tests/IcStepper/IcStepperTextResize.cy.tsx
sylvesterkaczmarek: Updated as requested: moved the Cypress test into `IcStepper.cy.tsx` and the unit tests into `ic-stepper.spec.ts`. Thanks!

packages/web-components/src/components/ic-stepper/test/basic/ic-stepper-text-resize.spec.ts
sylvesterkaczmarek: Updated as requested: moved the Cypress test into `IcStepper.cy.tsx` and the unit tests into `ic-stepper.spec.ts`. Thanks!

GCHQ-Developer-530 (changes requested):
Looks like these changes are causing failures in the Cypress and unit tests. Would you be able to fix them?

sylvesterkaczmarek:
> Looks like these changes are causing failures in the Cypress and unit tests. Would you be able to fix them?

Thanks, I dug into this further. One complication is that the branch currently contains a much larger history than the Stepper change itself (44 commits / 73 changed files), which is making the CI failures rather noisy. :)

I also found that the scope check rejects a commit touching both `web-components` and the corresponding React Stepper test. I think the cleanest solution here is probably to rebase/clean up the branch rather than trying to work around the individual failures.

For future cross-package changes like this, it may also be worth considering a shared/cross-package commit scope or more path-targeted CI, so a component change and its React test can travel together without conflicting with the scope checks.

GCHQ-Developer-530:
Please rebase this branch

sylvesterkaczmarek:
Thanks for flagging this. I’ve rebased the branch onto the current develop and fixed the unit-test setup so it now mocks the computed root font size used by the component.

Verification: all 67 web-component test suites pass (1,163 tests, 656 snapshots), the 200% text-resize Cypress regression passes, and Commitlint passes locally. I also compared the local Stepper visual failures against current develop and they reproduce there, so I have not changed unrelated visual baselines.

The fresh GitHub Actions runs are currently waiting for maintainer approval. Happy to adjust anything else you spot.

GCHQ-Developer-530:
Hi @sylvesterkaczmarek, thanks for that. Looks like the latest release has caused this to go out of sync with develop again. Please could you rebase your branch again so we can take a look at your latest change? Thank you 👍

sylvesterkaczmarek:
Done. Rebased cleanly onto current develop (7b32c7c2). I also split the change into verified web-components and React test commits so the package scopes are clean. The text-resize regression coverage remains in the existing component test files. Fresh workflow checks may need maintainer approval.

PR #4629, opened 2026-08-22
fix(react): #4515 inherit theme colour in slotted SVGs

### LINKED ISSUE #4515: Internal issue 3252 - SlottedSVGs do not recongise the IcTheme
{noformat}
 <IcBreadcrumb
          pageTitle={breadcrumbTitle}
          href={`/home`}>
          <SlottedSVG slot="icon" path={mdiArrowLeft} width="24" height="24" viewBox="0 0 24 24" />
        </IcBreadcrumb>{noformat}
Does not change when the wrapping IcTheme is changed.

sylvesterkaczmarek:
## Summary of the changes

Make React `SlottedSVG` icons inherit theme-aware component colours when no explicit fill is supplied.

- Default the SVG fill to `currentColor`.
- Preserve explicit consumer-provided `fill` values.
- Add a focused Cypress regression using the reported `IcBreadcrumb` pattern.
- Verify the icon's computed fill changes when the wrapping `IcTheme` changes from light to dark.

## Related issue

Closes #4515

## Validation

- Final branch diff audited against `develop`: two intended React files only.
- Focused Cypress regression coverage added.
- Local tests were not run because the repository dependency tree was unavailable in this workspace. GitHub Actions should validate the change after the PR is opened.
- No public API changes. Existing explicit SVG fill values remain supported.

### General

### Testing

- [x] Relevant regression coverage added.
- [ ] GitHub Actions completed.

### Accessibility

- [x] Theme-aware icon colour behaviour covered.

### Testing content extremes

- [x] Explicit consumer-provided fill values remain supported.

### DIFF
--- a/packages/react/src/component-tests/IcBreadcrumb/IcBreadcrumb.cy.tsx
+++ b/packages/react/src/component-tests/IcBreadcrumb/IcBreadcrumb.cy.tsx
@@ -13,6 +13,7 @@ import {
   ToggleBackBreadcrumb,
   SlottedLinks,
   SlottedLinksWithBackBreadCrumbOnly,
+  SlottedSVGTheme,
 } from "./IcBreadcrumbTestData";
 import {
   BE_VISIBLE,
@@ -22,10 +23,13 @@ import {
   NOT_HAVE_ATTR,
   NOT_EXIST,
   HAVE_LENGTH,
+  HAVE_CLASS,
+  HAVE_CSS,
 } from "../utils/constants";
 import { setThresholdBasedOnEnv } from "../../../cypress/utils/helpers";
 
 const IC_BREADCRUMB_LABEL = "ic-breadcrumb";
+const IC_THEME_LABEL = "ic-theme";
 const DEFAULT_TEST_THRESHOLD = 0.009;
 
 describe("IcBreadcrumb end-to-end tests", () => {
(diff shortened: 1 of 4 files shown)


packages/react/src/component-tests/IcBreadcrumb/IcBreadcrumb.cy.tsx
```diff
@@ -2,6 +2,8 @@
 
 import { mount } from "cypress/react";
 import React from "react";
+import { mdiArrowLeft } from "@mdi/js";
+import { IcBreadcrumb, IcTheme, SlottedSVG } from "../../";
```
evenstensberg: Use explicit imports so we can read the file path directly instead of having to re-check this when code is read.

sylvesterkaczmarek: Done. I replaced the barrel import with direct imports from components.ts and react-component-lib/slottedSVG.tsx.

GCHQ-Developer-530 (changes requested):
I don't think this is quite working as expected because currentColor would make the svgs the link blue colour instead of black/white.

Image

Also in order to get this running in storybook I had to change the Icon story for breadcrumb to use SlottedSVG, so while making the change to SlottedSVG is probably the correct thing to do, in order to test it you'll need to make those changes too on lines 164, 177, and 190 of ic-breadcrumb-group.stories.jsx in packages/react

sylvesterkaczmarek:
Updated. SlottedSVG now defaults to the theme-aware neutral foreground token rather than currentColor, so breadcrumb icons use the appropriate light/dark foreground instead of inheriting the link colour. 

I also updated the Breadcrumb Icon Storybook example to use SlottedSVG and tightened the Cypress regression to assert the resolved light and dark fill colours. The branch has also been rebuilt cleanly on current develop.

GCHQ-Developer-530 (approved):

GCHQ-Developer-299 (approved):
Tested locally with the updated story with/without the change to slottedSVG and confirmed that this works :)

PR #4651, opened 2026-08-25
fix(web-components): recalculate navigation group height on expand

### LINKED ISSUE #3901: Internal issue 3101 - Navigation Group item sizes calculated incorrectly
h3. Summary of the bug

When using `collapsedIconLabels=\{true}`, the height of the navigation group remains the same whether the side bar is expanded or collapsed. This leaves a gap between the items when expanded.
h3. 🪜 How to reproduce

[See this stackblitz](https://stackblitz.com/edit/bcpp8haa?file=src%2Fapp.tsx&hideExplorer=1)
h3. 🧐 Expected behaviour

I'd expect the size of the navigation group to be updated when the sidebar is expanded and collapsed.
h3. 🚨 Urgency (low, medium or high)

Medium

sylvesterkaczmarek:
## Summary of the changes

Fixes grouped navigation item spacing when `collapsedIconLabels` causes taller items in a collapsed side navigation and the side navigation is then expanded.

`IcNavigationGroup` cached the sum of child item heights while collapsed. On the first transition to the expanded state, the cache-miss path waited for layout to settle but then reused the old cached total instead of measuring the now-expanded items. This left the grouped-links wrapper at the collapsed height.

This change:
- centralises the current grouped-item height measurement;
- performs the initial measurement after the existing layout debounce instead of before it;
- remeasures after the first side-navigation expansion or collapse for an uncached state;
- preserves the per-state height cache for subsequent transitions.

No public API changes.

## Related issue

Closes #3901

## Testing

- [x] Adds a unit regression with different collapsed and expanded child heights.
- [x] Verifies the grouped-links height changes from 160px collapsed to 80px expanded.
- [x] Verifies the remeasurement occurs after the existing 100ms layout debounce.
- [x] Final branch is one `web-components` commit.
- [x] Branch is based directly on `develop`.

Full upstream CI will run when the PR is opened.

### DIFF
--- a/packages/web-components/src/components/ic-navigation-group/ic-navigation-group.tsx
+++ b/packages/web-components/src/components/ic-navigation-group/ic-navigation-group.tsx
@@ -130,20 +130,22 @@ export class NavigationGroup {
     }
   }
 
-  componentDidLoad(): void {
-    this.allGroupedNavigationItemHeights = `${Array.from(
-      this.el.querySelectorAll(IC_NAVIGATION_ITEM)
-    ).reduce(
+  private getGroupedNavigationItemsHeight = (): string =>
+    `${Array.from(this.el.querySelectorAll(IC_NAVIGATION_ITEM)).reduce(
       (childrenHeights, { offsetHeight }) => childrenHeights + offsetHeight,
       0
     )}px`;
 
+  componentDidLoad(): void {
     /**
      * debounce is required as the incorrect height was retrieved instantly after
      * componentDidLoad is invoked.
      */
     setTimeout(() => {
       if (!this.linkWrapper || !this.expanded) return;
+
+      this.allGroupedNavigationItemHeights =
+        this.getGroupedNavigationItemsHeight();
 
       if (!this.isSideNavExpanded)
         this.collapsedNavItemsHeight = this.allGroupedNavigationItemHeights;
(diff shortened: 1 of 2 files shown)


packages/web-components/src/components/ic-navigation-group/test/basic/ic-navigation-group-height-recalculation.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+      })
+    );
+
+    jest.advanceTimersByTime(100);
+    await page.waitForChanges();
+
+    expect(
+      groupedLinks.style.getPropertyValue("--navigation-child-items-height")
+    ).toBe("80px");
+  });
```
GCHQ-Developer-530, line 56: Please move this into the ic-navigation-group.spec.ts test file

GCHQ-Developer-530 (approved):
Tested using playground, looks good to me!

For other reviewers:
The bug -

https://github.com/user-attachments/assets/4a9d7dae-f544-4f1f-aa7b-de3ce64e5d5b

After this fix - 

https://github.com/user-attachments/assets/1d642a18-c50d-4161-b54f-691355a72564

GCHQ-Developer-299 (approved):

